PR #2, opened 2015-12-14
Activate travis

wence-:
Sets up build testing on travis.  This branch is tested, for example, here: https://travis-ci.org/wence-/cofs/builds/96775578

Also pytestifies some of the tests using fixtures.

The travis testing framework uses travis caching to mean that we don't have to reinstall firedrake every time (also the FFC kernel cache is cached, so only the first run should require building the kernels).  These need to be manually invalidated if the dependencies are upgraded (but this is a one-click thing on the travis website).

I didn't activate code linting, because most of the source tree fails it!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,59 @@
+sudo: false
+notifications:
+  irc:
+    channels: "chat.freenode.net#firedrake"
+    skip_join: true
+    on_success: change
+    on_failure: always
+    template: "%{repository}#%{build_number} (%{branch} - %{commit} : %{author}): %{message} | %{build_url}"
+
+language: python
+python: "2.7"
+addons:
+  apt:
+    packages:
+        - build-essential
+        - python-dev
+        - git
+        - mercurial
+        - python-pip
+        - python-scipy
+        - libopenmpi-dev
+        - openmpi-bin
+        - libblas-dev
+        - liblapack-dev
+        - gfortran
+        - libspatialindex-dev
+        - swig
+os:
+  - linux
+
+cache:
+  directories:
+    - $HOME/install/firedrake/
+    - $HOME/ffc_kernels/
+
+env:
+  global:
+    - CC=mpicc
+    - FIREDRAKE_FFC_KERNEL_CACHE_DIR=$HOME/ffc_kernels/
+    - PETSC_CONFIGURE_OPTIONS="--download-mumps --download-scalapack --download-parmetis --download-metis"
+
+before_install:
+  - pip install -U pip
+  - pip install -U virtualenv
+
+install:
+  - mkdir -p $HOME/install
+  - pushd $HOME/install
+  - curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install
+  # Check for cached install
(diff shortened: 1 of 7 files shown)


wence-:
Currently not passing. Tester is being killed, maybe OOM?

tkarna:
OOM is possible, the 3d tests require quite a lot of memory. How much memory is available? We could try to reduce memory footprint by allocating only the functions that are used in each test instead of the whole 3d model. Also the high order functions used to compute L2 errors hog a lot of memory.

tkarna:
How can I test for lint? Is it just pylint?

dham:
The basic lint infrastructure is in https://github.com/tkarna/cofs/tree/lint

wence-:
Builders have max 4GB of RAM.  Running the continuity3d tests uses a peak of ~8GB on my local machine, so that is probably the issue.

tkarna:
OK, makes sense. I'll try to reduce the memory footprint of continuity3d test.

wence-:
We can also mark some of the tests as not to be run on travis (I will push a commit for this momentarily)

tkarna:
This looks good to me. If the current fixture runs we can merge.

wence-:
It's getting there slowly.  I think I have fixed the timeout problems, am waiting for a green travis light.

tkarna:
This version e6ca0f033dd797b82c61d5faa30b25f90563d6bd of continuity3d test consumes "only" 2.7GB and might run on travis.

wence-:
I spotted another memory leak problem (that is really only an issue running the test suite, rather than individual runs).  The complicated forms that we build lead to huuuuge Kernels (their in memory pickled representation can easily be 40MB).  These stick around until the end of time (end of the python process) in memory.  And every test makes more of them.  I suspect this is one of the reasons travis is unhappy.  My most recent commit hackily works around this.  Hopefully it will green-light! (https://travis-ci.org/wence-/cofs/builds/97280058)

wence-:
Ok, that build ran through fine in a little under an hour. One convergence failure in tracerEq/test_steady_adv_diff_mms.py which could I guess be an actual problem. Unless it only happens in my branch and I borked something.

tkarna:
OK great! tracerEq/test_steady_adv_diff_mms.py passes on master, so something may be broken. Nevertheless I'd suggest that you skip that test so we can merge this. I'll go through the failing tests separately.

wence-:
Ah, I see what I did.  I moved the clearing of the linProblemCache and tmpFunctionCache to a teardown method that is run once per test.  But it seems that for the convergence tests they still need to be cleared in the refinement. I've put that back in and things now pass.

I'll clean this up, then I think it's good to go.

wence-:
OK, I think this is now fine.  I have some test failures, but I see those on master too, AFAICT.

tkarna:
Ah, that makes sense. OK, I'll go ahead then and merge.

wence-:
To actually get testing running, you need to manually add the appropriate travis web hook.  Go to travis-ci.org and log in with your github account, then in your account, select the cofs repository and click the "test" switch.

tkarna:
Thanks, travis appears to be up and running.  ( ^_^) b

PR #10, opened 2016-02-01
Ip viscosity pep8

stephankramer:
Implements Interior Penalty viscosity scheme for discontinuous velocity fields. Also adds (optionally) grad-div term, and makes the -grad(H)/H*stress term optional. Defaults are the same as before, i.e. only include div-grad term but do include -grad(H)/H term. The opposite choice for both, is more standard for SWE coastal models with some sort of turbulence scheme.

Test added to swe2d/test_steady_state_basin_mms.py

Any opinions on option names? Not sure tensor_form (for "div-grad only") is all that standard, so maybe use_div_grad_only_viscosity is better - or have a include_grad_div_viscosity_term instead?

### DIFF
--- a/cofs/options.py
+++ b/cofs/options.py
@@ -29,6 +29,10 @@ class ModelOptions(AttrDict):
         """bool: Apply log layer bottom stress"""
         self.use_parabolic_viscosity = False
         """bool: Compute parabolic eddy viscosity"""
+        self.use_tensor_form_viscosity = True
+        """bool: Use tensor form instead of symmetric partial stress form"""
+        self.use_grad_depth_term_viscosity_2d = True
+        """bool: Include grad(H)-term in the depth averated viscosity"""
         self.use_ale_moving_mesh = True
         """bool: 3D mesh tracks free surface"""
         self.use_mode_split = True
--- a/cofs/shallowwater_eq.py
+++ b/cofs/shallowwater_eq.py
@@ -379,9 +379,9 @@ class ShallowWaterEquations(Equation):
                 funcs = self.bnd_functions.get(bnd_marker)
                 ds_bnd = self.ds(int(bnd_marker))
                 if funcs is not None:
-                    h = self.bathymetry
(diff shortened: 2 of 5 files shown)


tkarna:
Is the travis failure just a timeout?

stephankramer:
> What is the factor 2 in the non-tensor form?

Hm, I guess it depends what you compare with. sym() has a half in its definition - so my definition is really div[nu ((grad u)+(grad u)^T)] which for constant nu gives nu [nabla^2 u+grad(div(u)] and so the second term drops out for divergence free flow, giving the expected result.

> I  used similar formulation in 3D but without the factor 2 to match with analytical solution.

Not sure - where does the analytical solution come from?

tkarna:
OK, that makes sense. I used a test case where the initial condition for u was a heaviside step. When you apply constant diffusion it turns into an erf function which is easy to check. I'll double the check the test case and get back to you.

tkarna:
About the options, maybe `include_grad_div_viscosity_term` is the cleanest?

wence-:
`with_grad_div_viscosity` ?

tkarna:
Sure.

tkarna:
OK, I've ported the test case to 2D, see
https://gist.github.com/tkarna/f57043f79a88839bdf13

Setting `use_tensor_form_viscosity = False` seems to diffuse the front a bit too fast, and the solution is closer to the analytical if you drop the factor 2 from the definition of the stress. I don't know if that's expected or not... `use_tensor_form_viscosity = True` gives solution that matches well with the analytical.

stephankramer:
Ah I see now - your analytical solution is not divergence free. So the full viscosity term is:

  d/dx_i \nu (d/dx_i u_i + d/dx_j u_i)

Only if \nu is constant we can rewrite this into

  \nu \nabla^2 u + \nu grad(div(u))

Thus for incompressible flow the second term drops out and you get the same as 'tensor form' viscosity.

PR #11, opened 2016-02-01
Pndg pnplus1

stephankramer:
Implements functionality for PnDG-P(n+1) family (SWE2D only at the moment). Changes are fairly minimal, only needed to add a case for "not integrating pressure gradient by parts".

Test added by running swe2d/test_steady_state_basin_mms.py with P1DG-P2 as well.

Again not really convinced of the options name (use_continuous_pressure). Any better ideas? Maybe we should have just one switch that offers mimetic/equal-order-dg/mixed-dg-cg ?

### DIFF
--- a/cofs/options.py
+++ b/cofs/options.py
@@ -19,6 +19,8 @@ class ModelOptions(AttrDict):
         """int: Polynomial degree of elements"""
         self.mimetic = True
         """bool: Use mimetic elements for uvw instead of DG"""
+        self.continuous_pressure = False
+        """bool: Use PnDG-P(n+1) velocity, pressure pair"""
         self.nonlin = True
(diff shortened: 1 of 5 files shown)


tkarna:
Looks good. I agree that it would be clearer to have just a single switch. We could have `options.element_family` that could be `'rt-dg'`, `'equal-order-dg'` or `'mixed-dg-cg'` for example.

By the way, should we use `order` or `degree` for the degree of the basis functions? I think firedrake uses degree which might be more accurate terminology.

tkarna:
I think we can merge this. Implementing `options.element_family` or the like requires changes in the 3D model as well so I'd do that in a separate branch.

PR #13, opened 2016-02-05
No export option

tkarna:
Test cases should not create any files.

### DIFF
--- a/test/bottomFriction/test_implicit_friction_parabolic.py
+++ b/test/bottomFriction/test_implicit_friction_parabolic.py
@@ -9,10 +9,10 @@ from cofs.momentum_eq import VerticalMomentumEquation
 import time as time_mod
 
 
-def test_implicit_friction_parabolic(do_assert=True):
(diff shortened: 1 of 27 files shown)


test/bottomFriction/test_implicit_friction_parabolic.py
```diff
@@ -56,6 +56,7 @@ def test_implicit_friction_parabolic(do_assert=True):
     options.t_export = t_export
     options.dt = dt
     options.t_end = t_end
+    options.no_exports = ~do_export
```
wence-: Idiomatically one would write `not do_exports`

wence-:
Looks sane

test/bottomFriction/test_implicit_friction_parabolic.py
tkarna: Yes you are right.

tkarna:
Verified that py.test does not create any files on my laptop, should be good to go.

PR #14, opened 2016-02-10
Rename OuterProductElement -> TensorProductElement

wence-:
This renaming has now happened in firedrake, so do the same here.


tkarna:
Looks familiar, merging.

PR #15, opened 2016-02-11
rename OuterProductElement to TensorProductElement

tkarna:
Rename to catch up with current firedrake master.


wence-:
Oh, snap in #14

tkarna:
Ah missed that :/

PR #20, opened 2016-02-22
MixedFunctionSpace type no longer user exposed

wence-:
Change required when firedrakeproject/firedrake#713 is merged.

### DIFF
--- a/thetis/exporter.py
+++ b/thetis/exporter.py
@@ -4,6 +4,7 @@ Routines for handling file exports.
 Tuomas Karna 2015-07-06
 """
 from utility import *
+import ufl
 
 
 class ExporterBase(object):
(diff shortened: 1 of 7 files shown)


wence-:
Oops, still wrong. Will fix tomorrow

tkarna:
I think I got the isinstance check working now.

tkarna:
These checks in `momentum_eq.py` are still broken
```
        self.horizontal_dg = ufl_elem._A.family() != 'Lagrange'
        self.vertical_dg = ufl_elem._B.family() != 'Lagrange'
        self.HDiv = ufl_elem._A.family() == 'Raviart-Thomas'
```

wence-:
OK, I'll fix those up now.

wence-:
OK, so these checks are pretty fragile.

For the horizontal_dg and vertical_dg checks, I think the right thing to do is to make the fiat element and check if dofs are only associated with the cell in, respectively, the horizontal and vertical directions.  For the HDiv check, I guess one should check that the outside element is an HDivElement.

wence-:
I've tried to fix the cases where the code was doing checking based on the element family.  I think it should now be more robust (and hopefully work!).

tkarna:
Great, looks good to me!

Issue #5, opened 2015-12-17
LinearProblemCache (and temp Function Cache) are somewhat icky


wence-:
I think I understand why these are here, however, they can cause some slightly strange side effects.

I think the function cache is easier to address.  Effectively you're using this for "work vectors".  When you want a temporary function, you ask for one given a particular function space.

I think we could add something like this to Firedrake.  Here's how I would do it.

The FunctionSpace object gains a `get_function(name=None)`, `restore_function(function)` pair.  You can then get a temporary function that you can use:

```python
f = V.get_function()
... do stuff with f
V.restore_function(f)
```

We can wrap this up in a context manger so one could do:

```python
with V.work_space() as f:
    ... do stuff with f
```

This is neater, because if the only reference to the work space is inside V, once V gets collected, so do the temporary functions.

The linear problem cache is a little trickier, however, I think it is still possible to fix.  The entries in the problem cache become invalid when the mesh they were built on top of go out of scope.  So we could make these collectable by adding a level on top using a `WeakKeyDictionary`.  Sketch:

```python
import weakref
class ProblemCache(object):

    def __init__(self):
        self._cache = weakref.WeakKeyDictionary()

    def __getitem__(self, key):
        mesh, key = key
        return self._cache[mesh][key]

    def __setitem__(self, key, val):
        mesh, key = key
        d = self._cache.get(mesh, {})
        d[key] = val
        self._cache[mesh] = d

    def __contains__(self, key):
        mesh, key = key
        return mesh in self._cache and self._cache[mesh].get(key) is not None

    def clear(self):
        self._cache.clear()
```

So the idea is to use a weakref to the Mesh as a first-level key, the second level acts as before.  Now if the Mesh goes out of scope, the cached problems are no longer strongly referenced by the cache and can be collected.

wence-:
Humph, that doesn't work I think, because the value in the dict holds a strong reference to the key and so it's never collected.

tkarna:
I agree, both of these dicts were implemented as temporary fixes to issues I had.

I use temporary functions essentially to save memory. I think storing them in the function space directly is a great idea. Why is `restore_function` necessary?

tkarna:
As for the LinearProblemCache, I think instead of caching them, another solution would be to implement classes that hold the solver objects. For functions like `computeVertVelocity` that certainly makes sense. So essentially all those functions in `utility` would be implemented as objects that are stored in the solver class:

```python
[...]

PR #21, opened 2016-02-24
Remove temp function cache

tkarna:
Addressing #5 remove hacky `tmp_function_cache` and use `FunctionSpace.get_work_function` instead.


tkarna:
Also removed homemade `Projector` class because firedrake has one now.

wence-:
Rebased now that firedrakeproject/firedrake#719 is merged.  Also added an assertion that the work function cache is of size 1, to avoid weird later errors.  Will merge assuming a green light appears.

PR #23, opened 2016-03-11
Extrude upwards

tkarna:
3d mesh used to be extruded downwards from free surface to bottom which had some nasty side effects. For example, the top/bottom faces in firedrake corresponded to bottom/top faces in thetis. It also affected the sign of the jacobian of 3d elements etc.

This changes the extrusion direction to be from bottom to top, achievable with fairly minor changes to the code.


tkarna:
Also updated test/examples a little.

PR #27, opened 2016-03-25
New VTK outputs

tkarna:
Updates related to new firedrake VTK output.

### DIFF
--- a/thetis/exporter.py
+++ b/thetis/exporter.py
@@ -63,7 +63,7 @@ class VTKExporter(ExporterBase):
         # ensure correct output function name
         old_name = tmp_proj_func.name()
         tmp_proj_func.rename(name=self.func_name)
-        self.outfile << (tmp_proj_func, self.next_export_ix)
+        self.outfile.write(tmp_proj_func, time=self.next_export_ix)
         self.next_export_ix += 1
         # restore old name
         tmp_proj_func.rename(name=old_name)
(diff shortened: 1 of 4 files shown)


tkarna:
This changes outputs to p1dg for discontinuous fields, p1 otherwise. The visualization spaces and projection/interpolation is still handled in Thetis as Firedrake's `File` object allocates it's own output functions which can be problematic in 3d.

Discontinuous outputs seem to work OK, I'm still seeing a modest increase in memory usage at run time, but it doesn't seem to be associated with the exports. (~6 months ago exporting dg fields used to cause some mysterious memory leak).

thetis/exporter.py
```diff
@@ -5,6 +5,23 @@
...
+    """
+    Figure out the appropriate linear visualization space for fs
+    """
+    is_vector = len(fs.ufl_element().value_shape()) == 1
+    mesh = fs.mesh()
+    family = 'Lagrange' if is_cg(fs) else 'Discontinuous Lagrange'
+    if is_vector:
+        visu_fs = VectorFunctionSpace(mesh, family, 1)
+    else:
+        visu_fs = FunctionSpace(mesh, family, 1)
```
tkarna: Most of the function spaces created here already exist in the Thetis solver object. I presume that function spaces are cached somehow, and this does not actually create any new instances?

wence-: Yes, the heavyweight data is shared between function space instances. So you're paying the cost of a new Python object, but not new dof maps, etc...

tkarna: ok good.

tkarna:
Merging this version to master to fix broken exports.

PR #29, opened 2016-04-13
Slope limiter boundary treatment

tkarna:
Extending slope limiters to work correctly at boundaries

### DIFF
--- a/thetis/limiter.py
+++ b/thetis/limiter.py
@@ -141,24 +141,36 @@ class VertexBasedP1DGLimiter(object):
         if not self.is_2d:
             # Add nodal values from surface/bottom boundaries
             # NOTE calling firedrake par_loop with measure=ds_t raises an error
-            kernel = op2.Kernel("""
-                void my_kernel(double **qmax, double **qmin, double **centroids) {
-                    for (int i=0; i<%(nodes)d; i++) {
-                        qmax[i][0] = fmax(qmax[i][0], centroids[i][0]);
-                        qmin[i][0] = fmin(qmin[i][0], centroids[i][0]);
+            bottom_nodes = self.P1CG.bt_masks['geometric'][0]
+            top_nodes = self.P1CG.bt_masks['geometric'][1]
+            bottom_idx = op2.Global(len(bottom_nodes), bottom_nodes, dtype=np.int32, name='node_idx')
+            top_idx = op2.Global(len(top_nodes), top_nodes, dtype=np.int32, name='node_idx')
+            code = """
+                void my_kernel(double **qmax, double **qmin, double **centroids, int *idx) {
+                    double face_mean = 0;
+                    for (int i=0; i<%(nnodes)d; i++) {
+                        face_mean += centroids[idx[i]][0];
                     }
-                }""" % {'nodes': self.max_field.cell_node_map().arity}, 'my_kernel')
+                    face_mean /= %(nnodes)d;
+                    for (int i=0; i<%(nnodes)d; i++) {
+                        qmax[idx[i]][0] = fmax(qmax[idx[i]][0], face_mean);
+                        qmin[idx[i]][0] = fmin(qmin[idx[i]][0], face_mean);
+                    }
+                }"""
+            kernel = op2.Kernel(code % {'nnodes': len(bottom_nodes)}, 'my_kernel')
 
             op2.par_loop(kernel, self.mesh.cell_set,
                          self.max_field.dat(op2.WRITE, self.max_field.function_space().cell_node_map()),
                          self.min_field.dat(op2.WRITE, self.min_field.function_space().cell_node_map()),
                          field.dat(op2.READ, field.function_space().cell_node_map()),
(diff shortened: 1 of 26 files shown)


tkarna:
Added first implementation where the exterior facets are taken into account when computing the allowed min/max values. A couple of remarks:

When determining the allowed min/max values, this implementation uses the _original field_ on the boundary facets. This means that the solution on the boundary facets is not limited at all. This is probably not what we want. We should therefore get a p0 solution on the facets somehow. For triangles one could compute the facet average in pyop2 if we knew the correct facet node indices. For quads the facet node average is not exactly the p0 solution but it might be good enough.

Second, If I try to use measure `ds_t` or `ds_b` in firedrake `par_loop` I get an error:
```
  File "/home/tuomas/sources/firedrake/src/firedrake/firedrake/parloops.py", line 210, in par_loop
    _map = _maps[measure.integral_type()]
KeyError: 'exterior_facet_top'
```
I therefore implemented the top/bottom loop with `op2.par_loop`.

tkarna:
OK I think I got the top/bottom surfaces fixed using `bt_masks`. The lateral boundaries are still wrong.

thetis/limiter.py
```diff
@@ -125,6 +127,52 @@ def _update_min_max_values(self, field):
...
+                qmin[i][0] = fmin(qmin[i][0], field[i][0]);
+            }""",
+                 ds,
+                 {'qmax': (self.max_field, RW),
+                  'qmin': (self.min_field, RW),
+                  'field': (field, READ)})
+
+        if not self.is_2d:
+            # Add nodal values from surface/bottom boundaries
+            # NOTE calling firedrake par_loop with measure=ds_t raises an error
```
wence-: This requires an extension in firedrake, see firedrakeproject/firedrake#751

test/firedrake/test_diverence_2d.py
```diff
@@ -0,0 +1,142 @@
+"""
```
wence-: Typo in test file name

wence-:
UFL has symbolic operators for `FacetAvg` and `CellAvg`.  We've never had a use case for them, so have not implemented them.  However, we plausibly could.

I think we have a map from exterior facets to the nodes that are (topologically or geometrically) on the facet.  However, it's only really used internally and I think it doesn't fit the bill.

You can do it like this, I think:
```python

boundary_dofs = facet_support_dofs(V.fiat_element) # use vert_facet_support_dofs in extruded
local_facet_nodes = numpy.array(
            [boundary_dofs[e] for e in sorted(boundary_dofs.keys())])

kernel = """
void my_kernel(double **qmax, double **qmin, double **field, unsigned int *facet)
{
    local_facet_idx = 2d_array_of_local_facet_nodes;
   
    for (i = 0; i < nnode; i++ ) {
        idx = local_facet_idx[facet[0]][i];
        qmax[idx][0] = fmax(qmax[idx][0], field[idx][0]);
        ...
    }
}"""

op2.par_loop(my_kernel, V.mesh().exterior_facets.set, qmax.dat(op2.RW, qmax.exterior_facet_node_map()), qmin.dat(...), field.dat(...), V.mesh().exterior_facets.local_facet_dat(op2.READ))
```

Untested!  This assumes all the fields are in the same space.  If you have different spaces you'll need to encode the local_facet_idx array for each of them.

tkarna:
Thanks. I got this far, just trying to write to the max field using the indexing. But it segfaults.

```python
from FIAT.finite_element import facet_support_dofs
from FIAT.tensor_product import horiz_facet_support_dofs, vert_facet_support_dofs
import numpy

if self.is_2d:
    boundary_dofs = facet_support_dofs(self.P1CG.fiat_element)
    local_facet_nodes = numpy.array([boundary_dofs[e] for e in sorted(boundary_dofs.keys())])
    local_facet_idx = op2.Global(local_facet_nodes.shape, local_facet_nodes, dtype=np.int32, name='node_idx')

    code = """
    void my_kernel(double **qmax, double **qmin, double **field, unsigned int *facet, int **local_facet_idx)
    {
        for (int i = 0; i < %(nnodes)d; i++) {
            int idx = local_facet_idx[facet[0]][i];
            qmax[idx][0] = 4.5;
        }
    }"""
    facet_kernel = op2.Kernel(code % {'nnodes': local_facet_nodes.shape[0]}, 'my_kernel')

    op2.par_loop(facet_kernel,
                    self.P1CG.mesh().exterior_facets.set,
                    self.max_field.dat(op2.RW, self.max_field.exterior_facet_node_map()),
                    self.min_field.dat(op2.RW, self.min_field.exterior_facet_node_map()),
                    field.dat(op2.READ, field.exterior_facet_node_map()),
                    self.P1CG.mesh().exterior_facets.local_facet_dat(op2.READ),
                    local_facet_idx(op2.READ)
                    )
```

tkarna:
For the time being I've omitted the lateral boundary treatment. I'd like to merge this version as the turbulence branch depends on it. We can revisit this issue this once migrated to new equation api.

PR #30, opened 2016-04-15
Ip viscosity thetis

stephankramer:
This is a follow up from #10. Implements Interior Penalty viscosity scheme for the depth-integrated equations.

### DIFF
--- a/thetis/shallowwater_eq.py
+++ b/thetis/shallowwater_eq.py
@@ -33,7 +33,9 @@ class ShallowWaterEquations(Equation):
                  mu_manning=None, lin_drag=None, baroc_head=None,
                  coriolis=None, wind_stress=None, uv_lax_friedrichs=None,
                  uv_source=None, elev_source=None,
-                 nonlin=True):
+                 nonlin=True,
+                 use_tensor_form_viscosity=True,
+                 use_grad_depth_term_viscosity_2d=True):
         self.space = solution.function_space()
         self.mesh = self.space.mesh()
         self.U_space, self.eta_space = self.space.split()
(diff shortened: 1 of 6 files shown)


thetis/shallowwater_eq.py
```diff
@@ -342,6 +346,65 @@ def rhs_implicit(self, solution, wind_stress=None,
...
+            stress_jump = avg(nu)*2.*sym(tensor_jump(uv, n))
+        else:
+            stress = nu*grad(uv)
+            stress_jump = avg(nu)*tensor_jump(uv, n)
+
+        f = inner(grad(self.U_test), stress)*dx
+
+        if self.u_is_dg:
+            # projection to DG0 is only necessary because of a bug that's now fixed on master firedrake:
+            h = project(CellSize(self.mesh), FunctionSpace(self.mesh, "DG", 0))
```
tkarna: Is this projection still necessary?

thetis/shallowwater_eq.py
```diff
@@ -342,6 +346,65 @@ def rhs_implicit(self, solution, wind_stress=None,
...
+            # from Epshteyn et al. 2007 (http://dx.doi.org/10.1016/j.cam.2006.08.029)
+            # the scheme is stable for alpha > 3*X*p*(p+1)*cot(theta), where X is the
+            # maximum ratio of viscosity within a triangle, p the degree, and theta
+            # with X=2, theta=6: cot(theta)~10, 3*X*cot(theta)~60
+            p = self.U_space.ufl_element().degree()
+            alpha = 60.*p*(p+1)
```
tkarna: I've used a smaller value `alpha = 3.5*p*(p+1)`. I presume that some of your test cases require such a high penalty?

tkarna:
This looks all right to me.

stephankramer:
Thanks! The projection of CellSize() is indeed no longer needed. The motivation for the value of alpha is in the comments above it - so your findings might get worse if you use a (rapidly) varying viscosity, or anisotropic meshes. I have reduced the value however as it seems to work indeed for cases I'm running now. If this becomes an issue, I may have a look in making it spatially varying (I think the bound still applies if you base it on local viscosity contrast and minimum angle).

PR #31, opened 2016-04-19
Viscosity 3D

tkarna:
Reimplemented 3D diffusion operators with interior penalty method. Added MES tests against analytical solution.

### DIFF
--- a/thetis/momentum_eq.py
+++ b/thetis/momentum_eq.py
@@ -19,6 +19,7 @@ class MomentumEquation(Equation):
                  viscosity_v=None, viscosity_h=None,
                  coriolis=None, source=None,
                  baroc_head=None,
+                 v_elem_size=None, h_elem_size=None,
                  lax_friedrichs_factor=None, uv_mag=None,
                  uv_p1=None,
                  nonlin=True):
@@ -26,6 +27,8 @@ class MomentumEquation(Equation):
         self.mesh = self.space.mesh()
         self.nonlin = nonlin
         self.solution = solution
+        self.v_elem_size = v_elem_size
+        self.h_elem_size = h_elem_size
         # this dict holds all time dep. args to the equation
         self.kwargs = {'eta': eta,
                        'w': w,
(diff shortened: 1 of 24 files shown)


thetis/momentum_eq.py
```diff
@@ -297,31 +300,80 @@ def rhs(self, solution, eta, w=None, viscosity_v=None,
...
+                                              self.normal[1]**2) +
+                            self.v_elem_size*self.normal[2]**2)
+                sigma = 3.5*degree_h*(degree_h + 1)/elemsize
+                if degree_h == 0:
+                    raise NotImplementedError('horizontal visc not implemented for p0')
+                alpha = avg(sigma)
+                ds_interior = (dS_h + dS_v)
+                # oriented diffusive flux n.K.n = nu*(n_xy.n_xy)
+                # Pestiaux (2014)  DOI: 10.1002/fld.3900
+                ip_flux_jump = avg(dot(dot(self.normal, visc_tensor), self.normal))*jump(solution)
```
stephankramer: I don't think this is the correct extension from a scalar diffusion scheme to vectorial viscosity. The term you implement gives;

    (v_+ . n_+ + v_- . n_-) * (u_+ . n_+ + u_- . n_-)

where v and u are the vectorial test and trial functions. This is the same as:

     (v_+ - v_-) . n_+n_+ . (u_+ - u_-)

This indicates that for a facet with normal direction in the x-direction (n=(1,0,0)) you only get a contribution for the u-component. Moreover for facet normals that are not Cartesian aligned the n_+n_+ tensor has off-diagonal components which indicates coupling between the components.

What you want however is a diffusion operator that acts independently on each velocity component (and is the same linear operator for each). This is obtained by:

  (v_+ n_+ + v_- n_-) : (u_+ n_+ + v_- n_-)

where : indicates the inner product that contracts the index of the v and u components, and the index of the normal components.

The way I think about it is that the jump in u is associated with an integration by parts of grad(u). grad(u) is a tensor however so the integration by parts gives a boundary term with the tensor outer product u n.

The same applies to the second line. I think that should be:

    f += inner(outer_jump(solution, self.normal), avg(grad_test)) * ds_interior

That's obviously ignoring the viscosity tensor so far. With viscosity I think the terms should be:

    f += alpha*inner(outer_jump(n, self.test), dot(avg(visc_tensor), outer_jump(self.normal, solution)))*ds_interior
    f += -inner(avg(dot(visc_tensor, nabla_grad(self.test))), outer_jump(self.normal, solution))*ds_interior
    f += -inner(outer_jump(self.normal, self.test), avg(dot(visc_tensor, nabla_grad(solution))))*ds_interior

Note that:
* I've swapped the normal and the solution in the outer_jump terms, and used nabla_grad instead of grad so that the first index is always the normal/derivative and the second is the velocity component. This is to ensure we contract with the correct index (the first) when contracting with the viscosity tensor
* If we contract with the viscosity tensor correctly, I don't think we need a grad_h (or nabla_grad_h) because the viscosity tensor will automatically project out the vertical direction.
* In all the terms we only contract the self.test index with the solution index, meaning that we indeed get the same operator that acts independently on each component.
* There's some subtleties with regards to the averaging of viscosity (does not matter for continuous viscosity) depending on the choice of the auxilary variable. I've here followed Pestiaux.

Please check my maths :-)

tkarna: Thanks a lot for checking this. Your formulations look legit, I'll go through them more carefully and the implementation.

tkarna:
@stephankramer I think your formulation is fine. What always confuses me is that Epshteyn and Riviere (2007) do not have the normal in their definition of the jump operator.  But if you do the math, I think what you have written is equivalent to their formulation. I also checked that this is equivalent to the scalar case.

I have now updated the horizontal/vertical diffusion/viscosity operators to be consistent with each other.
In the horizontal case, I do not get the optimal convergence rate in the tests against a smooth analytical solution, even with finer meshes. For the time being I have increased the tolerance for the convergence rate, so the tests pass.

PR #32, opened 2016-04-19
decrease exporter memory footprint

tkarna:
This is mostly relevant to 3D runs where memory consumption is a concern

- exporters use function spaces allocated in solver class
- ExportManager has dg coordinate function to avoid allocating one in File

### DIFF
--- a/thetis/exporter.py
+++ b/thetis/exporter.py
@@ -8,17 +8,20 @@ import ufl
 from firedrake.output import is_cg
 
 
-def get_visu_space(fs):
+def is_2d(fs):
+    """Tests wether a function space is 2d or 3d"""
+    return fs.mesh().geometric_dimension() == 2
+
+
+def get_visu_space(fs, function_space_pool):
     """
     Figure out the appropriate linear visualization space for fs
     """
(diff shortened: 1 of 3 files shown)


thetis/exporter.py
```diff
@@ -8,17 +8,20 @@
...
     """
     Figure out the appropriate linear visualization space for fs
     """
     is_vector = len(fs.ufl_element().value_shape()) == 1
-    mesh = fs.mesh()
-    family = 'Lagrange' if is_cg(fs) else 'Discontinuous Lagrange'
-    if is_vector:
-        visu_fs = VectorFunctionSpace(mesh, family, 1)
-    else:
-        visu_fs = FunctionSpace(mesh, family, 1)
```
wence-: Is the function space pool really necessary?  I had hoped that making the same function space again only pays the cost of building a python object (that may be a few kb max).

wence-: Or is this just "belt and braces"?

tkarna: It actually is not needed - I've removed it now.

PR #35, opened 2016-04-25
New equation API

tkarna:
Implementation of new equation API, implemented on top of #34.

There's a slight modification to the plan: terms and equations know the function space of the solution and boundary information
```python
t = Term(function_space, boundary_markers, boundary_len)
```
and provide some members (e.g. `term.test`, `term.normal`) that are useful when implementing the equations.


tkarna:
- At the moment both old and new equations co-exist for debugging purposes
  - new equations are named `TracerEquationNew` etc
  - and time integrators are `CrankNicolsonNew` etc
  - old formulations should be removed before merging
- New formulation should be exactly equivalent to the old one
  - linearization and treatment of time levels in time integrators have not been changed
  - for example nearly all terms in shallow water eqns. are still fully nonlinear
  - these should be fixed (after merging this?)
- Not all time integrators are ported yet
- `EquationNew.get_jacobian` method is in the base class but not used in any application

thetis/equation.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    def __init__(self, function_space, boundary_markers, boundary_len):
+        # define bunch of members needed to construct forms
+        self.function_space = function_space
+        self.mesh = self.function_space.mesh()
+        self.test = TestFunction(self.function_space)
+        self.tri = TrialFunction(self.function_space)
+        self.normal = FacetNormal(self.mesh)
+        # TODO construct them here from mesh ?
+        self.boundary_markers = boundary_markers
```
wence-: You can get the complete set of boundary markers on the mesh from: `mesh.exterior_facets.unique_markers`.

thetis/momentum_eq.py
```diff
@@ -480,3 +521,379 @@ def source(self, w=None, viscosity_v=None,
...
+    """
+    Generic term for momentum equation that provides commonly used members and
+    mapping for boundary functions.
+    """
+    def __init__(self, function_space, boundary_markers, boundary_len,
+                 bathymetry=None, v_elem_size=None, h_elem_size=None,
+                 nonlin=True, use_bottom_friction=False):
+        super(MomentumTerm, self).__init__(function_space, boundary_markers, boundary_len)
+        self.boundary_markers = boundary_markers
+        self.boundary_len = boundary_len
```
wence-: The superclass constructor already set these, no?

thetis/equation.py
```diff
@@ -0,0 +1,153 @@
...
+        self.boundary_len = boundary_len
+
+    def get_residual(self, solution, solution_old, fields, fields_old, bnd_conditions):
```
wence-: So as mentioned in the meeting, one could do away with the `get_` prefix for these and just use `def residual(...)`.  Plausibly this is bikeshedding a little.

thetis/timeintegrator.py
```diff
@@ -566,6 +834,73 @@ def advance_linear(self, t, dt, solution, update_forcings):
...
+        bnd = bnd_conditions
+        f = self.fields
+        f_old = self.fields_old
+
+        # Crank-Nicolson
+        gamma_const = Constant(gamma)
+        # FIXME this is consistent with previous implementation but time levels are incorrect
+        self.F = (self.equation.mass_term(u) - self.equation.mass_term(u_old) -
+                  self.dt_const*(gamma_const*self.equation.get_residual('all', u, u, f, f, bnd) +
+                                 (1-gamma_const)*self.equation.get_residual('all', u_old, u_old, f_old, f_old, bnd)
```
wence-: So if you were using the same field values in `f` and `f_old` this could be simpler, you would just write:

```python
u = gamma_const*u + (1 - gamma_const)*u_old
self.equation.get_residual('all', u, u, f_old, f_old, bnd)
```
This gives you a residual that has C-N splitting for the solution and lags all the fields.  If one wants to use old and new field values as well as implicit solution values in the lagged `u`, we might want to think about redefining get_residual to take something other than a dict of fields.

thetis/timeintegrator.py
```diff
@@ -786,6 +1121,144 @@ def get_final_solution(self, output=None):
...
+        self.dt = dt
+        self.dt_const = Constant(dt)
+        self.solution_old = solution
+
+        test = self.equation.test
+
+        from firedrake.functionspaceimpl import MixedFunctionSpace, WithGeometry
+        mixed_space = (isinstance(fs, MixedFunctionSpace) or
+                       (isinstance(fs, WithGeometry) and
+                        isinstance(fs.topological, MixedFunctionSpace)))
```
wence-: You can use `len(fs) > 1` instead for this check if you like.  Alternately, you can always use `fs.topological`, even if `fs` is not a `WithGeometry` (in which case you have `fs.topological is fs` returning `True`).

thetis/timeintegrator.py
```diff
@@ -804,6 +1277,27 @@ def __init__(self, equation, dt, solver_parameters={}, terms_to_add='all'):
...
+                 bnd_conditions=None, solver_parameters={}):
+        a = [[1.0]]
+        b = [1.0]
+        c = [1.0]
+        super(BackwardEulerNew, self).__init__(equation, solution, fields, dt,
+                                               a, b, c,
+                                               bnd_conditions=bnd_conditions,
+                                               solver_parameters=solver_parameters)
+
+
```
wence-: The only thing that these dirk classes do is define the tableau.  You can simplify the code like this:

```python

class DIRKGeneric(object):
    
    def __init__(self, equation, solution, fields, dt, bnd_conditions=None, solver_parameters={})
        self.n_stages = len(self.b)
        ...


class BackwardEuler(DIRKGeneric):
    a = [[1.0]]
    b = [1.0]
    c = [1.0]

class DIRKLSPUM2(DIRKGeneric):
     a = [[2.0/11.0, 0, 0],
          [205.0/462.0, 2.0/11.0, 0],
          [2033.0/4620.0, 21.0/110.0, 2.0/11.0]]
     b = [24.0/55.0, 1.0/5.0, 4.0/11.0]
     c = [2.0/11.0, 289.0/462.0, 751.0/924.0]
```

IOW, you can set the Butcher coefficients as class attributes in the concrete classes and just use them directly generic class.  If you wanted to be really fancy you could require that the subclasses set those coefficients like so:

```python
from abc import ABCMeta, abstractproperty

class DIRK(object):
    __metaclass__ = ABCMeta

    @abstractproperty
    def a(self):
        pass

    @abstractproperty
    def b(self):
        pass

    @abstractproperty
    def c(self):
        pass

class MyDIRK(DIRK):
    a = ...
    b = ...
    c = ...
```

Now any class derived from `DIRK` cannot be instantiated unless it implements all of `a`, `b` and `c`.

thetis/timeintegrator.py
stephankramer: Even assuming f==f_old, this would give a different answer for nonlinear terms. Not necessarily saying one is better than the other...

thetis/equation.py
tkarna: Yes that's cleaner, I'll change that.

thetis/momentum_eq.py
tkarna: That's correct.

thetis/equation.py
tkarna: I'll change that.

thetis/timeintegrator.py
wence-: That's true.  I see standard C-N takes a weighted sum of the residuals rather than the residual with the solution replaced by a weighted sum.

thetis/equation.py
wence-: So in the meeting the concrete suggestion for these things (and boundary_len and similar) is just to splat them onto the mesh you make:

```python

mesh = ...

mesh.boundary_len = ...

# much later

blen = V.mesh().boundary_len

```

etc...

thetis/timeintegrator.py
tkarna: Yes I think we need to revisit these time integrators once this pr has landed. Currently I have just reimplemented the existing formulation so we can check that the results are exactly the same. 

Here for instance I think it's better to keep the time lagged residual in memory instead of doing that hacky fields_old update.

thetis/timeintegrator.py
tkarna: I see, that's definitely shorter.

thetis/timeintegrator.py
tkarna: Sweet, I'll update the classes.

tkarna:
I have now implemented all your suggestions, removed the old formulations and cleaned up some obsolete time integration bits. Travis was timing out so I skipped one bottom friction test also. If we get green light this is good to go?

stephankramer:
All looks good to me now. Happy for it to be merged.

thetis/shallowwater_eq.py
```diff
@@ -370,10 +307,10 @@ def horizontal_viscosity(self, uv, nu, total_h):
 
             # Dirichlet bcs only for DG
             for bnd_marker in self.boundary_markers:
-                funcs = self.bnd_functions.get(bnd_marker)
+                funcs = bnd_conditions.get(bnd_marker)
                 ds_bnd = ds(int(bnd_marker))
                 if funcs is not None:
-                    eta_ext, uv_ext = self.get_bnd_functions(None, uv, bnd_marker)
+                    eta_ext, uv_ext = self.get_bnd_functions(eta, uv, bnd_marker, bnd_conditions)
```
tkarna: @stephankramer Note I had to change this: Some boundary conditions actually need elevation, and passing None will raise an error.

stephankramer:
Ah, thanks for the heads up.  We need to take that into account indeed when resolving the conflict with my fix for 'free slip' I pushed to master.

tkarna:
Yes, I'll rebase this branch and fix it.

tkarna:
Test passed, hence merging.

Issue #28, opened 2016-04-09
Memory consumption


tkarna:
Memory consumption seems to grow boundlessly with large simulations. Here's an example of the idealized estuary example (turbulence branch):

```
$pidstat -h -r -p 18846 30
Linux 4.2.0-35-generic (hyakutake)      04/09/2016      _x86_64_        (4 CPU)

#      Time   UID       PID  minflt/s  majflt/s     VSZ    RSS   %MEM  Command
 1460219513  1000     18846    514.13      0.00 2312504 1374720   8.42  python
 1460219543  1000     18846    368.17      0.07 2534136 1592508   9.75  python
 1460219573  1000     18846    399.67      0.00 2787808 1846216  11.30  python
 1460219603  1000     18846    310.50      0.00 2970008 2028532  12.42  python
 1460219633  1000     18846    183.33      0.00 3077760 2136284  13.08  python
 1460219663  1000     18846      0.00      0.00 3077760 2136284  13.08  python
 1460219693  1000     18846      1.30      0.00 3077760 2136388  13.08  python
 1460219723  1000     18846      0.00      0.00 3077760 2136388  13.08  python
 1460219753  1000     18846      8.73      0.00 3107268 2165872  13.26  python
 1460219783  1000     18846     98.73      0.00 3262180 2320784  14.21  python
 1460219813  1000     18846    306.13      0.00 3469808 2528008  15.47  python
 1460219843  1000     18846    363.87      0.00 3715588 2773976  16.98  python
 1460219873  1000     18846    338.30      0.00 3903300 2961496  18.13  python
 1460219903  1000     18846    380.43      0.00 4152840 3210848  19.65  python
 1460219933  1000     18846    298.73      0.00 4333568 3391808  20.76  python
 1460219963  1000     18846    318.77      0.00 4555468 3613784  22.12  python
 1460219993  1000     18846    354.37      0.00 4772844 3831080  23.45  python
 1460220023  1000     18846    276.20      0.00 4951260 4009200  24.54  python
 1460220053  1000     18846    353.80      0.00 5171488 4229532  25.89  python
 1460220083  1000     18846    365.97      0.00 5356540 4414732  27.02  python
 1460220113  1000     18846    258.50      0.00 5567644 4625752  28.32  python
 1460220143  1000     18846    386.77      0.00 5787396 4845760  29.66  python
 1460220173  1000     18846    383.00      0.00 6004996 5063300  30.99  python
[...]

PR #36, opened 2016-05-06
Add temperature equation and equation of state

tkarna:
Adds temperature equation and density field as a diagnostic variable. Also fixes a bug in turbulence model.


examples/lockExchange/runTestSuite.py
```diff
@@ -145,12 +143,13 @@
...
+temp_init3d = Function(solver_obj.function_spaces.H, name='initial temperature')
 # vertical barrier
-# salt_init3d.interpolate(Expression(('(x[0] > 0.0) ? 20.0 : 25.0')))
+# temp_init3d.interpolate(Expression('(x[0] > 0.0) ? v_l : v_r',
+#                                    v_l=T_left, v_r=T_right))
 # smooth condition
-salt_init3d.interpolate(Expression('22.5 - 2.5*tanh(x[0]/sigma)',
-                                   sigma=1000.0))
+temp_init3d.interpolate(Expression('v_l - (v_l - v_r)*0.5*(tanh(x[0]/sigma) + 1.0)',
+                                   sigma=1000.0, v_l=temp_left, v_r=temp_right))
```
wence-: FWIW, I recommend moving to interpolating using [UFL expressions](http://firedrakeproject.org/interpolation.html#ufl-expressions) rather than C strings.  That way you get nice type-checking if you mistype things.

thetis/coupled_timeintegrator.py
```diff
@@ -493,8 +491,8 @@ def __init__(self, solver):
             explicit_v_diff = solver.tot_v_diff.get_sum()
 
         vdiff_sp = {'snes_monitor': False,
-                    'ksp_type': 'gmres',
-                    'pc_type': 'ilu',
+                    # 'ksp_type': 'gmres',
+                    # 'pc_type': 'ilu',
```
wence-: We could revisit these bits next week, but if you want ilu but safe in parallel use `'pc_type': 'bjacobi', 'sub_pc_type': 'ilu'`.  Which does process-block jacobi and ilu on the blocks.

thetis/solver.py
```diff
@@ -618,3 +650,4 @@ def iterate(self, update_forcings=None, update_forcings3d=None,
                         print '{0:25s} : {1:11.6f} {2:11.2f}'.format(
                             label, c, relcost)
                         sys.stdout.flush()
+                gc.collect()
```
wence-: I guess this is related to the discussion in #28?

tkarna: Yes, simulations eventually die on stampede as the nodes run out of memory. AFAIK invoking garbage collector manually doesn't really help. At least now I'm able to restart the interrupted runs.

examples/lockExchange/runTestSuite.py
tkarna: OK, good to know.

thetis/solver.py
wence-: I would really like to track this down, as it must be an issue *somewhere*.  However, in the past I have had absolutely zero luck getting to the bottom of this kind of thing.

PR #37, opened 2016-05-06
Enable restart

tkarna:
Simulations can be restarted with `FlowSolver.load_state(i_export)`. This feature was partially supported in the case of 2D model, now works with 3D model as well. Depends on #36


thetis/exporter.py
```diff
@@ -82,7 +83,9 @@ def __init__(self, fs_visu, func_name, outputdir, filename,
     def set_next_export_ix(self, next_export_ix):
         """Sets the index of next export"""
         # NOTE vtk io objects store current export index not next
-        super(VTKExporter, self).set_next_export_ix(next_export_ix - 1)
+        super(VTKExporter, self).set_next_export_ix(next_export_ix)
+        # FIXME hack to change correct output file count
+        self.outfile.counter = itertools.count(start=self.next_export_ix)
```
tkarna: This is hijacks `File.counter` as a workaround to set custom start index. See firedrakeproject/firedrake#772

PR #46, opened 2016-05-13
Improve performance

tkarna:
Set quadrature rule degree to 2p + 1 to avoid overly high degrees (automatically estimated degree can be up to 15). Also sets better solver option for most of the solvers.


thetis/coupled_timeintegrator.py
```diff
@@ -69,7 +69,7 @@ def _update_mesh_velocity(self):
     def _update_baroclinicity(self):
         """Computes baroclinic head"""
         if self.options.baroclinic:
-            with timed_stage('aux_barolinicity'):
+            with timed_stage('aux_barolin'):
```
wence-: Baroclin?

thetis/timeintegrator.py
```diff
@@ -89,13 +89,13 @@ def update_solver(self):
...
-        self.solver_k0 = LinearVariationalSolver(prob_k0,
+        self.solver_k0 = LinearVariationalSolver(prob_k0, options_prefix=self.name + '_k0',
                                                  solver_parameters=self.solver_parameters)
         prob_k1 = LinearVariationalProblem(self.a_rk, self.L_RK1, self.K1)
-        self.solver_k1 = LinearVariationalSolver(prob_k1,
+        self.solver_k1 = LinearVariationalSolver(prob_k1, options_prefix=self.name + '_k1',
                                                  solver_parameters=self.solver_parameters)
         prob_k2 = LinearVariationalProblem(self.a_rk, self.L_RK2, self.K2)
-        self.solver_k2 = LinearVariationalSolver(prob_k2,
+        self.solver_k2 = LinearVariationalSolver(prob_k2, options_prefix=self.name + '_k2',
```
wence-: a_rk is just a mass matrix I think, and can therefore be reused over the stages. If so we should use a LinearSolver which we can solve three times with the different right hand sides

thetis/timeintegrator.py
```diff
@@ -210,13 +210,13 @@ def update_solver(self):
...
-        self.solver_k0 = LinearVariationalSolver(prob_k0,
+        self.solver_k0 = LinearVariationalSolver(prob_k0, options_prefix=self.name + '_k0',
                                                  solver_parameters=self.solver_parameters)
         prob_k1 = LinearVariationalProblem(self.a_rk, self.L_RK, self.K1)
-        self.solver_k1 = LinearVariationalSolver(prob_k1,
+        self.solver_k1 = LinearVariationalSolver(prob_k1, options_prefix=self.name + '_k1',
                                                  solver_parameters=self.solver_parameters)
         prob_k2 = LinearVariationalProblem(self.a_rk, self.L_RK, self.K2)
-        self.solver_k2 = LinearVariationalSolver(prob_k2,
+        self.solver_k2 = LinearVariationalSolver(prob_k2, options_prefix=self.name + '_k2',
```
wence-: Similarly here I think

thetis/coupled_timeintegrator.py
```diff
@@ -104,11 +104,15 @@ def _update_all_dependencies(self, t,
...
-            if do_vert_diffusion and self.options.solve_vert_diffusion:
+        if do_vert_diffusion and self.options.solve_vert_diffusion:
+            with timed_stage('impl_mom_vvisc'):
                 self.timestepper_mom_vdff_3d.advance(t, self.solver.dt, self.fields.uv_3d)
-                if self.options.solve_salt:
+            if self.options.solve_salt:
+                with timed_stage('impl_salt_vdiff'):
                     self.timestepper_salt_vdff_3d.advance(t, self.solver.dt, self.fields.salt_3d)
+            if self.options.solve_temp:
+                with timed_stage('impl_salt_vdiff'):
```
stephankramer: Should that be impl_temp_vdiff ?

tkarna: yes!

PR #52, opened 2016-06-15
Update for communicator changes in Firedrake

wence-:
Rather than there always being a global communicator, we now build
meshes by passing in a communicator.  All subsequent operations are then
collective only over that communicator.  This defaults to COMM_WORLD.


tkarna:
Thanks, this looks all right. Travis errors seem to be due to obsolete firedrake, do we need to blow the cache or such?

wence-:
Probably.  The PR is rebuilding firedrake so let's see how that goes.

PR #54, opened 2016-07-01
Add element family option

tkarna:
Currently supported options are `option.element_family = 'dg-dg'`, `'rt-dg'` and `'dg-cg'`. 
Replaces `options.mimetic` and `options.continuous_pressure`.


stephankramer:
Looks all good. One query: I don't really object but you seem to have changed the default from "rt-dg" to "dg-dg"; Any particular reason?

tkarna:
Good call, I'll revert it back to `'rt-dg'` for consistency.

PR #55, opened 2016-07-04
Generic callbacks

tkarna:
New implementation of callback functions.

### DIFF
--- a/test/callback/test_diagnostic_hdf5_output.py
+++ b/test/callback/test_diagnostic_hdf5_output.py
@@ -4,10 +4,16 @@ Tests diagnostic callbacks and hdf file output.
 from thetis import *
 from thetis.callback import VolumeConservation3DCallback
 import h5py
-import shutil
+import pytest
 
 
-def test_callbacks():
+@pytest.fixture(scope='session')
+def tmp_outputdir(tmpdir_factory):
+    fn = tmpdir_factory.mktemp('outputs')
+    return str(fn)
+
+
+def test_callbacks(tmp_outputdir):
 
     lx = 45000.0
     ly = 3000.0
@@ -20,8 +26,8 @@ def test_callbacks():
     # estimate of max advective velocity used to estimate time step
     u_mag = Constant(0.5)
 
-    outputdir = 'outputs'
-    print_info('Exporting to '+outputdir)
+    outputdir = tmp_outputdir
+    print_info('Exporting to ' + outputdir)
 
     # bathymetry
     p1_2d = FunctionSpace(mesh2d, 'CG', 1)
@@ -81,7 +87,7 @@ def test_callbacks():
         def __call__(self, solver_obj):
             value = self.const_val
             c = Constant(self.const_val,
-                         domain=solver_obj.mesh.coordinates.ufl_domain())
+                         domain=solver_obj.mesh)
             integral = assemble(c*dx)
             return value, integral
 
@@ -145,10 +151,6 @@ def test_callbacks():
         assert np.allclose(reldiff, correct_reldiff)
         assert np.allclose(integral, correct_integral)
 
-    # remove all output files
-    print_info('Removing dir: {:}'.format(solver_obj.options.outputdir))
-    shutil.rmtree(solver_obj.options.outputdir)
-
 
 if __name__ == '__main__':
-    test_callbacks()
+    test_callbacks('outputs')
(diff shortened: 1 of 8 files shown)


thetis/callback.py
```diff
@@ -5,35 +5,152 @@
...
+    """Creates hdf5 files for diagnostic time series arrays."""
+    def __init__(self, filename, varnames, comm=COMM_WORLD):
+        self.comm = comm
+        self.filename = filename
+        self.varnames = varnames
+        self.nvars = len(varnames)
+        if comm.rank == 0:
+            # create empty file with correct datasets
+            hdf5file = h5py.File(filename, 'w')
+            dataarrays = {}
```
tkarna: It was used in my first implementation, not really needed anymore. I'll remove it.

tkarna:
Cleaned up the interface a little. It's now possible to do

```python
solver_obj.add_callback(VolumeConservation3DCallback())
```

test/callback/test_diagnostic_hdf5_output.py
```diff
@@ -0,0 +1,154 @@
...
+                     append_to_log=True):
+            super(ConstCallback, self).__init__(outputdir,
+                                                export_to_hdf5,
+                                                append_to_log)
+            self.const_val = const_val
+
+        def __call__(self, solver_obj):
+            value = self.const_val
+            c = Constant(self.const_val,
+                         domain=solver_obj.mesh.coordinates.ufl_domain())
```
wence-: It is always true that `mesh is mesh.coordinates.ufl_domain()`

test/callback/test_diagnostic_hdf5_output.py
```diff
@@ -0,0 +1,154 @@
...
+    nx = 25
+    ny = 2
+    mesh2d = RectangleMesh(nx, ny, lx, ly)
+    depth = 50.0
+    elev_amp = 1.0
+    n_layers = 6
+    # estimate of max advective velocity used to estimate time step
+    u_mag = Constant(0.5)
+
+    outputdir = 'outputs'
```
wence-: You can use the built in pytest fixture `tmpdir` to get a temporary directory that is unique per test and is automatically cleaned up by pytest for output.

thetis/callback.py
```diff
@@ -5,35 +5,159 @@
...
-        """Initializes the callback function.
+    def add(self, callback, mode):
+        key = callback.name
+        self.callbacks[mode][key] = callback
 
-        :kwarg comm: The communicator for the callback."""
-        self._initialized = True
+    def evaluate(self, solver_obj, mode):
+        for key in self.callbacks[mode]:
+            self.callbacks[mode][key].evaluate(solver_obj)
```
wence-: I you don't want to have to specify the set of `modes` up front, you could do:

```python
from collections import defaultdict

class CallbackManager(object):
    def __init__(self):
        self.callbacks = defaultdict(OrderedDict)
        super(CallbackManager, self).__init__()
```

wence-: Also, I would then inherit from `defaultdict`:

```python
class CallbackManager(defaultdict):
    def __init__(self):
        super(CallbackManager, self).__init__(OrderedDict)
```

That way you don't need to implement `__getitem__`.

test/callback/test_diagnostic_hdf5_output.py
tkarna: I need to explicitly set the domain to avoid error `UFLException: This integral is missing an integration domain.`

test/callback/test_diagnostic_hdf5_output.py
tkarna: Nice, I'll add that.

thetis/callback.py
tkarna: Sounds good, I'll change that.

test/callback/test_diagnostic_hdf5_output.py
wence-: Oh, I meant you can just use `solver_obj.mesh` as the domain argument

tkarna: I see. Fixed.

tkarna:
Good to go?

thetis/callback.py
```diff
@@ -5,35 +5,155 @@
...
+    def _create_hdf5_file(self, solver_obj):
+        """
+        Creates an empty hdf5 file with correct datasets.
+        """
+        if self.append_to_hdf5:
+            comm = solver_obj.comm
+            outputdir = self.outputdir
+            if outputdir is None:
+                outputdir = solver_obj.options.outputdir
+            create_directory(outputdir)
```
wence-: Probably want `comm=comm` here.

wence-:
Other than my minor comment looks good, will fix here and merge.

PR #58, opened 2016-07-13
Logging

tkarna:
Use python logger for model output and errors.

Implements two loggers:

- one for generic model output messages, issued with `print_output`. This is at `INFO` level. Will only print the message with no header:

```
>>> print_output('some output')
some output
```
- one for debug, warning, error etc messages, issued with `debug`, `warning`, `error`, ... Prints log level as a header:

```
>>> warning('some warning')
WARNING some warning
```

Sits on top of #55 .


wence-:
Looks good, will merge.

PR #60, opened 2016-07-20
Drone CI configuration

wence-:
Adds a configuration file for drone (hooked up with http://jenkins.ese.ic.ac.uk:8000/thetisproject/thetis).  Let's see if it works!

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,63 @@
+workspace:
+  base: /drone
+  path: src
+
+pipeline:
+  build:
+    image: ubuntu:xenial
+    pull: true
+    environment:
+      # Change the following to reflect your local preferences
+      - PACKAGE=thetis
+      - THETIS_HOME=/drone/install
+      - THETIS_GID=1000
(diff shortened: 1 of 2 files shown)


wence-:
Sort of works, but it seems to default timeout of 1 hour on drone is not enough...

wence-:
Now that the drone install we have here is 0.5, everything seems much nicer.  So I've ported this forward and merged it.  The full test suite runs including build in a little under an hour, e.g. http://jenkins.ese.ic.ac.uk:8000/thetisproject/thetis/12 and doesn't use any caching (so is "safer" than travis).  The default build timeout is 100 minutes, but I can up this if required.

tkarna:
Looks great! Should we start flagging tests separately for travis and drone? Currently we haven't been able to add more tests as travis will time out.

wence-:
I think we should explicitly mark the tests we want to run on travis.  And by default run everything on drone.

wence-:
At the moment we mark `not_travis` instead.

PR #62, opened 2016-07-25
Lock exhange updates

tkarna:
Updates for lock exchange test

- linear equation of state
- command line interface for setting options
- uses callbacks to compute diagnostics at run time


examples/lockExchange/diagnostics.py
```diff
@@ -0,0 +1,134 @@
...
+        if not hasattr(self, '_initialized') or self._initialized is False:
+            self._initialize(solver_obj)
+        # compute x center of mass on top/bottom surfaces
+        self.extractor_bot.solve()
+        x_bot = self.compute_front_location()
+        self.extractor_surf.solve()
+        x_top = self.compute_front_location()
+        return x_bot, x_top
+
+    def __str__(self, args):
```
stephankramer: This is more a query about the DiagnosticCallback class I guess, but I'm a little confused about these extra `args` to the __str__() method. I've never seen this patttern before - does this not break `str(callback)` ?

wence-: Yes, I would make them `*args`.  I think the way `__str__` is used here, it's always called as `self.__str__(args)`.

examples/lockExchange/diagnostics.py
```diff
@@ -0,0 +1,134 @@
...
+        self.rho_ho.dat.data[ix] = 1
+        mass = assemble(self.rho_ho*dx)
+        if mass < 1e-20:
+            return np.nan
+        center_x = assemble(self.rho_ho*self.xyz[0]*dx)/mass
+        return center_x
+
+    def __call__(self, solver_obj):
+        if not hasattr(self, '_initialized') or self._initialized is False:
+            self._initialize(solver_obj)
```
stephankramer: This sort of pattern breaks if you call the object with two different solver_obj. If the object is meant to permanently store (some aspects of) solver_obj, would it not be better to overload its __init__() and have solver_obj as its argument?

thetis/solver.py
```diff
@@ -71,29 +71,75 @@ def __init__(self, mesh2d, bathymetry_2d, n_layers,
...
+
+    def set_time_step(self, alpha=0.05):
         comm = self.comm
         if self.options.use_mode_split:
             self.dt = self.options.dt
             if self.dt is None:
-                mesh_dt = self.eq_sw.get_time_step_advection(u_mag=self.options.u_advection)
-                dt = self.options.cfl_3d*float(np.floor(mesh_dt.dat.data.min()/20.0))
+                mesh_dt = self.compute_time_step_3d(u_mag=self.options.u_advection)
+                dt = self.options.cfl_3d*alpha*float(np.floor(mesh_dt.dat.data.min()))
```
stephankramer: What's the meaning of alpha here?
Why do we floor mesh_dt? What happens if mesh_dt<1? `numpy.floor` already returns a `float`

examples/lockExchange/diagnostics.py
wence-: I am inclined to agree.  I don't recall why we went for this setup, other than you might want to build the diagnostic objects before you have the solver available and/or fully setup.  One should probably check if already initialised that the provided solver is the same as the first one.  That could be done in the superclass `__call__`.  WDYT?  I can do this if yes.

tkarna: Yes, I wanted to avoid storing the solver in the class as a property (solver has a dict of callbacks, who all have solver as a property), but I guess it would be cleaner. In that case `__call__` takes no arguments.

thetis/solver.py
tkarna: So alpha is a dt factor that depends on the time integration scheme and spatial discretization, here it's just implemented as a hacky number. In the ALE time integrator branch alpha is actually computed: all the time integrators define a CFL number and the element families have a scaling factor. The floor operator is a remnant of the past, it is removed in the ALE branch. I'd suggest that we leave this bit as it is because it will be revised in the (hopefully near) future.

stephankramer:
Happy for this to be merged btw.

examples/lockExchange/diagnostics.py
tkarna: Yes, I'd call this function something else than `__str__` to avoid confusion. `message_str(self, *args)`?

tkarna:
Going to merge this and address the callback changes in another pr.

PR #63, opened 2016-07-25
Viscosity for DG0/RT spaces

tkarna:
Updated diffusion operators to handle DG0/RT spaces. DG0 uses the same IP term implementation, only a different penalty parameter value.

Sits on top of #62.


stephankramer:
Looks all good to me

Issue #33, opened 2016-04-20
Lateral boundaries in slope limiter


tkarna:
#29 fixes the top/bottom conditions for p1dg slope limiters but lateral boundaries are still broken. That is, the elements at boundaries only see the interior neighbors which often leads to too strong limiting.

PR #65, opened 2016-07-26
limiter: Subclass new Firedrake VertexBasedLimiter

wence-:
Now that some limiting functionality is in Firedrake itself, remove some
code in thetis.  Currently we just have to override the compute_bounds
function to also compute min/max at top and bottom.

This is in part to open a discussion...


tkarna:
This looks all right to me. Per our earlier discussion I gather that

- We will not try to fix the lateral boundary issue #33 in thetis but rather wait for firedrake's limiter to be generalized
- Moving ALE mesh requires an update mechanism for the limiter centroid solvers. This issue will however also go away once limiters use UFL cell/face averages.

PR #66, opened 2016-08-10
Callback update

tkarna:
So here's a draft what the callback api could look like.

Changes:

- solver object is now given to `DiagnosticCallback` constructor and stored in the class
- thus `__call__` and `evaluate` do not take any arguments
- renamed `__str__` as `message_str`


tkarna:
Ah, this appears to be on #63 , so only the last commit is relevant.

stephankramer:
Nice. Looks good

PR #67, opened 2016-08-11
Pressure projection

stephankramer:
New timestepper for 2D equations that combines a pressure projection approach with Picard iterations. Solves a preliminary (depth-averaged) separate momentum equation first (using the free surface elevation from the previous timestep or Picard iteration). Then solves the coupled system containing only the wave equations terms (mass terms and free surface gradient and continuity divergence term). These two solves can be iterated with a Picard iteration to increase the coupling between the free surface update and the full momentum equation, and in case this is combined with the use_linearized_semi_implicit_2d option (recommended) to reduce the error due to linearizing the advection and continuity divergence terms. With nonlinear terms that are merely quadratic, the second order accuracy of Crank Nicolson can thus be restored with 2 Picard iterations.

If use_linearized_semi_implicit_2d is combined with DG velocity (dg-dg or dg-cg), the mass term is easily inverted using preonly+ilu and thus the coupled linearized wave equations can be cheaply solved by solving the Schur complement exactly, i.e. no further iterations of the outer solver on the coupled system are necessary. For rt-dg, preonly+jacobi is a pretty good approximation to the inverse rt mass matrix. Further speedup can be achieved (this is still on a branch) by assembling the Schur complement explicitly as a matrix.

### DIFF
--- a/examples/stommel2d/stommel2d_picard.py
+++ b/examples/stommel2d/stommel2d_picard.py
@@ -1,0 +1,92 @@
+# Stommel gyre test case in 2D
+# ============================
+#
+# Wind-driven geostrophic gyre in larege basin.
+# Setup is according to [1].
+#
+# [1] Comblen, R., Lambrechts, J., Remacle, J.-F., and Legat, V. (2010).
+#     Practical evaluation of five partly discontinuous finite element pairs
+#     for the non-conservative shallow water equations. International Journal
+#     for Numerical Methods in Fluids, 63(6):701-724.
+#
+# Tuomas Karna 2015-04-28
+
+# This is a version that runs much faster using much larger timesteps (2 hr instead of 45s.)
+# using an implicit time integration scheme (PressureProjectionPicard)
+
+from thetis import *
+
+lx = 1.0e6
+nx = 20
+mesh2d = RectangleMesh(nx, nx, lx, lx)
+outputdir = 'outputs_picard'
+print_output('Loaded mesh '+mesh2d.name)
+print_output('Exporting to '+outputdir)
+depth = 1000.0
+t_end = 75*12*2*3600
+t_export = 3600*2
+
+# bathymetry
+P1_2d = FunctionSpace(mesh2d, 'CG', 1)
+P1v_2d = VectorFunctionSpace(mesh2d, 'CG', 1)
+bathymetry_2d = Function(P1_2d, name='Bathymetry')
+bathymetry_2d.assign(depth)
+
+# Coriolis forcing
+coriolis_2d = Function(P1_2d)
+f0, beta = 1.0e-4, 2.0e-11
+coriolis_2d.interpolate(
+    Expression('f0+beta*(x[1]-y_0)', f0=f0, beta=beta, y_0=0.0))
+
+# Wind stress
+wind_stress_2d = Function(P1v_2d, name='wind stress')
+tau_max = 0.1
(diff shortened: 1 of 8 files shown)


examples/stommel2d/stommel2d_picard.py
```diff
@@ -0,0 +1,87 @@
+# Stommel gyre test case in 2D
+# ============================
+#
+# Wind-driven geostrophic gyre in larege basin.
```
tkarna: typo: large

test/swe2d/test_standing_wave.py
```diff
@@ -0,0 +1,104 @@
...
+    elev_init = cos(pi*x[0]/lx)
+
+    # bathymetry
+    p1_2d = FunctionSpace(mesh2d, 'CG', 1)
+    bathymetry_2d = Function(p1_2d, name="bathymetry")
+    bathymetry_2d.assign(depth)
+
+    # --- create solver ---
+    solver_obj = solver2d.FlowSolver2d(mesh2d, bathymetry_2d, order=1)
+    solver_obj.options.nonlin = True
```
tkarna: If `nonlin=True` is this really a standing wave? Advection will deform the waves over time, no?

thetis/timeintegrator.py
```diff
@@ -562,6 +562,133 @@ def advance(self, t, dt, solution, update_forcings=None):
...
+        super(PressureProjectionPicard, self).__init__(equation, solver_parameters)
+        self.equation_mom = equation_mom
+        self.solver_parameters_mom = solver_parameters_mom
+        if semi_implicit:
+            # solve a preliminary linearized momentum equation before
+            # solving the linearized wave equation terms in a coupled system
+            self.solver_parameters.setdefault('snes_type', 'ksponly')
+            self.solver_parameters_mom.setdefault('snes_type', 'ksponly')
+        else:
+            # not sure this combination makes much sense: keep both systems nonline
```
tkarna: typo

tkarna:
Sweet, looks all right to me. Glad to see that such solvers can be implemented in a pretty clean manner.

test/swe2d/test_standing_wave.py
```diff
@@ -0,0 +1,104 @@
...
+            'snes_type': 'ksponly',  # we've linearized, so no snes needed
+            'ksp_type': 'preonly',  # we solve the full schur complement exactly, so no need for outer krylov
+            'pc_type': 'fieldsplit',
+            'pc_fieldsplit_type': 'schur',
+            'pc_fieldsplit_schur_fact_type': 'full',
+            'pc_fieldsplit_schur_precondition': 'selfp',
+            # velocity mass block:
+            'fieldsplit_0_ksp_type': 'preonly',  # NOTE: this is only an exact solver for the velocity mass block if velocity is DG
+            'fieldsplit_0_pc_type': 'ilu',
+            'fieldsplit_1_ksp_type': 'fgmres',
```
wence-: This guy can be `gmres`.

wence-: Also, if you need this many iterations you may want to either bump the restart length, or use modified gram schmidt orthogonalisation (more expensive, but more numerically stable)

test/swe2d/test_standing_wave.py
stephankramer: Yes, indeed. However with a small amplitude (wrt depth) within a single period it should remain very very close to the linear solution (see also comments on top). Not really testing the nonlinear terms therefore (just that the split doesn't mess things up).

PR #69, opened 2016-08-25
New firedrake matrix types

wence-:
Update to new way of specifying assembled matrix type in firedrake.  The requisite change in dolfin-adjoint will hopefully merge imminently.  That side of things also fixes remembering the assembled matrix type (so no need to specify it globally in the adjoint tests).


wence-:
Dolfin-adjoint changes now merged.

Issue #68, opened 2016-08-15
TimeIntegrator API update


tkarna:
I think we could simplify the time integrator API a little:
1. Omit solution function from `TimeIntegrator.advance` call
   
   Currently `TimeIntegrator.advance` takes the solution function as an argument. This is redundant as in practice `TimeIntegrator` needs to store the solution function as a property in any case. It's also difficult to ensure meaningful operation in case the user happens to pass a different function to the `advance` method.
2. Omit `dt` from `TimeIntegrator.advance` + add `set_dt` method
   
   We normally use a constant time step so there's no need to pass the time step to every call. In case the user wants to change the time step, I would add `set_dt` method instead. Then one could make sure that the time step gets updated correctly. In cases where changing time step is not allowed (e.g. linear multistep methods) we'd raise an exception.

Here's a draft of the base classes

``` python
class TimeIntegratorBase(object):
    """Base class for all time integrator objects."""
    __metaclass__ = ABCMeta

    @abstractproperty
    def advance(self, t, update_forcings=None):
        """Advances equations for one time step."""
        pass

    @abstractproperty
    def initialize(self, *args, **kwargs):
        """Initializes the time integrator"""
        pass


class TimeIntegrator(TimeIntegratorBase):
    """Base class for all time integrator objects that solve a single equation"""
    def __init__(self, equation, solution, fields, dt, solver_parameters={}):
        super(TimeIntegrator, self).__init__()

        self.equation = equation
        self.solution = solution
        self.fields = fields
        self.dt_const = Constant(dt)

        # unique identifier for solver
        self.name = '-'.join([self.__class__.__name__,
                              self.equation.__class__.__name__])
        self.solver_parameters = {}
        self.solver_parameters.update(solver_parameters)

    def set_dt(self, dt):
        """Updates time step"""
        self.dt_const.assign(dt)
```

Here I'm separating `TimeIntegratorBase` and `TimeIntegrator` classes: All time integrators are derived from the former (e.g. 3D coupled time integrators) while all "normal" time integrators (that iterate a single equation) inherit the latter.

Here I haven't changed the signature for `update_forcings` function which is still passed to the `advance` method. There might be a better way of handling that too.

stephankramer:
Yes, both things seem very sensible.

tkarna:
Now implemented in #70

PR #70, opened 2016-10-25
Timeintegration ale

tkarna:
Updates related to time integrators and the ALE moving mesh:
- implements new TimeIntegrator API outlined in #68 
- all Runge-Kutta schemes are moved to `rungekutta.py` and inherit from an abstract base class `AbstractRKScheme`
- Runge-Kutta classes are defined with multiple inheritage:

``` python
class SSPRK33(ERKGenericShuOsher, SSPRK33Abstract):
    pass
```
- `FlowSolver` object computes max. stable time step for advection/diffusion terms
- mesh movement is handled by single `ALEMeshUpdater` class
- implements leap frog ALE time integrator
- adds `option.log_output` that dumps all `print_output` messages to `outputdir/log` file


tkarna:
Any comments on these?
[timeintegrator.py](https://github.com/thetisproject/thetis/blob/96d833a4ad5308aa4b57ae25421035663a4afa17/thetis/timeintegrator.py)
[rungekutta.py](https://github.com/thetisproject/thetis/blob/96d833a4ad5308aa4b57ae25421035663a4afa17/thetis/rungekutta.py)
[implicitexplicit.py](
https://github.com/thetisproject/thetis/blob/96d833a4ad5308aa4b57ae25421035663a4afa17/thetis/implicitexplicit.py)

The rest is just updated machinery.

thetis/timeintegrator.py
```diff
@@ -695,342 +324,96 @@ def advance(self, t, dt, solution, updateForcings=None):
...
+        (MT)_{n+1} = (MT)_{n} + dt*L_{n+1/2}
 
-    def advance(self, t, dt, solution, update_forcings=None):
+        This is Euler ALE step: LHS is evaluated in Omega_{n+1}, RHS in Omega_n
+        """
+        if self._nontrivial:
+            # NOTE must call eval_rhs in the old mesh first
+            self.rhs_func += self.msolution_old
+            assemble(self.a, self.mass_matrix)
+            solve(self.mass_matrix, self.solution, self.rhs_func)
```
wence-: If the mass matrix is DG, I think you're better off (performance-wise) directly assembling the inverse:

```python
assemble(self.a, self.mass_matrix, inverse=True)
with self.solution.dat.vec as x:
    with self.rhs_func.dat.vec_ro as b:
        self.mass_matrix.force_evaluation()
        self.mass_matrix.petscmat.mult(b, x)
```

The same in the `predict` step too.

It's not obvious why the mass_matrix changes between steps (I guess due to outside ALE?  Maybe mention that?)

thetis/rungekutta.py
```diff
@@ -0,0 +1,949 @@
...
+    def solve_stage(self, i_stage, t, update_forcings=None):
+        self.update_solution(i_stage)
+        self.solve_tendency(i_stage, t, update_forcings)
+        if i_stage == self.n_stages - 1:
+            self.get_final_solution()
+
+
+class ERKSemiImplicitGeneric(RungeKuttaTimeIntegrator):
+    """
+    Generic implementation of explicit RK schemes.
```
wence-: semi-implicit?

thetis/rungekutta.py
```diff
@@ -0,0 +1,949 @@
...
+
+            # solve tendency
+            self.solver.solve()
+
+            # solve the next internal solution
+            # BUG starting with sol_sum = 0 results in different solution!
+            sol_sum = float(self.beta[i_stage + 1][i_stage])*self.tendency
+            for j in range(i_stage + 1):
+                sol_sum += float(self.alpha[i_stage + 1][j])*self.stage_sol[j]
+            self.solution.assign(sol_sum)
```
wence-: So if you write:

```python

sol_sum = 0
for j in range(...):
    sol_sum += ...
solution.assign(sol_sum)
```

You get a different answer?

wence-: I think I know what's going on (maybe).  This kind of formulation is intermingling symbolic computation and not.  UFL does some simplification of `a*b` if either of `a` or `b` are scalar values (or a symbolic zero).  In particular, if `a == 1`, then `1*b => b`.  In this case you assign the coefficient `self.tendency` to `sol_sum` and each intermediate `+=` now does the increment numerically into `self.tendency`, which is somewhat unsafe?

Moreover, you make more `par_loops` than necessary.

I would rewrite this as:

```python
import operator
sol_expr = reduce(operator.add, map(operator.mul, self.alpha[i_stage + 1], self.stage_sol), self.beta[i_stage+1][i_stage]*self.tendency)
self.solution.assign(sol_expr)
```

Or, if you don't like the functional style:

```python
sol_sum = float(self.beta[i_stage + 1][i_stage])*self.tendency
for j in range(i_stage + 1):
    sol_sum = sol_sum + float(self.alpha[i_stage + 1][j])*self.stage_sol[j]
self.solution.assign(sol_sum)
```
This is not equivalent to your code, because `__iadd__` is overridden differently from `__add__`.

thetis/rungekutta.py
```diff
@@ -0,0 +1,949 @@
...
+        Updates solution to i_stage sub-stage.
+
+        Tendencies must have been evaluated first.
+        """
+        if self._nontrivial:
+            # construct full form: L = c*dt*F + M_n*sol_n + ...
+            # everything is assembled, just sum up functions
+            self.l_form.assign(self.msol_old)
+            for j in range(i_stage):
+                self.l_form += float(self.a[i_stage][j])*self.stage_mk[j]
```
wence-: Here I would write:

```python
self.l_form.assign(self.msol_old + reduce(operator.add, map(operator.mul, self.a[i_stage][:i_stage], self.stage_mk[:i_stage])))
```

So you do one loop, rather than 1 + i_stage loops.

thetis/rungekutta.py
```diff
@@ -0,0 +1,949 @@
...
+        if self._nontrivial:
+            if update_forcings is not None:
+                update_forcings(t + self.c[i_stage]*self.dt)
+            assemble(self.l_rk, self.stage_mk[i_stage])
+
+    def get_final_solution(self):
+        if self._nontrivial:
+            self.l_form.assign(self.msol_old)
+            for j in range(self.n_stages):
+                self.l_form += float(self.b[j])*self.stage_mk[j]
```
wence-: Likewise here.

thetis/rungekutta.py
```diff
@@ -0,0 +1,949 @@
...
+        if not self._initialized:
+            error('Time integrator {:} is not initialized'.format(self.name))
+        if update_forcings is not None:
+            update_forcings(t + self.c[i_stage]*self.dt)
+        self.solver[i_stage].solve()
+
+    def get_final_solution(self):
+        self.solution.assign(self.solution_old)
+        for j in range(self.n_stages):
+            self.solution += self.dt_const*self.b[j]*self.k[j]
```
wence-: Ditto.

thetis/timeintegrator.py
```diff
@@ -8,407 +8,50 @@
...
 
-    CFL coefficient is 1.0
-    """
-    def __init__(self, equation, solution, fields, dt, bnd_conditions=None, solver_parameters={}):
-        """Creates forms for the time integrator"""
-        super(SSPRK33Stage, self).__init__(equation, solver_parameters)
-        self.explicit = True
-        self.CFL_coeff = 1.0
-        self.n_stages = 3
+class TimeIntegrator(TimeIntegratorBase):
```
wence-: It seems like TimeIntegratorBase is an unnecessary class?  You could just make TimeIntegrator the meta class and add the `advance` and `initialise` abstract methods to it?  In passing, given that they are methods, not properties, you probably want to decorate them with `@abstractmethod`, rather than `@abstractproperty` (also importable from `abc`).

thetis/implicitexplicit.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    SSP-IMEX time integration scheme based on [1], method (17).
+
+    [1] Higueras et al (2014). Optimized strong stability preserving IMEX
+        Runge-Kutta methods. Journal of Computational and Applied
+        Mathematics 272(2014) 116-140.
+    """
+    def __init__(self, equation, solution, fields, dt, bnd_conditions=None,
+                 solver_parameters={}, solver_parameters_dirk={}):
+        super(SSPIMEX, self).__init__(equation, solution, fields, dt, solver_parameters)
```
wence-: Add a warning that that this is absolute code (that the user sees)?

wence-:
Other than my minor comments, this looks great!

thetis/timeintegrator.py
tkarna: `TimeIntegratorBase` is in fact the base class for `CoupledTimeIntegrators`. So essentially `TimeIntegrator` is for cases where you only solve a single equation, and provides constructor for it. Naming could be better I guess. I'll change the methods to `@abstractmethod`.

thetis/timeintegrator.py
tkarna: Thanks, that gives ~10% speedup in the lock exchange test. I'll add some documentation to clarify the usage of this time integrator.

tkarna: I'll take that back, the gain is more moderate - assembly seems to dominate the timings.

PR #72, opened 2016-11-15
drone: Enable code coverage reporting

wence-:
for example, see here: https://codecov.io/gh/thetisproject/thetis/tree/1c87bd9a611f0d381b99ba0f02a2ad3ab88b838f


tkarna:
Looks great

tkarna:
This is good to go on my behalf - travis error seems to be time out which is fixed in master.

PR #73, opened 2016-11-18
New mode splitting

tkarna:
Implements more reasonable mode splitting following 
http://dx.doi.org/10.1006/jcph.1997.5733
http://dx.doi.org/10.1006/jcph.2002.7003

The main difference is that the 3D momentum equation is written for the u', i.e. deviation from depth averaged velocity: u' = u - depth_average(u).

Sits on top of #70.


tkarna:
Jenkins still throws strange errors. All tests pass on my computer, I'm going to merge this.

thetis/solver.py
```diff
@@ -727,8 +731,13 @@ def add_callback(self, callback, eval_interval='export'):
 
     def export(self):
         self.callbacks.evaluate(mode='export')
+        # set uv to total uv instead of deviation from depth average
+        # TODO find a cleaner way of doing this ...
```
stephankramer: Yeah, would it not be cleaner to have `uv_3d` always contain the total velocity and have a separate `uv_dev_3d` and pass that into the relevant places. Not sure how much re-engineering that requires...

PR #74, opened 2016-11-18
SSPRK22 time integrator

tkarna:
New time integrator based on SSPRK(2,2) scheme. Formally second order, seems to be more robust than leap-frog-AM3.

Sits on top of #73


stephankramer:
Oh I see, I've just commented on a PR that was merged already! (#73). This looks fine as well

PR #80, opened 2016-12-09
adds slf2gmsh python script, which converts selafin mesh files to gmsh files

slmouradian:
This script converts selafin files generated for the TELEMAC model to gmsh files compatible with Thetis.

### DIFF
--- a/scripts/slf2gmsh.py
+++ b/scripts/slf2gmsh.py
@@ -123,6 +123,7 @@ def add_tags_to_gmsh_from_cli(meshFile, cliFile):
     fixed.close()
     meshFile.close()
 
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description=print_desc())
     parser.add_argument('selafin_file', nargs=1,
--- a/scripts/slf2gmsh.py
+++ b/scripts/slf2gmsh.py
@@ -23,15 +23,16 @@ import logging
 try:
     import meshio
 except ImportError:
-    logging.error( 'Can not import meshio')
-    logging.exception( 'Please try: $ pip install meshio')
+    logging.error('Can not import meshio')
+    logging.exception('Please try: $ pip install meshio')
     sys.exit(1)
 
 try:
     from parsers import parserSELAFIN
 except ImportError:
     logging.error('Can not import selafin parser')
-    logging.error('Have you set your PYTHONPATH to include the selafin parsers?')
+    logging.error('Have you set your PYTHONPATH to' +
+                  'include the selafin parsers?')
     sys.exit(1)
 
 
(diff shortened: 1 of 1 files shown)


scripts/slf2gmsh.py
```diff
@@ -0,0 +1,169 @@
...
+            found = True
+
+        fixed.write(line)
+    fixed.write('\n')
+    fixed.close()
+    meshFile.close()
+
+if __name__ == '__main__':
+    try:
+        infile = sys.argv[1]
```
dham: Please use the `argparse` module for consistency with other Firedrake scripts.

scripts/slf2gmsh.py
```diff
@@ -0,0 +1,169 @@
...
+        clifile = sys.argv[2]
+    except IndexError:
+        print 'ERROR setting source and/or destination files'
```
dham: Please use the module `logging` instead of print statements in production code.

wence-:
Given that this builds on `meshio`, just adding a bit more information?  Would it not make more sense to incorporate those changes there?

slmouradian:
Thanks both for useful feedback!

@dham , I will modify to use argparse and logging. New committ shortly.
As for the thetis/firedrake: This might be useful to many people, including Fluidity / dolfin / or other Firedrake users. However, the TELEMAC model is primarily a *coastal* ocean model, and thus I think the Thetis repo is the most appropriate. Please advise, happy for it to be moved to Firedrake (or removed completely, if you think it's not appropriate).

@wence- There are two reasons I am not adding this to meshio.
a) meshio does not support writing gmsh files with tags (.msh file without tags are perfectly valid gmsh files, according to the gmsh format). The need for tags is specific to DMPlex, and this script adds tags to the output of meshio.
b) I do not want to implement a selafin file reader/writer in meshio; it would be of no use to me or anyone in our group.

wence-:
The mesh reading issue sounds like a bug in the DMPlex reader that we can fix (then you don't need to do as much fixing up).  Can you provide a mesh with no tags?  Equally, do you not think, even if you don't implement the selafin reader in meshio, would it not make sense to allow meshio to export tags?

slmouradian:
@wence- I produced a small mesh with no tags, and it was read with no issues, so I was wrong about DMPlex choking on that.

As for meshio, I have no intention to meddle in their source at this time. I think this simple workaround, although less elegant, works best for now.

wence-:
Looking at the code in the DMPlex gmsh reader, I think there is a bug (memory overrun) for meshes with no tags.  It's just you may not have hit it.

tkarna:
I'm closing this due to inactivity. If you'd want to merge this feature please make a new PR.

PR #83, opened 2016-12-20
Website

wence-:
Migrates source of thetis website into main thetis repo (from thetisproject.github.io).

Also implements version-numbering that just tracks the git revision using versioneer.


wence-:
To make building seamless, do:

```
cd docs
mkdir build
ln -s THETIS_WEB_DIR html
```

Now when you do `make html`, the output goes directly into the thetisproject.github.io repo where you can commit it.

tkarna:
I guess you meant
```
cd docs
mkdir build
cd build
ln -s THETIS_WEB_DIR html
```

tkarna:
Otherwise looks good to me!

PR #84, opened 2017-01-18
Documentation for website

tkarna:
Add documentation to the website.

Adds
- simple 2D and 3D demos
- short description of  the 2D and 3D model formulations
- description of model outputs, visualization, and restarting a run


tkarna:
Some comments:

- Currently docstrings are not generated for anything under `scripts` subdir. How can it be done?
- Is there an easy way to link to firedrake documentation (for example for `Function`)?
- The demos are not currently compiled into pure python scripts, needs to add pylit

wence-:
Docstrings for `scripts` turns out to be very hard.  Basically, AFAICT, sphinx can *only* document things that it can import (i.e. modules).  Therefore, to document the scripts in the same way as everything else, we need to refactor into something like:

```
thetis_scripts/__init__.py
thetis_scripts/visualize_output.py
scripts/visualize-output
```

Where `visualize-output` is just

```python
#!/usr/bin/env python
if __name__ == "__main__":
    from thetis_scripts.visualize_output import main
    main()
```

and the importable `visualize_output` contains all the meat.  Then `sphinxcontrib.argdoc` can be used.

I can have a go at this after next week's workshop.

tkarna:
OK, good to know. I wonder if it is worth the trouble. My idea was to avoid re-writing the same docstrings in the website docs, but I could just do that. I think we won't be having many scripts, so this should not be an issue.

wence-:
To link to firedrake docs, you need to link to the fully-qualified firedrake thing.  e.g.

```python
:class:`firedrake.function.Function`
```

links to `Function` via intersphinx.  intersphinx does not commute with the "fuzzy" lookup (cf sphinx-doc/sphinx#3351

dham:
I think the answer for scripts might also be to use pylit.

tkarna:
I see. So you'd write the documentation in the script and then generate a rst file from it?

wence-:
Yes (or write the script in the documentation)..

tkarna:
I've included pylit conversion for demos. Sanity check might be in order.

I decided to do the opposite of what done in firedrake: `demos/` directory contains the `*.py` files which are then converted to `*.py.rst` when one makes documentation. That way the user always gets executable python without the need to run `make` in `demos/` dir. Should also be straightforward to add demos in to tests.

tkarna:
About the script documentation issue, AFAIK pylit will convert the whole document, so `myscript.py.rst` would contain all of the python code. This is probably not what we want; I'd just like to be able to link to the docstrings in `visualize-output.py` file.

So Lawrence's suggestion seems better (the dummy wrapper `scripts/visualize-output` files could potentially be generated automatically). But as for the moment there's only one script I think we can deal with this issue later.

tkarna:
AFAIK this is good to go.

tkarna:
Merci bien, closing!

PR #86, opened 2017-02-09
New pressure gradient

tkarna:
Reimplemented internal pressure gradient formulation

- Remove the silly mode-split formulation
- Compute internal pressure gradient as a new diagnostic field
- Add new options to tackle pressure gradient errors:
    - `use_quadratic_pressure`: use P2DGxP2 for r and int_pg
    - `use_quadratic_density`: use P2DGxP2 for density (for nonlinear equation of state)
- Adds DOME example


tkarna:
Travis seems to be broken:
```
Specified virtualenv '/home/travis/install/firedrake' already exists
Can't install into existing virtualenv '/home/travis/install/firedrake'
```

wence-:
That is an odd error on travis: I don't understand what's borked with the caches.  Travis claims there are no caches to delete, and yet it is loading something (but obviously not everything) from cache.

I fixed an actual bug that drone had picked up.

We could drop the travis builder: is it testing anything that drone is not?

tkarna:
Thanks. Travis runs just a subset of the full test suite. So yes, we could drop it. Thus far Travis has been more robust than drone though. E.g. the latest drone build shows some system error that I don't understand. But if we can make drone more reliable we don't really need travis.

tkarna:
Everything seems to work now, I'm going to merge this.

PR #88, opened 2017-02-21
Update to firedrake changes

### LINKED ISSUE #87: Issue with running thetis - bug report
Hi all! I am having a very odd issue with firedrake/thetis. I am trying to run the same code which runs normally in my other machine. The same script used to run in this machine until yesterday, so in order to be sure I reinstalled the firedrake environment and thetis since there was no change in the script I am trying to run. 

Could you please guide me as to what might have gone wrong in case you came across something similar?
I attach the error message produced. :

[bug_report1.txt](https://github.com/thetisproject/thetis/files/789713/bug_report1.txt)

In order to simplify this, I went back to the examples and run channel2d. I tried this both in the master branch and the wetting and drying branch I normally use. I added the output in a text file:

[channel2d_bug_report.txt](https://github.com/thetisproject/thetis/files/789774/channel2d_bug_report.txt)

Also, since this is my first bug report, can you please advise if that was clear enough for future reference?

Kind regards,
Thanasis

wence-:
FIAT elements are now finat elements.  Fixes #87.

### DIFF
--- a/thetis/utility.py
+++ b/thetis/utility.py
@@ -114,32 +114,29 @@ ElementContinuity = namedtuple("ElementContinuity", ["dg", "horizontal_dg", "ver
 """A named tuple describing the continuity of an element."""
 
 
-def element_continuity(fiat_element):
+def element_continuity(finat_element):
     """Return an :class:`ElementContinuity` instance with the
     continuity of a given element.
 
-    :arg fiat_element: The fiat element to determine the continuity
+    :arg finat_element: The finat element to determine the continuity
         of.
     :returns: A new :class:`ElementContinuity` instance.
     """
     import FIAT
-    cell = fiat_element.get_reference_element()
+    cell = finat_element.cell
 
+    dofs = finat_element.entity_dofs()
     if isinstance(cell, FIAT.reference_element.TensorProductCell):
-        # Pull apart
-        horiz = element_continuity(fiat_element.A).dg
-        vert = element_continuity(fiat_element.B).dg
-        return ElementContinuity(dg=horiz and vert,
-                                 horizontal_dg=horiz,
+        A, B = cell.cells
+        dimA = A.get_spatial_dimension()
+        dimB = B.get_spatial_dimension()
+        horiz = all(sum(map(len, dofs[(a, b)].values())) == 0 for a in range(dimA) for b in range(dimB+1))
+        vert = all(sum(map(len, dofs[(a, b)].values())) == 0 for a in range(dimA+1) for b in range(dimB))
+        return ElementContinuity(dg=horiz and vert, horizontal_dg=horiz,
                                  vertical_dg=vert)
     else:
-        edofs = fiat_element.entity_dofs()
         dim = cell.get_spatial_dimension()
-        dg = True
-        for i in range(dim - 1):
-            if any(len(k) for k in edofs[i].values()):
-                dg = False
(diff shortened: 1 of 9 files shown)


thetis/utility.py
```diff
@@ -177,9 +174,10 @@ def extrude_mesh_sigma(mesh2d, n_layers, bathymetry_2d):
     fs_2d = bathymetry_2d.function_space()
     new_coordinates = Function(fs_3d)
 
+    from tsfc.fiatinterface import create_element
+    element = create_element(fs_3d.ufl_element(), vector_is_mixed=False)
```
miklos1: You don't use this?

wence-:
As noted in #89, this patch (which fixes `element_continuity`) reveals a bunch of issues in the weak forms.  We suspect that when integrating by parts, there are some jump terms in the hdiv cases that need to be included.  (Previously, we were treating hdiv spaces as fully discontinuous and therefore including all jump terms).  This was probably hiding issues in the formulation?

tkarna:
Thanks for this update. Currently we use the element continuity check to decide whether we should integrate some terms by parts or not. So I suspect that what happens now is that some terms that should be integrated by parts for hdiv spaces are not treated that way, causing the errors. Thus the quick fix is to just revisit the logic in those terms. I'll run the tests and propose a fix.

Note that currently we are using the same weak forms for the hdiv and dg spaces, which is not ideal. We should add special cases for hdiv spaces as many of the interface terms can be dropped. But that's more involved and I'd do it elsewhere.

tkarna:
OK, so based on what I just wrote, `element_continuity` method should distinguish the three possible cases: `cg`, `hdiv`, and `dg`, because the weak form is going to be different in all cases.

dham:
Rather than the roll-your own element continuity function, would it not be better to extend the UFL element Sobolev space functionality so that tensor product elements know which Sobolev space they are in, and just use that?

tkarna:
I've added a proposed change to `element_continuity`, please amend as needed. Vertical diffusion tests seem to fail still, I'll check those tomorrow.

tkarna:
@dham Err, yes... There should be an easy way to check if a function space is hdiv'd, and also get the element family of the horizontal/vertical sub-spaces.

dham:
Well this works for me:
```
In [6]: e.sub_elements()
Out[6]: (FiniteElement('Q', quadrilateral, 6), FiniteElement('Lagrange', interval, 6))

In [7]: e.sub_elements()[0].sobolev_space()
Out[7]: SobolevSpace('H1', [SobolevSpace('HDiv', [SobolevSpace('L2', [])]), SobolevSpace('HCurl', [SobolevSpace('L2', [])]), SobolevSpace('L2', [])])

In [8]: e.sub_elements()[1].sobolev_space()
Out[8]: SobolevSpace('H1', [SobolevSpace('HDiv', [SobolevSpace('L2', [])]), SobolevSpace('HCurl', [SobolevSpace('L2', [])]), SobolevSpace('L2', [])])
```

tkarna:
Can we merge this as it is, or do you wish to make further changes? Master is currently broken without the API update. All the tests pass.

wence-:
I'm happy to merge this, we will go round and try and do it "right" by exposing appropriate information (probably through UFL).

tkarna:
Sounds good. Merging.

PR #90, opened 2017-02-23
Fix slope limiter at lateral boundary

tkarna:
This (partially) fixes lateral boundary issues discussed in #23  and #29.

To avoid limiting too much at the boundaries, we add the mean value of the field at the lateral boundary facets to the min/max fields. This works in cases where the solution is a linear function that does not attain min/max at a corner. In this case the limiter does not alter the linear function at all. If min/max occurs at the corner, then the corner element will still be limited.

This sits on top of #88.

### DIFF
--- a/thetis/limiter.py
+++ b/thetis/limiter.py
@@ -86,19 +86,44 @@ class VertexBasedP1DGLimiter(VertexBasedLimiter):
         """
         # Call general-purpose bound computation.
         super(VertexBasedP1DGLimiter, self).compute_bounds(field)
-        # NOTE This does not limit solution at lateral boundaries at all
-        # NOTE Omit for now
-        # # Add nodal values from lateral boundaries
-        # par_loop("""
-        #     for (int i=0; i<qmax.dofs; i++) {
-        #         qmax[i][0] = fmax(qmax[i][0], field[i][0]);
-        #         qmin[i][0] = fmin(qmin[i][0], field[i][0]);
-        #     }""",
-        #          ds,
-        #          {'qmax': (self.max_field, RW),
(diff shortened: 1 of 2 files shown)


tkarna:
So here's my first attempt at computing the facet averages with pyop2. Note that I'm just computing the arithmetic mean of the nodal values at the boundary, which is OK for 2D triangles and P1DG space, but it's not exact for the extruded case (quad facets). Using `FacetAvg` might be a better choice.

thetis/limiter.py
```diff
@@ -86,19 +86,44 @@ def compute_bounds(self, field):
...
+        boundary_dofs = entity_support_dofs(self.P1DG.finat_element, entity_dim)
+        local_facet_nodes = np.array([boundary_dofs[e] for e in sorted(boundary_dofs.keys())])
+        n_bnd_nodes = local_facet_nodes.shape[1]
+        local_facet_idx = op2.Global(local_facet_nodes.shape, local_facet_nodes, dtype=np.int32, name='local_facet_idx')
+        code = """
+            void my_kernel(double **qmax, double **qmin, double **field, unsigned int *facet, unsigned int *local_facet_idx)
+            {
+                double face_mean = 0.0;
+                for (int i = 0; i < %(nnodes)d; i++) {
+                    unsigned int idx = local_facet_idx[facet[0]*%(nnodes)d + i];
```
tkarna: The indexing in here is a bit hacky... It wasn't obvious how to use a 2D int array in pyop2.

wence-: No that's right.  Everything is just a 1D array when passed as a global.

tkarna:
All green, merging.

PR #91, opened 2017-03-20
Wetting and drying

swarder:
Wetting-drying scheme as in Karna et al (2010).

The test case I've added uses the Thacker case and compares the elevation at the centre of the domain after one oscillation period, and does this for two (relatively coarse) meshes.

If something more sophisticated than this is required, I can revisit.

### DIFF
--- a/docs/source/model_formulation_2d.rst
+++ b/docs/source/model_formulation_2d.rst
@@ -12,6 +12,11 @@ The prognostic variables are the water elevation :math:`\eta` and depth
 averaged velocity :math:`\bar{\mathbf{u}}`.
 The corresponding fields in Thetis are called ``'elev_2d'`` and  ``'uv_2d'``.
 
+Wetting and drying
+------------------
+
+Wetting and drying is included through the modified bathymetry formulation of Karna et al. (2010). The modified equations are given by :eq:`swe_freesurf_wd`\-:eq:`swe_momentum_wd`.
+
 Spatial discretization
 ----------------------
 
--- a/docs/source/model_formulation_2d.rst
+++ b/docs/source/model_formulation_2d.rst
@@ -15,7 +15,7 @@ The corresponding fields in Thetis are called ``'elev_2d'`` and  ``'uv_2d'``.
 Wetting and drying
 ------------------
 
-Wetting and drying is included through the modified bathymetry formulation of Karna et al. (2010). The modified equations are given by :eq:`swe_freesurf_wd`\-:eq:`swe_momentum_wd`.
+Wetting and drying is included through the modified bathymetry formulation of Karna et al. (2011). The modified equations are given by :eq:`swe_freesurf_wd`\-:eq:`swe_momentum_wd`.
 
 Spatial discretization
 ----------------------
--- a/examples/balzano/balzano.py
+++ b/examples/balzano/balzano.py
@@ -36,7 +36,8 @@ bathymetry.interpolate(x[0] / 2760.0)
 
 # bottom friction suppresses reflection from wet-dry front
 mu_manning = 0.02
-# wetting-drying parameter turns on wetting and drying
+# wetting-drying options
+wetting_and_drying = True
 wd_alpha = 0.4
 
 # --- create solver ---
@@ -49,6 +50,7 @@ options.check_vol_conservation_2d = True
 options.fields_to_export = ['uv_2d', 'elev_2d']
 options.timestepper_type = 'cranknicolson'
 options.shallow_water_theta = 0.5
+options.wetting_and_drying = wetting_and_drying
 options.wd_alpha = wd_alpha
 options.mu_manning = mu_manning
 options.dt = dt
--- a/examples/balzano/balzano.py
+++ b/examples/balzano/balzano.py
@@ -1,13 +1,13 @@
 #  Balzano wetting-drying test case
 # ==============================================
 #
-# Details from e.g. Gourgue et al (2009)
-#
 # Solves shallow water equations with wetting and drying in
 # rectangular domain with sloping bathymetry, with periodic
 # free surface boundary condition applied at deep end.
 #
 # Initial water elevation and velocity are zero everywhere.
+#
+# Further details can be found in, e.g, [1]
 #
 # Demonstrates the use of wetting and drying within Thetis
 #
@@ -16,6 +16,13 @@
 # (h_tilde) and total water depth imposed on original
 # bathymetry (eta_tilde), the latter being useful for
 # comparisons with other WD models.
+#
+# [1] O. Gourgue, R. Comblen, J. Lambrechts, T. Kärnä, V.
+#     Legat, and E. Deleersnijder. A flux-limiting wetting–
+#     drying method for finite-element shallow-water models,
+#     with application to the scheldt estuary. Advances in
+#     Water Resources, 32:1726 – 1739, 2009.
+#     doi: 10.1016/j.advwatres.2009.09.005.
 #
 # Simon Warder 2017-03-21
 
--- a/examples/balzano/balzano.py
+++ b/examples/balzano/balzano.py
@@ -11,6 +11,11 @@
 #
 # Demonstrates the use of wetting and drying within Thetis
 #
+# In addition to bathymetry, elevation and velocity fields,
+# user-specified outputs are used for the moving bathymetry
+# (h_tilde) and total water depth imposed on original
+# bathymetry (eta_tilde), the latter being useful for
+# comparisons with other WD models.
 #
 # Simon Warder 2017-03-21
 
@@ -72,16 +77,21 @@ solverObj.bnd_functions['shallow_water'] = {
     2: ocean_funcs
 }
 
-# User-defined output: moving bathymetry
+# User-defined output: moving bathymetry and eta_tilde
 wd_bathfile = File(os.path.join(outputdir, 'moving_bath.pvd'))
 moving_bath = Function(P1_2d, name="moving_bath")
+eta_tildefile = File(os.path.join(outputdir, 'eta_tilde.pvd'))
+eta_tilde = Function(P1_2d, name="eta_tilde")
 
 
+# user-specified export function
 def export_func():
     wd_bath_displacement = solverObj.eq_sw.bathymetry_displacement_mass_term.wd_bathymetry_displacement
     eta = solverObj.fields.elev_2d
     moving_bath.project(bathymetry + wd_bath_displacement(eta))
     wd_bathfile.write(moving_bath)
+    eta_tilde.project(eta+wd_bath_displacement(eta))
+    eta_tildefile.write(eta_tilde)
 
 
 # callback function to update boundary forcing
--- a/examples/balzano/balzano.py
+++ b/examples/balzano/balzano.py
@@ -77,6 +77,7 @@ def update_forcings(t):
     print_output("Updating boundary condition at t={}".format(t))
     ocean_elev.assign(ocean_elev_func(t))
 
+
 # initial condition: assign non-zero velocity
 solverObj.assign_initial_conditions(uv=Constant((1e-7, 0.)))
 
--- a/examples/balzano/balzano.py
+++ b/examples/balzano/balzano.py
@@ -1,0 +1,82 @@
+#  Balzano wetting-drying test case
+# ==============================================
+#
+# Details from e.g. Gourgue et al (2009)
+#
+# Solves shallow water equations with wetting and drying in
+# rectangular domain with sloping bathymetry, with periodic
+# free surface boundary condition applied at deep end.
+#
+# Initial water elevation and velocity are zero everywhere.
+#
+# Demonstrates the use of wetting and drying within Thetis
+#
+#
+# Simon Warder 2017-03-21
+
+from thetis import *
+
+outputdir = 'outputs'
+mesh2d = RectangleMesh(12, 6, 13800, 7200)
+print_output('Loaded mesh '+mesh2d.name)
+print_output('Exporting to '+outputdir)
+
+# time step in seconds
+dt = 600.
+# total duration in seconds: 4 periods
+t_end = 2*24*3600.
+# export interval in seconds
+t_export = 600.
+
+# bathymetry: uniform slope with gradient 1/2760
+P1_2d = FunctionSpace(mesh2d, 'CG', 1)
+bathymetry = Function(P1_2d, name='Bathymetry')
+x = SpatialCoordinate(mesh2d)
+bathymetry.interpolate(x[0] / 2760.0)
+
+# bottom friction suppresses reflection from wet-dry front
+mu_manning = 0.02
+# wetting-drying parameter turns on wetting and drying
+wd_alpha = 0.4
+
+# --- create solver ---
+solverObj = solver2d.FlowSolver2d(mesh2d, bathymetry)
+options = solverObj.options
+options.t_export = t_export
+options.t_end = t_end
+options.outputdir = outputdir
+options.check_vol_conservation_2d = True
+options.fields_to_export = ['uv_2d', 'elev_2d']
+options.timestepper_type = 'cranknicolson'
+options.shallow_water_theta = 0.5
+options.wd_alpha = wd_alpha
+options.mu_manning = mu_manning
+options.dt = dt
+options.solver_parameters_sw = {
+    'snes_type': 'newtonls',
+    'snes_monitor': True,
+    'ksp_type': 'gmres',
+    'pc_type': 'fieldsplit',
+}
+
+# boundary conditions
+h_amp = -2.0      # ocean boundary forcing amplitude
+h_T = 12*3600.    # ocean boundary forcing period
+ocean_elev_func = lambda t: h_amp * sin(2 * pi * t / h_T)
+solverObj.create_function_spaces()
+H_2d = solverObj.function_spaces.H_2d
+ocean_elev = Function(H_2d, name="ocean boundary elevation").assign(ocean_elev_func(0.))
+ocean_funcs = {'elev': ocean_elev}
+solverObj.bnd_functions['shallow_water'] = {
+    2: ocean_funcs
+}
+
+# callback function to update boundary forcing
+def update_forcings(t):
+    print_output("Updating boundary condition at t={}".format(t))
+    ocean_elev.assign(ocean_elev_func(t))
+
+# initial condition: assign non-zero velocity
+solverObj.assign_initial_conditions(uv=Constant((1e-7,0.)))
+
+solverObj.iterate(update_forcings=update_forcings)
--- a/examples/balzano/balzano.py
+++ b/examples/balzano/balzano.py
@@ -18,7 +18,6 @@ from thetis import *
 
 outputdir = 'outputs'
 mesh2d = RectangleMesh(12, 6, 13800, 7200)
-print_output('Loaded mesh '+mesh2d.name)
 print_output('Exporting to '+outputdir)
 
 # time step in seconds
@@ -35,10 +34,10 @@ x = SpatialCoordinate(mesh2d)
 bathymetry.interpolate(x[0] / 2760.0)
 
 # bottom friction suppresses reflection from wet-dry front
-mu_manning = 0.02
+mu_manning = Constant(0.02)
 # wetting-drying options
 wetting_and_drying = True
-wd_alpha = 0.4
+wd_alpha = Constant(0.4)
 
 # --- create solver ---
 solverObj = solver2d.FlowSolver2d(mesh2d, bathymetry)
@@ -73,6 +72,17 @@ solverObj.bnd_functions['shallow_water'] = {
     2: ocean_funcs
 }
 
+# User-defined output: moving bathymetry
+wd_bathfile = File(os.path.join(outputdir, 'moving_bath.pvd'))
+moving_bath = Function(P1_2d, name="moving_bath")
+
+
+def export_func():
+    wd_bath_displacement = solverObj.eq_sw.bathymetry_displacement_mass_term.wd_bathymetry_displacement
+    eta = solverObj.fields.elev_2d
+    moving_bath.project(bathymetry + wd_bath_displacement(eta))
+    wd_bathfile.write(moving_bath)
+
 
 # callback function to update boundary forcing
 def update_forcings(t):
@@ -83,4 +93,4 @@ def update_forcings(t):
 # initial condition: assign non-zero velocity
 solverObj.assign_initial_conditions(uv=Constant((1e-7, 0.)))
 
-solverObj.iterate(update_forcings=update_forcings)
+solverObj.iterate(update_forcings=update_forcings, export_func=export_func)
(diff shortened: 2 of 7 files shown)


tkarna:
Thanks this looks very good. You should run `make lint` in the repo root dir to make sure your code is pep8 compliant. The test looks OK I think. You might also add some examples that use wetting-drying, for example the Balzano test case no 1 (that could also be added as a test).

test/swe2d/test_thacker.py
```diff
@@ -0,0 +1,60 @@
+# Test for wetting-drying scheme functionality, using Thacker test case
+# Test case details in Gourgue et al (2009)
+
+from thetis import *
+import pytest
+import math
```
tkarna: `math` is not used. In general `numpy` is imported as `np` so you don't need this.

test/swe2d/test_thacker.py
```diff
@@ -0,0 +1,60 @@
...
+# Test case details in Gourgue et al (2009)
+
+from thetis import *
+import pytest
+import math
+
+@pytest.mark.parametrize("n,dt,alpha,max_err", [(25, 300., 4., 0.009), (10, 600., 8., 0.06)])
+def test_thacker(n, dt, alpha, max_err):
+    # Mesh
+    mesh2d = SquareMesh(n, n, 951646.46)
```
tkarna: You could assign the domain size to a variable and then do `X0 = Y0 = l_mesh/2`

test/swe2d/test_thacker.py
```diff
@@ -0,0 +1,60 @@
...
+    L = 430620.
+    eta0 = 2.
+    A = ((D0+eta0)**2-D0**2)/((D0+eta0)**2+D0**2)
+    X0 = 475823.23 # Domain offset
+    Y0 = 475823.23
+
+    # Bathymetry
+    bathymetry = Function(FunctionSpace(mesh2d, "CG", 1), name='bathymetry')
+    x = SpatialCoordinate(mesh2d)
+    bathymetry.interpolate(Expression(D0*(1-((x[0]-X0)*(x[0]-X0)+(x[1]-Y0)*(x[1]-Y0))/(L*L))))
```
tkarna: You don't need to construct an `Expression` object for UFL expressions.

test/swe2d/test_thacker.py
```diff
@@ -0,0 +1,60 @@
+# Test for wetting-drying scheme functionality, using Thacker test case
+# Test case details in Gourgue et al (2009)
+
+from thetis import *
+import pytest
+import math
+
+@pytest.mark.parametrize("n,dt,alpha,max_err", [(25, 300., 4., 0.009), (10, 600., 8., 0.06)])
```
tkarna: You can add `ids`, e.g. `ids=['fine', 'coarse']`, to give human readable names to these configurations.

thetis/options.py
```diff
@@ -245,6 +245,8 @@ def __init__(self):
 
         Bottom stress is :math:`\tau_b/\rho_0 = -g \mu^2 |\mathbf{u}|\mathbf{u}/H^{1/3}`
         """
+        self.wd_alpha = None
+        """Coefficient (units of length) used in bathymetry displacement function that ensures positive water depths."""
```
tkarna: Other options docstrings have the format `'python type: docstring'`, so this should be something like `Coefficient or None: Wetting-drying parameter...` Should also mention that the default is `None` which disables wetting-drying. Unit is meters.

thetis/shallowwater_eq.py
```diff
@@ -217,12 +219,23 @@ def get_bnd_functions(self, eta_in, uv_in, bnd_id, bnd_conditions):
             raise Exception('Unsupported bnd type: {:}'.format(funcs.keys()))
         return eta_ext, uv_ext
 
+    def wd_bathymetry_displacement(self, eta):
+        """
+        Returns wetting and drying bathymetry discplacement as described in:
+        K\"arna et al.,  2011.
```
tkarna: Either two a or two ä ;)

thetis/shallowwater_eq.py
```diff
@@ -317,8 +329,7 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
                     # assume land boundary
                     # impermeability implies external un=0
                     un_jump = inner(uv, self.normal)
-                    h = self.bathymetry
-                    head_rie = head + sqrt(h/g_grav)*un_jump
+                    head_rie = head + sqrt(total_h/g_grav)*un_jump
```
tkarna: This changes behavior. Previously, even with `nonlin=True` we set `h=bathymetry` which implies that the factor in front of the jump is independent of the simulation state (`eta`). Because `total_h` is defined with `eta_old` the proposed modification might be OK (the jump term is still linear), but we need to make sure that everything works.

thetis/shallowwater_eq.py
```diff
@@ -455,7 +467,8 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
                     eta_ext_old, uv_ext_old = self.get_bnd_functions(eta_old, uv_old, bnd_marker, bnd_conditions)
                     # Compute linear riemann solution with eta, eta_ext, uv, uv_ext
                     eta_jump = eta_old - eta_ext_old
-                    un_rie = 0.5*inner(uv_old + uv_ext_old, self.normal) + sqrt(g_grav/self.bathymetry)*eta_jump
+                    total_h = self.get_total_depth(eta_old)
+                    un_rie = 0.5*inner(uv_old + uv_ext_old, self.normal) + sqrt(g_grav/total_h)*eta_jump
```
tkarna: same comment here, `total_h` is time-dependent.

thetis/shallowwater_eq.py
```diff
@@ -708,6 +721,18 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
         return -f
 
 
+class BathymetryDisplacementMassTerm(ShallowWaterContinuityTerm):
+    """
+    Contribution to water level mass term of bathymetry displacement function, used in wetting and drying.
+    Note, that this is a mass term that only takes eta as argument (unlike all other shallow water terms)
```
tkarna: You should probably change the interface to be compliant with the rest of the terms. This function is only used here so it's easy to change.

thetis/shallowwater_eq.py
```diff
@@ -755,30 +782,41 @@ class ShallowWaterEquations(BaseShallowWaterEquation):
...
+        self.add_continuity_terms(eta_test, eta_space, u_space, bathymetry, nonlin, wd_alpha)
+        self.bathymetry_displacement_mass_term = BathymetryDisplacementMassTerm(eta_test, eta_space, u_space, bathymetry, nonlin, wd_alpha)
+
+    def mass_term(self, solution):
+        f = super(ShallowWaterEquations, self).mass_term(solution)
+        if isinstance(solution, list):
+            uv, eta = solution
+        else:
+            uv, eta = split(solution)
+        f += self.bathymetry_displacement_mass_term.residual(eta)
```
tkarna: `self.bathymetry_displacement_mass_term.residual(solution)` would be cleaner.

thetis/shallowwater_eq.py
```diff
@@ -854,9 +893,16 @@ def __init__(self, eta_test, eta_space, u_space,
...
         """
-        super(FreeSurfaceEquation, self).__init__(eta_space, bathymetry, nonlin)
-        self.add_continuity_terms(eta_test, eta_space, u_space, bathymetry, nonlin)
+        super(FreeSurfaceEquation, self).__init__(eta_space, bathymetry, nonlin, wd_alpha)
+        self.add_continuity_terms(eta_test, eta_space, u_space, bathymetry, nonlin, wd_alpha)
+        self.bathymetry_displacement_mass_term = BathymetryDisplacementMassTerm(eta_test, eta_space, u_space, bathymetry, nonlin, wd_alpha)
+
+    def mass_term(self, solution):
+        f = super(ShallowWaterEquations, self).mass_term(solution)
+        f += self.bathymetry_displacement_mass_term.residual(solution)
```
tkarna: A bug here, currently the correct argument is `eta`. Would be cleaner to comply with the standard.

tkarna:
We should also update the documentation. At least the wetting-drying enabling equations should be added to the header of `shallowwater_eq.py`. Would also be nice to have something on the website. Maybe add a wetting-drying section to `docs/source/model_formulation_2d.rst`.

test/swe2d/test_thacker.py
```diff
@@ -0,0 +1,60 @@
...
+        'snes_type': 'newtonls',
+        'snes_monitor': True,
+        'ksp_type': 'gmres',
+        'pc_type': 'fieldsplit',
+    }
+
+    # Initial conditions
+    x = SpatialCoordinate(mesh2d)
+    elev_init = D0*(sqrt(1-A*A)/(1-A) - 1 - ((x[0]-X0)*(x[0]-X0)+(x[1]-Y0)*(x[1]-Y0))*((1+A)/(1-A)-1)/(L*L))
+    solverObj.assign_initial_conditions(uv=Constant((1e-7,0.)), elev=elev_init)
```
tkarna: Could you remind me why the velocity must be non-zero initially? @stephankramer ?

swarder: Ah, this was to avoid issues with the drag term when velocity is zero. Not required here, as this case doesn't use a drag term.

thetis/shallowwater_eq.py
swarder: OK, how do we become sure that everything works?!

I guess adding in another `if self.wd_alpha is None` here could avoid the issue, but that wouldn't be the cleanest solution?

test/swe2d/test_thacker.py
wence-: In fact, I have no idea how this would have worked!

thetis/shallowwater_eq.py
tkarna: Assuming that our test coverage is good enough, passing all the tests should be sufficient. So the first thing to do is to fix the lint issues, so that drone can run the test suite.

swarder: OK, I've now fixed the issues found by `make lint`

swarder: Looks like the tests passed on drone. So this is OK as it is?

swarder:
Regarding the documentation, do you have any recommendations for editors/IDEs for writing rst?

wence-:
I use emacs and `rst-mode`.  Lots of people seem to like atom.io these days

thetis/shallowwater_eq.py
tkarna: Yes, this is probably fine. We can fix issues downstream should anything arise.

tkarna:
AFAIK atom does not give you much more than syntax highlighting. To see the math you'll need to `make html` under `docs/`.

thetis/options.py
```diff
@@ -245,6 +245,12 @@ def __init__(self):
 
         Bottom stress is :math:`\tau_b/\rho_0 = -g \mu^2 |\mathbf{u}|\mathbf{u}/H^{1/3}`
         """
+        self.wd_alpha = None
+        r"""Coefficient or None: Wetting-drying parameter :math:`\alpha`
+
+        Used in bathymetry displacement function that ensures positive water depths. Unit is meters.
+        Default is None, which disables wetting and drying.
```
wence-: FWIW, I'm not a fan of this style of "two different meanings" option.  I would prefer to have:

```python
self.wetting_and_drying = False
r"""Should the shallow water solver handle dry cells?

Uses the wetting and drying scheme from XXX.
If ``True``, one should also set :attr:`wd_alpha` to control the bathymetry displacement."""
self.wd_alpha = 0.1 # some reasonable default value?
r"""Coefficient: Wetting and drying parameter :math:`alpha`.

...
"""

Comment (tkarna) [inline thetis/options.py]: I agree.

Comment (swarder) [inline thetis/options.py]: I'm assuming you're suggesting this as a possibility further down the road? I can change it now if you like, although for now it's consistent with the other options.

Comment (wence-) [inline thetis/options.py]: I'd prefer to do it now.  Don't perpetuate existing bad practice, it just makes it harder to fix later.

Comment (swarder) [inline thetis/options.py]: OK, I'll change it

Comment (swarder) [inline thetis/options.py]: Is setting a default value for `wd_alpha` just the cleanest way to do this, or can we force the user to specify a value? Any default value we set is likely to be inappropriate (and potentially unstable if too small), because the appropriate choice of `wd_alpha` depends on the mesh and bathymetry.

Could we have something like `self.wd_alpha = None` as the default, and then `if wetting_and_drying: assert(wd_alpha is not None)` somewhere?

Or do we just assume that the user has read the docs and knows that they shouldn't leave the default value unchanged?

Comment (swarder) [inline thetis/options.py]: Actually I suppose that's a bad idea, as the default value should be a plausible user specified value, which `None` certainly wouldn't be?

Comment (tkarna) [inline thetis/options.py]: I'd prefer setting a default value. Sanity checking of the options is difficult to maintain and it's not obvious where it should be done. You could mention in the documentation that the parameter is problem specific, and give guidelines on how to choose the value (if there is a rule).

Comment (swarder) [inline thetis/options.py]: OK cool, will do. Thanks

**Comment by swarder** [comment]
OK so something's wrong, but I don't understand the output from the builds.

Firstly, what's the difference between the "push" and "pr" builds? I would guess that the "push" build is building this branch, while "pr" is doing an automatic merge with master first?

The push build got as far as my Thacker test case (which failed, but I can't tell why), but the pr build crashed sooner. I can't make sense of the tracebacks...

**Comment by tkarna** [comment]
Looks like drone system errors in both cases. This happens fairly often. Can you restart these builds? The drone login thingie doesn't work for me (too many redirects error). Lawrence I think has a golden account that magically works. The new build system should fix all these issues.

File: thetis/shallowwater_eq.py [hunk 14]
```diff
@@ -122,6 +122,39 @@
...
+If the option :attr:`.ModelOptions.wetting_and_drying` is ``True``, then wetting and drying is included through the formulation of Karna et al. (2010).
+
+The method introduces a modified bathymetry :math:`\tilde{h} = h + f(H)`, which ensures positive total water depth, with :math:`f(H)` defined by
+
+.. math::
+   f(H) = \frac{1}{2}(\sqrt{H^2 + \alpha^2} - H),
+
+introducing a wetting-drying parameter :math:`\alpha`, with dimensions of length. This results in a modified total water depth :math:`\tilde{H}=H+f(H)`.
+
+The value for :math:`\alpha` is specified by the user through the option :attr:`.ModelOptions.wd_alpha`, in units of metres. The default value for :attr:`.ModelOptions.wd_alpha` is 0.5, but the appropriate value is problem specific and should be set by the user.
```
tkarna: metres -> meters. Options say "Unit is meters". Either spelling is fine as long as we're consistent.

thetis/shallowwater_eq.py
```diff
@@ -122,6 +122,39 @@
...
+
+An approximate method for selecting a suitable value for :math:`\alpha` is suggested by Karna et al. (2010) as :math:`\alpha \approx L_x|\nabla h|`, where :math:`L_x` is the horizontal length scale and :math:`h` the bathymetry depth as defined in the shallow water equations. Smaller :math:`\alpha` leads to more accurate solutions to the shallow water equations in wet regions, but if :math:`\alpha` is too small the simulation will become unstable.
```
tkarna: Wrap long lines.

examples/balzano/balzano.py
```diff
@@ -0,0 +1,86 @@
...
+P1_2d = FunctionSpace(mesh2d, 'CG', 1)
+bathymetry = Function(P1_2d, name='Bathymetry')
+x = SpatialCoordinate(mesh2d)
+bathymetry.interpolate(x[0] / 2760.0)
+
+# bottom friction suppresses reflection from wet-dry front
+mu_manning = 0.02
+# wetting-drying options
+wetting_and_drying = True
+wd_alpha = 0.4
```
tkarna: This should be a Constant? Also `mu_manning`

thetis/options.py
```diff
@@ -245,6 +245,17 @@ def __init__(self):
 
         Bottom stress is :math:`\tau_b/\rho_0 = -g \mu^2 |\mathbf{u}|\mathbf{u}/H^{1/3}`
         """
+        self.wetting_and_drying = False
+        r"""bool: Turn on wetting and drying
+
+        Uses the wetting and drying scheme from Karna et al (2010).
+        If ``True``, one should also set :attr:`wd_alpha` to control the bathymetry displacement.
+        """
+        self.wd_alpha = 0.5
```
tkarna: Also a Constant

thetis/shallowwater_eq.py
```diff
@@ -122,6 +122,39 @@
...
+
+When wetting and drying is turned on, the governing equations are modified by replacing all instances of :math:`H` with :math:`\tilde{H}`, and the addition of a bathymetry displacement term in the free surface equation, resulting in, for the free surface equation and momentum equation respectively,
```
tkarna: Is this grammar correct? Anyway long sentence

examples/balzano/balzano.py
```diff
@@ -0,0 +1,86 @@
...
+# Demonstrates the use of wetting and drying within Thetis
+#
+#
+# Simon Warder 2017-03-21
+
+from thetis import *
+
+outputdir = 'outputs'
+mesh2d = RectangleMesh(12, 6, 13800, 7200)
+print_output('Loaded mesh '+mesh2d.name)
```
tkarna: printing mesh name is not useful for `RectangleMesh`

tkarna:
Looks good, only minor comments. Currently there's no method to export the time-dependent bathymetry. Would that be useful?

swarder:
Thanks. Yeah good idea, I'll export the modified bathymetry. I'll also add eta tilde as an output field, as it can also be helpful for understanding how the WD scheme works.

thetis/shallowwater_eq.py
```diff
@@ -160,11 +207,15 @@ class ShallowWaterTerm(Term):
     """
     def __init__(self, space,
                  bathymetry=None,
-                 nonlin=True):
+                 nonlin=True,
+                 wetting_and_drying=False,
+                 wd_alpha=0.5):
```
wence-: Maybe a general design issue: There are now multiple places (perhaps 10?) where these default values appear.  Would it make sense to always pull them out of the model options, assuming those are available in this constructor?

thetis/shallowwater_eq.py
```diff
@@ -217,12 +268,23 @@ def get_bnd_functions(self, eta_in, uv_in, bnd_id, bnd_conditions):
             raise Exception('Unsupported bnd type: {:}'.format(funcs.keys()))
         return eta_ext, uv_ext
 
+    def wd_bathymetry_displacement(self, eta):
+        """
+        Returns wetting and drying bathymetry displacement as described in:
+        Karna et al.,  2011.
```
swarder: It's 2011, just my mistake

swarder:
I've reproduced the current test failures on my workstation, after updating Firedrake (the tests worked on my machine before the update). And master appears to have similar failures.

wence-:
Failures appear to be due to firedrakeproject/tsfc#103.  Does it work if you turn off coffee?

swarder:
Ah right, thanks. Yeah it works if I turn off coffee.

wence-:
c.f. firedrakeproject/tsfc#104.

wence-:
OK, that fix has merged.

swarder:
Thanks. Any chance you could restart the drone build?

swarder:
Looks like drone system errors again? Could someone restart the build? @wence-?

wence-:
It's all getting monumentally flaky, we should just migrate the build testing to the new jenkins setup and retire drone, @dham is there a recipe I can follow?

examples/balzano/balzano.py
```diff
@@ -0,0 +1,113 @@
...
+eta_tildefile = File(os.path.join(outputdir, 'eta_tilde.pvd'))
+eta_tilde = Function(P1_2d, name="eta_tilde")
+
+
+# user-specified export function
+def export_func():
+    wd_bath_displacement = solverObj.eq_sw.bathymetry_displacement_mass_term.wd_bathymetry_displacement
+    eta = solverObj.fields.elev_2d
+    moving_bath.project(bathymetry + wd_bath_displacement(eta))
+    wd_bathfile.write(moving_bath)
```
tkarna: Hmm, the right way to export these new fields this is to use the exporting functionality: add the field metadata to `field_defs.py` and then users can export the fields just by adding `'moving_bath_2d'` (or such) to `options.fields_to_export` list. It will provide all supported output formats.

In this case, however, we need to call the `project` somewhere. So I think we need to add a mechanism to `Exporter` class to support evaluation callbacks. We'd probably need to rewrite parts of the exporter functionality (I'm not a huge fan of `fields_defs` for example). Thus I suggest that we leave this example as it is and address it in another PR.

swarder:
I'd like to add an atmospheric pressure term for storm surge modelling. Should this be done in a new PR, or can I add it here?

wence-:
It's best to do it separately, otherwise nothing ever gets merged.

thetis/shallowwater_eq.py
```diff
@@ -845,18 +908,24 @@ class FreeSurfaceEquation(BaseShallowWaterEquation):
...
+                 options):
         """
         :arg eta_test: test function of the elevation function space
         :arg eta_space: elevation function space
         :arg u_space: velocity function space
         :arg function_space: Mixed function space where the solution belongs
         :arg bathymetry: bathymetry of the domain
         :type bathymetry: :class:`Function` or :class:`Constant`
-        :kwarg bool nonlin: If False defines the linear shallow water equations
+        :arg options: Object containing all circulation model options
```
tkarna: we should probably state `options` is an `AttrDict` object.

tkarna:
All tests pass on my machine so as far as I'm concerned this is good to go.

docs/source/model_formulation_2d.rst
```diff
@@ -12,6 +12,11 @@ The prognostic variables are the water elevation :math:`\eta` and depth
 averaged velocity :math:`\bar{\mathbf{u}}`.
 The corresponding fields in Thetis are called ``'elev_2d'`` and  ``'uv_2d'``.
 
+Wetting and drying
+------------------
+
+Wetting and drying is included through the modified bathymetry formulation of Karna et al. (2010). The modified equations are given by :eq:`swe_freesurf_wd`\-:eq:`swe_momentum_wd`.
```
stephankramer: This should also be 2011

thetis/shallowwater_eq.py
```diff
@@ -798,32 +869,24 @@ class ModeSplit2DEquations(BaseShallowWaterEquation):
...
         """
         :arg function_space: Mixed function space where the solution belongs
         :arg bathymetry: bathymetry of the domain
         :type bathymetry: :class:`Function` or :class:`Constant`
-        :kwarg bool nonlin: If False defines the linear shallow water equations
-        :kwarg bool include_grad_div_viscosity_term: If True includes grad(nu div(u))
-            viscosity term
-        :kwarg bool include_grad_depth_viscosity_term: If True includes grad(H) term
-            in viscosity operator
+        :arg options: Object containing all circulation model options
```
tkarna: same here: `options` is an `AttrDict` object

swarder:
All tests pass for me too

PR #94, opened 2017-04-04
Add atmospheric pressure forcing term

swarder:
Atmospheric pressure term is useful for storm surge modelling

### DIFF
--- a/test/swe2d/test_atmospheric_pressure.py
+++ b/test/swe2d/test_atmospheric_pressure.py
@@ -93,5 +93,6 @@ def test_pressure_forcing(element_family, timestepper):
     assert(eta_errs[0]/eta_errs[-1] > (2.**expected_order)**(len(eta_errs)-1)*0.75)
     print_output("PASSED")
 
+
 if __name__ == '__main__':
     test_pressure_forcing(el_fam='dg-dg', timestepper='cranknicolson')
--- a/test/swe2d/test_atmospheric_pressure.py
+++ b/test/swe2d/test_atmospheric_pressure.py
@@ -1,72 +1,97 @@
 # Atmospheric pressure test case
-# Constant pressure gradient applied at surface
-# Result compared to analytic solution
-
+#
+# Flow in a rectangular channel is forced by a spatially sinusoidal atmospheric
+# pressure gradient until the solution reaches a steady state, in which the
+# pressure gradient is balanced by the elevation gradient.
+#
+# The elevation is compared to an analytic solution derived from the balance of
+# pressure and elevation gradients.
+#
+# A high bottom friction coefficient is used to suppress oscillations and reach
+# the steady state quickly.
+#
 from thetis import *
 import pytest
 import numpy as np
 
 
-@pytest.mark.parametrize("nx,dt,max_rel_err", [(10, 1200, 2e-6)])
-def test_pressure_forcing(nx, dt, max_rel_err):
-    lx = 21000
-    ly = 5000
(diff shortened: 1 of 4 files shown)


test/swe2d/test_atmospheric_pressure.py
```diff
@@ -0,0 +1,72 @@
+# Atmospheric pressure test case
+# Constant pressure gradient applied at surface
+# Result compared to analytic solution
```
tkarna: The description could be better: you force the flow in a rectangular channel with a constant atm pressure gradient until the solution reaches a steady state? In the steady state the elevation gradient balances out the pressure gradient, and you compare the elevation to the analytical solution?

thetis/options.py
```diff
@@ -257,6 +257,8 @@ def __init__(self):
         """2D Coefficient or None: Coriolis parameter"""
         self.wind_stress = None
         """Coefficient or None: Stress at free surface (2D vector function)"""
+        self.atmospheric_pressure = None
+        """Coefficient or None: Atmospheric pressure at free surface"""
```
tkarna: Should probably mention the units of the pressure

test/swe2d/test_atmospheric_pressure.py
```diff
@@ -0,0 +1,72 @@
...
+    # atmospheric pressure
+    atmospheric_pressure = Function(P1, name='atmospheric_pressure')
+    atmospheric_pressure.interpolate(x[0])
+
+    # --- create solver ---
+    solverObj = solver2d.FlowSolver2d(mesh2d, bathymetry)
+    options = solverObj.options
+    options.dt = dt
+    options.t_export = options.dt
+    options.t_end = t_end - 0.1*options.dt
```
tkarna: why is `- 0.1*options.dt` needed?

test/swe2d/test_atmospheric_pressure.py
```diff
@@ -0,0 +1,72 @@
...
+    bathymetry = Function(P1, name='bathymetry')
+    x = SpatialCoordinate(mesh2d)
+    bathymetry.interpolate(Constant(5.0))
+
+    # bottom drag
+    mu_manning = Constant(2.0)
+
+    # atmospheric pressure
+    atmospheric_pressure = Function(P1, name='atmospheric_pressure')
+    atmospheric_pressure.interpolate(x[0])
```
tkarna: For better test coverage the pressure gradient should not only be in the x direction.

tkarna:
Added minor comments. As you've added a new term in the equations it would be useful to do a convergence test to see that you get the correct convergence rate. Should be easy enough to do as you have an analytical solution. You can take a look at the 2D MMS tests (or 3D tracer MES tests) for examples. You should then run the test with all the supported element types and a bunch of time integrators to make sure that we catch errors should something break.

tkarna:
I'd also like to see a 3D test case. Could derive from `barotropicChannel/test_closed_channel.py` for example. If you do not want to learn how to run the 3D model at this time, I can add the test, just let me know. I'll need atm forcing in the 3D model soon, so this is very timely.

test/swe2d/test_atmospheric_pressure.py
swarder: This was to avoid overshooting (as the iterator keeps stepping `while self.simulation_time <= self.options.t_end + t_epsilon:`), but you're right, it's not necessary for this test.

tkarna:
Looks good to me, apart from the lint fail.

PR #96, opened 2017-05-16
Tidalfarm optimisation

stephankramer:
When using thetis_adjoint adds annotation to the 2d solver to annotate the start time (adj_start_timestep) and endings of each timestep (adj_inc_timestep).

Expands the tidalfarm optimisation example to do a fully time-dependent farm optimisation in the "continuous approach" (optimising for a turbine density). This example is based on an OpenTidalFarm example (headland-optimisation) and uses a time-integrated functional including power output and costs. All the turbine specific stuff is wrapped in the example. For the continuous approach this is not all that much; The future plan is to wrap this and the stuff needed for the bump turbines in the discrete approach in a separate module ("OpenTidalFarm 2.0").

For optimisation with thetis in general we could consider adding a ThetisReducedFunctional, which like the MyReducedFunctional in the example, overloads the standard ReducedFunctional to allow for extra output. In the example it is shown how to do this for outputting the control and gradient every optimisation iteration - I would also like to add the ability to output the normal per timestep (or whatever t_export is set to) output of the forward model every time it is run. This would require overloading the evaluate() method, where instead of replaying the forward model from the tape, we rerun the model by hand (using the solver object) and let it do its normal exporting (with a different output_dir for every optimisation iteration).


examples/tidalfarm/tidalfarm.py
```diff
@@ -1,119 +1,189 @@
...
-out = File('gradient_J.pvd')
-out.write(dJdc)
-J0 = assemble(integral)
-print_output("Functional evaluated by hand: ", J0)
+class TurbineDragTerm(shallowwater_eq.ShallowWaterMomentumTerm):
+    def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=None):
+        total_h = self.get_total_depth(eta_old)
+        C_D = turbine_friction
+        f = C_D * sqrt(dot(uv_old, uv_old)) * inner(self.u_test, uv) / total_h * self.dx(2)
+        return -f
```
tkarna: You are defining this as a custom term because the integral is only over a custom subdomain?
I wonder if we could incorporate this in the SW equations. Would be cleaner.

examples/tidalfarm/tidalfarm.py
```diff
@@ -1,119 +1,189 @@
...
+    print "Updating tidal elevation at t = ", t
+    P = assemble(power_integral)
+    N = assemble(cost_integral)
+    print "Power, N turbines, profit =", P, N, P-break_even_wattage*N
+    # NOTE: we need to explicitly tell dolfin-adjoint to annotate this as by default it seems to
+    # only annotate if the interpoland is a Function (is this reasonable?)
+    tidal_elev.interpolate(tidal_amplitude*sin(omega*t + omega/pow(g*H, 0.5)*x[0]), annotate=annotate)
+    # NOTE: it seems firedrake-adjoint (dolfin-adjoint in general?) cannot handle functions that are part of a
+    # time-integrated Functional which do not change over time and are therefore not annotated
+    turbine_friction.project(turbine_friction, annotate=annotate)
```
tkarna: You'd probably want to use `Interpolator` and `Projector` objects here.

examples/tidalfarm/tidalfarm.py
stephankramer: Yes, it would be good to discuss this. There are various bits in this example that I would like to move out (eventually) and offer as generic functionality. For the tidal drag here it's indeed just a standard quadratic drag term but applied over a subdomain (the farm areas). There will be various extensions of this however:
* we often do this for multiple farm areas, so you'd have a multiple of these terms
* in the discrete approach, where we optimise for the position of individual turbines, the turbine friction is not a function but an expression that basically creates individual localised bump functions at the position of each turbine
* for this discrete approach, there is a correction term that approximately translates between the depth-averaged model velocity and either the actual velocity through the turbine or the theoretical free stream velocity. This correction depends on the depth
* in the more realistic case where the thrust coefficient is not constant (thrust curves), the drag is not really quadratic any more as the friction coefficient itself also depends on the local velocity

So in general this could all get a little more messy, so I was thinking of having this sort of stuff in a separate module - but we could also have a generic turbine drag term in the SWE module where we allow the coefficient to be any sort of expression possibly dependent on uv_old and depth, and allow it to either be integrated over the entire domain (for the discrete approach) or over a sum of subdomains. I'm not sure that would be cleaner though.

examples/tidalfarm/tidalfarm.py
stephankramer: I'll give that a try. I'm not entirely sure it's going to work though: the project call here is just a work-around for a bug in dolfin-adjoint: if turbine_friction is not touched during a timestep, dolfin-adjoint fails to do the time-integral as it gets confused about what the value of turbine_friction during each timestep should be. I previously worked around this with a similar interpolate call which fixed the error message, but gave the wrong answer! I will submit a bug-report to dolfin-adjoint as well.

stephankramer:
@tkarna: would you mind having a look at this again? It's not all based on pyadjoint. Also all new functionality has now moved into thetis modules rather than being in the example itself.

thetis/shallowwater_eq.py
```diff
@@ -729,6 +729,32 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
...
+    r"""
+    Turbine drag parameterisation implemented through quadratic drag term
+    :math:`c_t \| \bar{\textbf{u}} \| \bar{\textbf{u}}`
+
+    where the turbine drag :math:`c_t` is related to the turbine thrust coefficient
+    :math:`C_T`, the turbine diameter :math:`A_T`, and the turbine density :math:`d`
+    (n/o turbines per unit area), by:
+
+    .. math::
+        C_D = (C_T A_T d)/2
```
tkarna: Should be `c_t` instead of `C_D`?

thetis/optimisation.py
```diff
@@ -0,0 +1,244 @@
...
+    not in the __call__ method (as this one is directly called from the :ReducedFunctional:). In addition,
+    like any DiagnosticCallback, the name and variable_names properties and a message_str method need to be defined.
+    """
+
+    def __init__(self, solver_obj, **kwargs):
+        """
+        :arg solver_obj: Thetis solver object
+        :arg kwargs: keyword arguments passed to :DiagnosticCallback:
+        """
+        kwargs.setdefault('include_time', False)
```
tkarna: Why are you omitting the time in the DiagnosticOptimisationCallback?

thetis/callback.py
```diff
@@ -95,12 +98,19 @@ def __init__(self, filename, varnames, array_dim=1, attrs=None,
...
         # TODO is there an easier way for doing this?
-        for var in self.varnames + ['time']:
+        for var in self.varnames:
             arr = hdf5file[var]
             shape = arr.shape
             arr.resize((shape[0] + 1, shape[1]))
+        if self.include_time:
+            arr = hdf5file['time']
+            shape = arr.shape
+            arr.resize((shape[0] + 1, shape[1]))
```
tkarna: You are duplicating the code here in lines 106-108. Could plausibly refactor differently to avoid duplication

stephankramer:
@tkarna I think I've addressed your comments. Also, rebased on master.

thetis/optimisation.py
```diff
@@ -20,20 +23,25 @@ class UserExportManager(ExportManager):
...
-    If function.name() is in field_defs.field_metadata, the filename and shortname are
-    obtained there. If the function.name() is unknown, both are based on function.name()
-    directly (with an optional additional filename_prefix). Filenames and shortnames
-    can be overruled by the shortnames and filenames arguments."""
+    In the standard :class:`.ExportManager` all provided functions need to have standard names
+    present in :py:data:`.field_metadata`. Here, any functions can be provided. If function.name() is in
+    :py:data:`.field_metadata`, the standard filename and shortname  are used.
+    If the function.name() is unknown, both are based on function.name()
+    directly (with an optional additional filename_prefix). Filenames and
+    shortnames can be overruled by the shortnames and filenames arguments."""
```
tkarna: We need to generalize the `field_metadata` mechanism, so that users can easily add new fields with all the metadata.

stephankramer: Agreed. This was a a bit of a workaround. Related, it would also be good to have a discussion about how to do advection of an arbitrary number of scalars, not just salinity and temperature.

tkarna: Yes, we'll definitely need that soon.

PR #97, opened 2017-05-16
Assembledschur

stephankramer:
Add a AssembledSchurPC preconditioner that can be used to solve the schur system with the 'pressureprojectionpicard' timestepper. This gives very good performance with the dg-cg element pair for the depth-averaged equations.

The AssembledSchurPC provides a preconditioner for the schur system of the velocity-pressure system [[K,G],[C, P]] by explicitly assembling (an approximation to) it, assuming the velocity is DG and the velocity block K is just a mass matrix M. This is true for the pressureprojectionpicard time integration with dg-cg as the other terms in the momentum equation are solved in a separate step. dg-cg has the particular advantage that the stencil of C M^-1 G is compact (it's the same as that of the standard Laplacian in pressure space). For dg-dg it becomes a lot bigger. Also currently, even with pressureprojectionpicard the velocity block K of the dg-dg discretisation contains an additional term which stems from the stabilisation of the pressure gradient. In cases where M != K the AssembledSchurPC could still be used as an approximate preconditioner.

This is perhaps a little specific, and maybe it should live somewhere else in the long run - but it is something that we use in all large-scale tidal runs at AMCG so it would be very useful to have it available out of the box.

### DIFF
--- a/thetis/assembledschur.py
+++ b/thetis/assembledschur.py
@@ -12,7 +12,6 @@ class AssembledSchurPC(PCBase):
         v and u are the test and trial of the 00 block. This gives the exact
         inverse of the mass matrix for a DG discretisation."""
     def initialize(self, pc):
-        print "Initializing AssembledSchurPC"
         _, P = pc.getOperators()
         ctx = P.getPythonContext()
         a = ctx.appctx['a']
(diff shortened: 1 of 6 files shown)


thetis/assembledschur.py
```diff
@@ -0,0 +1,78 @@
...
+        if r < 0:
+            raise RuntimeError("LinearSolver failed to converge after %d iterations with reason: %s", self.ksp.getIterationNumber(), solving_utils.KSPReasons[r])
+
+    def view(self, pc, viewer=None):
+        super(AssembledSchurPC, self).view(pc, viewer)
+        if viewer is None:
+            return
+        if viewer.getType() != PETSc.Viewer.Type.ASCII:
+            return
+        viewer.pushASCIITab()
```
wence-: Could print something here like:

```
viewer.printfASCII("Solves assembled Schur system D - C M.inv C^T\n")
```

thetis/assembledschur.py
```diff
@@ -0,0 +1,78 @@
...
+
+        self.A10_A00_inv = A10.matMult(self.A00_inv, self.A10_A00_inv, 2.0)
+        self.schur = self.A10_A00_inv.matMult(A01, self.schur, 2.0)
+        if self.schur_plus is None:
+          self.schur_plus = self.schur.duplicate(copy=True)
+          self.schur_plus.aypx(-1.0, A11, PETSc.Mat.Structure.DIFFERENT_NONZERO_PATTERN)
+        else:
+          self.schur_plus = self.schur.copy(self.schur_plus, PETSc.Mat.Structure.DIFFERENT_NONZERO_PATTERN)
+          self.schur_plus.aypx(-1.0, A11, PETSc.Mat.Structure.SAME_NONZERO_PATTERN)
+        self.ksp.setOperators(self.schur_plus)
```
stephankramer: Neat! Will give it a go...

tkarna:
Looks good to me. I presume that you'll need to set `options.solver_parameters_sw` in your problem file every time. At some point we should implement options hierarchies so that we could define meaningful default parameters for each time stepper.

stephankramer:
Yes indeed the solver options for this get very unwieldy - they tend to cover about half of my setup files! - so it would be good to have something more user friendly for that.

stephankramer:
I tried out @wence's suggestion of using SLATE for this, but it appears SLATE does not yet support subdomains (in this case for the exterior facet integrals). The gradient and divergence terms do have boundary terms which require those.

tkarna:
I guess this needs to be rebased now, but other than that this looks OK.

PR #98, opened 2017-05-23
Dome test

tkarna:
3D model updates for the DOME test case.

Changes in solver:
- 3D solver has `solver.create_fields` methods
- Added a method for computing hydrostatic consistency metric on 3D meshes
- `DiagnosticCallback` objects can store vector valued data and hdf5 attributes
- Added simple parametric Pacanowski-Philander vertical turbulence model


tkarna:
Thanks, Lawrence. Fixed tests and rebased on master.

thetis/turbulence.py
```diff
@@ -1077,3 +1077,146 @@ def __init__(self, function_space, gls_model,
...
+        """float: Exponent of viscosity numerator :math:`n`"""
+
+    def print_summary(self):
+        """Prints all defined parameters and their values."""
+        print_output('Pacanowski-Philander model parameters')
+        for k in sorted(self.keys()):
+            print_output('  {:16s} : {:}'.format(k, self[k]))
+
+
+class PacanowskiPhilanderModel(object):
```
stephankramer: Should we maybe have a base TurbulenceModel class? So we can document what the expected API is?

tkarna:
Now updated and rebased on master. AFAIC this is ready to go.

thetis/options.py
```diff
@@ -124,8 +124,34 @@ class SemiImplicitTimestepperOptions3d(ExplicitTimestepperOptions3d):
         help='implicitness parameter theta for 2D solver. Value 1.0 implies fully implicit formulation.').tag(config=True)
 
 
-class GLSModelOptions(FrozenHasTraits):
-    """Options for generic length scale turbulence model"""
+class TurbulenceModelOptions(FrozenHasTraits):
```
stephankramer: Could we perhaps define this in a more general Options class from which all other Options classes are derived (instead of FrozenHasTraits), ModelOptions2D, TimestepperOptions, etc. - there is nothing turbulence-model-specific about that update() function and could be useful elsewhere.

thetis/options.py
```diff
@@ -124,8 +124,34 @@ class SemiImplicitTimestepperOptions3d(ExplicitTimestepperOptions3d):
...
+            self.__setattr__(key, params_dict[key])
+
+
+class PacanowskiPhilanderModelOptions(TurbulenceModelOptions):
+    """Options for Pacanowski-Philander turbulence model"""
+    max_viscosity = PositiveFloat(5e-2, help=r"float: Constant maximum viscosity :math:`\nu_{max}`").tag(config=True)
+    alpha = PositiveFloat(10.0, help="float: Richardson number multiplier").tag(config=True)
+    exponent = PositiveFloat(2.0, help=r"float: Exponent of viscosity numerator :math:`n`").tag(config=True)
+
+    def print_summary(self):
```
stephankramer: Same here: we should have a method to print the contents of any of the Options classes. It would be nice to have this as a  `__str__()` method, so you can simply do:

    print(PacanowskiPhilanderModelOptions(...))

thetis/callback.py
```diff
@@ -44,6 +44,7 @@ def add(self, callback, mode):
         :arg mode: register callback under this mode
         :type mode: string
         """
+        assert isinstance(mode, (str, unicode))
```
stephankramer: In python 2 AKA: `isinstance(mode, basestring)`. I suppose we need to revisit this when we're switching to python 3.

thetis/options.py
tkarna: Good idea, I've now added `OptionsBase` class for this purpose.

thetis/options.py
tkarna: Yes, also added `__str__`.

thetis/callback.py
tkarna: fix-ed.

thetis/configuration.py
```diff
@@ -195,7 +195,38 @@ def info(self):
         return result + super(PairedEnum, self).info()
 
 
-class FrozenHasTraits(HasTraits):
+class OptionsBase(object):
+    """Abstract base class for all options classes"""
+
+    def name(self):
```
tkarna: I'd like to make this an `@abstractproperty` but if I use `_metaclass__ = ABCMeta` I get metaclass conflict in `FrozenHasTraits` class:
```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```
Some magic to work around it? @wence- ?

wence-: Yeah, since `HasTraits` is a metaclass, you need to interpose something else (ugh).  Like this perhaps?

```python

class OptionsBase(object):
    __metaclass__ = ABCMeta

class Interposer(OptionsBase, HasTraits):
    pass

class FrozenHasTraits(object):
    __metaclass__ = Interposer
```

I think

stephankramer: I don't think that's quite right: HasTraits is not a metaclass - but it does have a custom metaclass (MetaHasTraits), so subclasses of both HasTraits and OptionsBase need a metaclass that this a subclass of both ABCMeta and MetaHasTraits. I've pushed a fix in that direction.

wence-:
Fixed master for changes in Firedrake and rebased.

wence-:
Or rather, #102 is required to be merged first.

tkarna:
Added a fix for python3. Alles klar now.

PR #99, opened 2017-06-16
Interpolation routines for external forcings

tkarna:
Added generic routines for interpolating external data on Firedrake fields.

In general, we want to read time series from a file tree, where it may be necessary to interpolate the data in space onto the local Firedrake grid. Then we need to be able to locate previous/next defined time steps in the file tree, and then do linear interpolation in between. To this end the problem is divided into multiple parts:

- Generic `GridInterpolator` that can interpolate data between grids using Delaunay triangulation
- `SpatialInterpolator` that builds a `GridInterpolator` for the local mesh
- `FileTreeReader` that returns data for a certain time step in a file (applying necessary spatial interp.)
- `TimeParser` that parses all the time steps that are stored in a file
- `TimeSearch` object that scans a file tree and allows searching for next/previous time steps
- `LinearTimeInterpolator` that uses `TimeSearch` and `FileTreeReader` to do the actual time interpolation

To facilitate the above, also added coordinate system definition (using `pyproj` which is now a dependency) and support for time zones.

An example of a 2D atmospheric pressure forcing object:

```python
def to_latlon(x, y, positive_lon=False):
    # Converts mesh (x,y) points to coordinates used in the atm data
    lon, lat = coordsys_spcs.spcs2lonlat(x, y)
    if positive_lon and lon < 0.0:
        lon += 360.
    return lat, lon


class WRFInterpolator(object):
    # Interpolates WRF atmospheric model data on 2D fields
    def __init__(self, function_space, atm_pressure_field, ncfile_pattern, init_date):
        self.function_space = function_space
        self.atm_pressure_field = atm_pressure_field

        # object that interpolates forcing data from structured grid on the local mesh
        self.grid_interpolator = NetCDFLatLonInterpolator2d(self.function_space, to_latlon)
        # reader object that can read fields from netCDF files, applies spatial interpolation
        self.reader = NetCDFSpatialInterpolator(self.grid_interpolator, ['prmsl'])
        # object that can find previous/next time stamps in a collection of netCDF files
        self.timesearch_obj = NetCDFTimeSearch(ncfile_pattern, init_date, NetCDFTimeParser)
        # finally a linear intepolator class that performs linar interpolation in time
        self.interpolator = LinearTimeInterpolator(self.timesearch_obj, self.reader)

    def set_fields(self, time):
        # Evaluates forcing fields at the given time
        pressure = self.interpolator(time)
        self.atm_pressure_field.dat.data_with_halos[:] = pressure
```
Usage in input file:
```python
sim_tz = timezone.FixedTimeZone(-8, 'PST')
init_date = datetime.datetime(2015, 5, 16, tzinfo=sim_tz)
atm_pressure_2d = Function(solver_obj.function_spaces.P1_2d, name='atm pressure')
wrf_pattern = 'forcings/atm/wrf/wrf_air.2016_*_*.nc'
wrf_atm = WRFInterpolator(
    solver_obj.function_spaces.P1_2d, atm_pressure_2d, wrf_pattern, init_date)
simulation_time = 3600.
wrf_atm.set_fields(simulation_time)
```

### DIFF
--- a/thetis/coordsys.py
+++ b/thetis/coordsys.py
@@ -62,11 +62,11 @@ def get_vector_rotation_matrix(source_sys, target_sys, x, y, delta=None):
     """
     if delta is None:
         delta = 1e-6  # ~1 m in LL_WO
-    x, y = pyproj.transform(source_sys, target_sys, x, y)
+    x1, y1 = pyproj.transform(source_sys, target_sys, x, y)
 
     x2, y2 = pyproj.transform(source_sys, target_sys, x, y + delta)
-    dxdl = (x2 - x) / delta
-    dydl = (y2 - y) / delta
+    dxdl = (x2 - x1) / delta
+    dydl = (y2 - y1) / delta
     theta = np.arctan2(-dxdl, dydl)
 
     c = np.cos(theta)
--- a/thetis/coordsys.py
+++ b/thetis/coordsys.py
@@ -6,9 +6,6 @@ Uses pyproj library.
 import pyproj
 import numpy as np
 
-# NOTE this adds pyproj as dependencency
-# TODO generalize; define unique names for commonly-used systems
-
 
 UTM_ZONE10 = pyproj.Proj(
     proj='utm',
@@ -18,47 +15,56 @@ UTM_ZONE10 = pyproj.Proj(
     errcheck=True)
 SPCS_N_OR = pyproj.Proj(init='nad27:3601', errcheck=True)
 LL_WGS84 = pyproj.Proj(proj='latlong', datum='WGS84', errcheck=True)
-LL_WO = pyproj.Proj(proj='latlong', nadgrids='WO', errcheck=True)  # HARN
 
 
-def convertCoords(x, y, fromSys, toSys):
-    """Converts x,y from fromSys to toSys."""
+def convert_coords(source_sys, target_sys, x, y):
+    """
+    Converts coordinates from source_sys to target_sys
+
+    This function extends pyproj.transform method by handling NaNs correctly.
+
+    :arg source_sys: pyproj coordinate system where (x, y) are defined in
+    :arg target_sys: target pyproj coordinate system
+    :arg x: x coordinate
+    :arg y: y coordinate
+    :type x: float or numpy.array_like
+    :type y: float or numpy.array_like
+    """
     if isinstance(x, np.ndarray):
         # proj may give wrong results if nans in the arrays
-        lon = np.ones_like(x) * np.nan
-        lat = np.ones_like(y) * np.nan
+        lon = np.full_like(x, np.nan)
+        lat = np.full_like(y, np.nan)
         goodIx = np.logical_and(np.isfinite(x), np.isfinite(y))
         lon[goodIx], lat[goodIx] = pyproj.transform(
-            fromSys, toSys, x[goodIx], y[goodIx])
+            source_sys, target_sys, x[goodIx], y[goodIx])
     else:
-        lon, lat = pyproj.transform(fromSys, toSys, x, y)
+        lon, lat = pyproj.transform(source_sys, target_sys, x, y)
     return lon, lat
 
 
-def getVectorRotationMatrix(lon, lat, target_csys, source_csys=None,
-                            delta=None):
+def get_vector_rotation_matrix(source_sys, target_sys, x, y, delta=None):
     """
-    Estimate rotation matrix that converts vectors from source_csys to
-    target_csys at (lon, lat) location.
+    Estimate rotation matrix that converts vectors defined in source_sys to
+    target_sys.
 
-
-    A vector can defined at (lon, lat) can then be transformed as
+    Assume that we have a vector field defined in source_sys: vectors located at
+    (x, y) define the x and y components. We can then rotate the vectors to
+    represent x2 and y2 components of the target_sys by applying a local
+    rotation:
 
     .. code-block:: python
 
-        R, theta = getVectorRotationMatrix(lon, lat, target_csys)
-        v_ll = numpy.array([[v_lon], [v_lat]])
-        v_csys = numpy.matmul(R, v_ll)
-        v_x, v_y = v_csys
+        R, theta = get_vector_rotation_matrix(source_sys, target_sys, x, lat)
+        v_xy = numpy.array([[v_x], [v_y]])
+        v_new = numpy.matmul(R, v_xy)
+        v_x2, v_y2 = v_new
 
     """
-    if source_csys is None:
-        source_csys = LL_WO
     if delta is None:
         delta = 1e-6  # ~1 m in LL_WO
-    x, y = pyproj.transform(source_csys, target_csys, lon, lat)
+    x, y = pyproj.transform(source_sys, target_sys, x, y)
 
-    x2, y2 = pyproj.transform(source_csys, target_csys, lon, lat + delta)
+    x2, y2 = pyproj.transform(source_sys, target_sys, x, y + delta)
     dxdl = (x2 - x) / delta
     dydl = (y2 - y) / delta
     theta = np.arctan2(-dxdl, dydl)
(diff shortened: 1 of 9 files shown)


tkarna:
Please review. I'm sure some of the classes could be refactored and made more generic. Currently I'm using these methods to do time series interpolation (river flux/temperature), 2D atmospheric forcings (as shown) and also reading ROMS data to 3D scalar fields. But I'm sure there are many other use cases that aren't covered.

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+
+UTM_ZONE10 = pyproj.Proj(
+    proj='utm',
+    zone=10,
+    datum='NAD83',
+    units='m',
+    errcheck=True)
+SPCS_N_OR = pyproj.Proj(init='nad27:3601', errcheck=True)
+LL_WGS84 = pyproj.Proj(proj='latlong', datum='WGS84', errcheck=True)
+LL_WO = pyproj.Proj(proj='latlong', nadgrids='WO', errcheck=True)  # HARN
```
stephankramer: What is nadgrids='WO' ? Maybe this is all a little specific?

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def convertCoords(x, y, fromSys, toSys):
```
stephankramer: I prefer lower case for function names - same for arguments.

More generally: we expect users to use pyproj (to define the fromSys and toSys arguments for starters), so they'll be used to pyproj.transform. If we think this function is useful for general use, we should probably keep it similar to pyproj.transform and have the same argument order - and in the __doc__ say that this is just a "thin wrapper" around pyproj.transform that deals correctly with nans.

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+        lat = np.ones_like(y) * np.nan
+        goodIx = np.logical_and(np.isfinite(x), np.isfinite(y))
+        lon[goodIx], lat[goodIx] = pyproj.transform(
+            fromSys, toSys, x[goodIx], y[goodIx])
+    else:
+        lon, lat = pyproj.transform(fromSys, toSys, x, y)
+    return lon, lat
+
+
+def getVectorRotationMatrix(lon, lat, target_csys, source_csys=None,
```
stephankramer: lower_case ?

Does this work for all coordinate transformations - is there some assumption about orthogonality here? I guess lon, lat velocities are kind of special (in that they're not actually vectors in the lon, lat coordinate space), but I _think_ this still does the right thing?

lon, lat should probably be more generic x,y? Also from the __doc__ it's not quite clear whether the point itself is specified in target_csys or source_csys. Finally we have fromSys and toSys in the previous routine, and target_csys and source_csys here.

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+    .. code-block:: python
+
+        R, theta = getVectorRotationMatrix(lon, lat, target_csys)
+        v_ll = numpy.array([[v_lon], [v_lat]])
+        v_csys = numpy.matmul(R, v_ll)
+        v_x, v_y = v_csys
+
+    """
+    if source_csys is None:
+        source_csys = LL_WO
```
stephankramer: Is this a sensible default? Also this default could be set in the argument list? (so it's more easily discoverable)

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+    theta = np.arctan2(-dxdl, dydl)
+
+    c = np.cos(theta)
+    s = np.sin(theta)
+    R = np.array([[c, -s], [s, c]])
+
+    return R, theta
+
+
+class VectorCoordSysRotation(object):
```
stephankramer: Ah ok, so I guess this is the actual normal use case: converting *from* velocities that are aligned with the lat, lon directions to proper velocity vectors in a flat coordinate system? I presume LL_WO vs. LL_WGS84 doesn't matter here? Or does it?

thetis/timezone.py
```diff
@@ -0,0 +1,43 @@
+"""
+Timezone definitions and conversion methods
+"""
+import datetime
+import pytz
+
+utc_tz = pytz.timezone('UTC')
```
stephankramer: There's also pytz.utc - do we really need our own?

thetis/coordsys.py
wence-: I agree with this.

thetis/interpolation.py
```diff
@@ -0,0 +1,634 @@
...
+        if self.fill_nearest:
+            # find nearest neighbor in the data set
+            from scipy.spatial import cKDTree
+            dist, ix = cKDTree(ngrid_xyz).query(ntarget_xyz[self.outside])
+            self.outside_to_nearest = ix
+
+    def _normalize_coords(self, xyz):
+        if self.normalize:
+            return self.norm_a*xyz + self.norm_b
+        return xyz
```
wence-: I would just push this into `__init__`.

thetis/interpolation.py
```diff
@@ -0,0 +1,634 @@
...
+
+    def _get_slice(self, time_index):
+        """
+        Returns a slice object that extracts a single time index
+        """
+        if self.ndims == 1:
+            return time_index
+        slice_list = [slice(None, None, None)]*self.ndims
+        slice_list[self.time_dim] = slice(time_index, time_index+1, None)
+        return slice
```
wence-: This looks wrong.  Did you mean to `return slice_list`?

thetis/interpolation.py
```diff
@@ -0,0 +1,634 @@
...
+    """
+    Abstract spatial interpolator class that can interpolate onto a Function
+    """
+    def __init__(self, function_space, to_latlon):
+        """
+        :arg function_space: target Firedrake FunctionSpace
+        :arg to_latlon: Python function that converts local mesh coordinates to
+            latitude and longitude: 'lat, lon = to_latlon(x, y)'
+        """
+        self.function_space = function_space
```
wence-: No need to remember function space, it looks like.

thetis/interpolation.py
```diff
@@ -0,0 +1,634 @@
...
+
+        # construct local coordinates
+        xy = SpatialCoordinate(self.function_space.mesh())
+        fsx = Function(self.function_space).interpolate(xy[0]).dat.data_with_halos
+        fsy = Function(self.function_space).interpolate(xy[1]).dat.data_with_halos
```
wence-: You can destructure `x, y = SpatialCoordinate(...)`.  Should probably `assert function_space.ufl_element().value_shape() == ()`

thetis/interpolation.py
```diff
@@ -0,0 +1,634 @@
...
+
+    .. code-block:: python
+        fs = FunctionSpace(...)
+        myfunc = Function(fs, ...)
+        ncinterp2d = NetCDFLatLonInterpolator2d(fs, to_latlon)
+        val1, val2 = ncinterp2d.interpolate(nc_filename, ['var1', 'var2'], 10)
+        myfunc.dat.data_with_halos[:] = val1 + val2
+
+    """
+    def interpolate(self, nc_filename, variable_list, itime):
```
wence-: This kind of feels wrong in that the method signature doesn't match the abstract class.  Would it not make more sense to  push the setup this function does into the constructor?  After all, this method signature suggests that you can use one `NetCDFLatLonInterpolator2d` to interpolate into some specified function space from multiple different netcdf files.  But that isn't the case, because the interpolator object is made from the first netcdf file that is passed in.

thetis/coordsys.py
tkarna: Yes this is not actually needed in any test case, so I've removed it.

thetis/coordsys.py
tkarna: fixed

thetis/coordsys.py
tkarna: Thanks. Updated function name, signature and docstrings. Yes, as you mentioned below this is only used to rotate vectors defined in lat/lon components to local x/y directions.

thetis/coordsys.py
tkarna: No it doesn't. LL_WO is no longer there.

thetis/timezone.py
tkarna: Nope. Using `pytz.utc` now.

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+LL_WGS84 = pyproj.Proj(proj='latlong', datum='WGS84', errcheck=True)
+LL_WO = pyproj.Proj(proj='latlong', nadgrids='WO', errcheck=True)  # HARN
+
+
+def convertCoords(x, y, fromSys, toSys):
+    """Converts x,y from fromSys to toSys."""
+    if isinstance(x, np.ndarray):
+        # proj may give wrong results if nans in the arrays
+        lon = np.ones_like(x) * np.nan
+        lat = np.ones_like(y) * np.nan
```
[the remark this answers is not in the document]
tkarna: changed

thetis/coordsys.py
```diff
@@ -0,0 +1,97 @@
...
+        :arg lat: lat coordinates of mesh points
+        :arg target_csys: target pyproj coordinate system object
+        """
+        self.target_csys = target_csys
+        self._compute_rotation(lon, lat)
+
+    def _compute_rotation(self, lon, lat):
+        R, theta = getVectorRotationMatrix(lon, lat, self.target_csys)
+        self.rotation_sin = np.sin(theta)
+        self.rotation_cos = np.cos(theta)
```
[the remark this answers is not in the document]
tkarna: yes, done

thetis/interpolation.py
tkarna: done

thetis/interpolation.py
tkarna: oops, fixed.

thetis/interpolation.py
tkarna: you are correct.

thetis/interpolation.py
```diff
@@ -0,0 +1,634 @@
...
+        simplex = tri.find_simplex(target_xy)
+        vertices = np.take(tri.simplices, simplex, axis=0)
+        nodes = np.unique(vertices.ravel())
+        nodes_x, nodes_y = np.unravel_index(nodes, orig_shape)
+
+        # x and y bounds for reading a subset of the netcdf data
+        ind_x = slice(nodes_x.min(), nodes_x.max() + 1)
+        ind_y = slice(nodes_y.min(), nodes_y.max() + 1)
+
+        return nodes, ind_x, ind_y
```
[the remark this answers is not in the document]
tkarna: now pulled out of the class

thetis/interpolation.py
tkarna: I have changed the abstract classes now such that `interpolate` method always takes a filename.

I'd like to avoid defining the filename in the constructor because it's generally not known at that time. If the data set is divided into multiple files, the time search objects will seek for the correct file. Then, it's then easy to pass the filename and time index to the spatial interpolator who will initialize itself if needed. The assumption, of course, is all the files have the same spatial grid.

wence-: OK, that makes sense.

tkarna:
I think I've now addressed all the points you've raised.

PR #100, opened 2017-06-25
New type-checked options

tkarna:
New type-checked options implemented with `traitlets`.

- Most options have been renamed for better human readability, e.g.
    - `dt` -> `timestep`
    - `order` -> `polynomial_degree`
    - `t_end` -> `simulation_end_time`
    - `h_viscosity` -> `horizontal_viscosity`
    - `u_advection` -> `horizontal_velocity_scale` ...
- 2D and 3D model options are in separate classes
- All options are type-checked
- Options no nonger have dict access, e.g. `options['element_family']` must be replaced by `options.element_family`
- Some options are nested:
    - E.g. setting `options.timestepper_type` changes the options class `options.timestepper_options` and its default values
    - all `solver_options` dictionaries are now under `options.timestepper_options`
    - other nested options are `options.gls_model` and `options.equation_of_state`
- Automatic time step functionality has changed
    - `options.timestep` can no longer be `None`
    - If a time stepper supports automatic time step estimation (currently all explicit schemes), there's a boolean option `options.timestepper_options.use_automatic_timestep` which defaults to `True`

Roadmap
- [x] 2D examples and tests should be revised
    - I think in most cases it's no longer necessary to specify `solver_parameters` in the input file
- [x] Documentation should be revised/updated
    - [x] Need to generate documentation for the options


tkarna:
I've added a first attempt to generate docstrings for the options. Now there's just static pages for 2D/3D/turbulence/... model options that list all the docstrings that a `Configurable` object contains. This however does not allow referring to a specific option, e.g. `ModelOptions2d.timestep` as those attributes are not properly labeled.

I think we'd need to hook up the docstring geneneration to sphinx autodoc, or disable autodoc for the options classes and manually generate nice looking, annotated doc page. Or use [traitlet-documenter](https://github.com/simphony/traitlet-documenter) as Stephan mentioned, but we'd need to extend it.

stephankramer:
I think this is good to good now (after I've unbroken the 2d tests again), so we can all move on.

I'm still struggling a little with the default solver parameters in 2d. The best defaults are very different for dg-dg vs. dg-cg (the latter we use a lot in realistic setups at AMCG). As an example I have the following timings for the stommel2d example with PressureProjectionPicard:

stommel2d, pressure projection picard, 20x20, dt=7200, 5 days:
====================
dg-cg  preonly + full schur fieldsplit (gmres+sor/gmres+hypre)     7.0
dg-cg  gmres + multiplicative fieldsplit (current default)      27.4

dg-dg  preonly + full schur fieldsplit (gmres+sor/gmres+hypre)     31.1
dg-dg  gmres + multiplicative fieldsplit (current default)       15.8

stommel2d, pressure projection picard, 40x40, dt=7200, 5 days:
==========================
dg-cg  preonly + full schur fieldsplit (gmres+sor/gmres+hypre)     14.9
dg-cg  gmres + multiplicative fieldsplit (current default)        375.

dg-dg  preonly + full schur fieldsplit (gmres+sor/gmres+hypre)     256.
dg-dg  gmres + multiplicative fieldsplit (current default)       173.

I see very similar timings for realistic cases. This makes dg-cg a lot more performant for us - but only if I set the full Schur fieldsplit options, which get very wieldy (even more so with the assembled schur that shaves off even more of the timings, and which I'd like to merge shortly). Unfortunately the Schur approach doesn't work very well for dg-dg, mainly because the selfp Schur approximation is a Laplacian with a massive stencil (siimilar happens with the assembledschur approach), so I can't really make this the default.

Would it be terrible to also change the timestepper_options (defaults) when flipping the element_family? I guess the danger is if someone does:

    options.timestepper_type = ...
    options.timestepper_options.foo = ...
    options.element_family = ...

in which case the timestepper_options would be overwritten. At the moment I don't really see any better alternative though, and I do want to get rid of having all the solver parameters options in people's setups: it should be easy for people to switch between dg-dg and dg-cg and still have reasonable performance.

tkarna:
Mmm I see. Maybe we could check what the `element_type` is when setting the default `timestepper_options`? That might require some changes to the magic Lawrence wrote.

stephankramer:
I'm doing some more timings - it might be that `PressureProjectionPicard` is not actually worth it for `dg-dg` in which case I could just use the defaults that work well for `dg-cg` for `PresureProjectionPicard`, and tell people to use `CrankNicolson` with `dg-dg` (assuming that's just as fast), and  use `PressureProjectionPicard` with `dg-cg` only.

tkarna:
Thank you very much, merging.

PR #101, opened 2017-07-15
Columbia plume

tkarna:
3D model updates for Columbia river plume application and 3D baroclinic tests


tkarna:
This PR will be split in smaller pieces.

PR #103, opened 2017-07-27
Add two videos to website frontpage.

stephankramer:
Adding two videos where Thetis is used to simulate the operation of tidal lagoons & barrages


wence-:
Pushed the updated website, will push merge when jenkins says OK.

PR #105, opened 2017-08-24
Fixes for non-deterministic iteration order

wence-:
We have to be careful that we iterate over dict-like objects in a
consistent order.  Arrange for that either by using OrderedDict or
iterating over sorted keys.


tkarna:
Added some fixes related to `PressureProjectionPicard` time integrator. @stephankramer could you sanity check, please?

tkarna:
Actually, it didn't work :) I was reverting that bit of wence-'s commit.

PR #107, opened 2017-09-22
3D solver updates

tkarna:
Updates and fixes for the 3D solver that are needed for baroclinic eddies test and Columbia River plume application.

### DIFF
--- a/thetis/rungekutta.py
+++ b/thetis/rungekutta.py
@@ -453,8 +453,8 @@ class RungeKuttaTimeIntegrator(TimeIntegrator):
     def advance(self, t, update_forcings=None):
         """Advances equations for one time step."""
         if not self._initialized:
-            error('Time integrator {:} is not initialized'.format(self.name))
-        for i in range(self.n_stages):
+            self.initialize(self.solution)
+        for i in xrange(self.n_stages):
             self.solve_stage(i, t, update_forcings)
         self.get_final_solution()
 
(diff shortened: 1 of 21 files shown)


tkarna:
I think this is good to go now (provided Jenkins gives green light). Please review.

thetis/rungekutta.py
```diff
@@ -424,8 +424,8 @@ class ESDIRKMidpointAbstract(AbstractRKScheme):
 
 class ESDIRKTrapezoidAbstract(AbstractRKScheme):
     a = [[0.0, 0.0],
-         [0.5, 0.5]]
-    b = [0.5, 0.5]
+         [0.499, 0.501]]
+    b = [0.499, 0.501]
```
stephankramer: Why?

tkarna: In the `baroclinic_channel` large-scale test a numerical mode develops in the elevation field. It's really subtle, takes more than a year to show up. My guess is that it's related to the weak instability that Crank-Nicolson scheme manifests. This is just to nudge the scheme to be a bit more diffusive - seems to solve the issue. I'll double check that it's still needed, there have been some changes in the time integrator.

tkarna: OK, the weak instability still seems to occur in the `baroclinic_channel` test. However this commit does affect the convergence rate of the free surface solver, so I'm going to drop it for now. I hope I can find another workaround for that specific test.

PR #111, opened 2017-10-12
Columbia River plume simulation

tkarna:
Updates for Columbia river plume simulation.

This example makes use of the grid interpolation routines added in #99.
Adds `uptide` as a requirement. Also allows using UFL expressions in some of the options.


thetis/configuration.py
```diff
@@ -166,6 +167,23 @@ def default_value_repr(self):
         return 'Function'
 
 
+class FiredrakeExpression(TraitType):
+    default_value = None
+    info_text = 'a UFL expression, Firedrake Constant or Function'
+
+    def validate(self, obj, value):
+        if isinstance(value, (Constant, Function, ufl.core.operator.Operator)):
```
tkarna: I'm not sure if `ufl.core.operator.Operator` is the best way of checking whether `value` is a UFL expression.

stephankramer: I think the most generic would be:

    if isinstance(value, ufl.core.expr.Expr):

this would also cover things like `uv[0]` or `0*u` (where u is a Function). Only small downside I can see is that this allow TestFunction and TrialFunction as well. We could also consider to split this into a `FiredrakeScalarExpression` and `FiredrakeVectorExpression` and use `ufl.checks.is_ufl_scalar()`.

examples/columbia_plume/tidal_forcing.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def to_latlon(x, y):
+    lon, lat = coordsys.convert_coords(coordsys.SPCS_N_OR, coordsys.LL_WGS84, x, y)
+    if lon < 0.0:
+        lon += 360.
+    return lat, lon
+
+
+class TidalBoundaryForcing(object):
```
stephankramer: This looks like it could be useful functionality to include in Thetis - we have a similar "tidal_forcing.py" that gets copied around with all our different tidal setups. I can have a go at  moving this into a Thetis utility module  and generalize it (for coordinate systems and add the option of TPXO) after this merge?

thetis/configuration.py
tkarna: Good idea, implemented.

examples/columbia_plume/tidal_forcing.py
tkarna: Yes, that sounds very good.

PR #115, opened 2017-11-08
Detector callback

stephankramer:
Add a callback to write out interpolated fields at detector locations. Based on DiagnosticCallback, writes to diagnostic_detectors.h5 file.

Also includes a helper function that can select those detector locations
that are actually within the domain, and potentially move them inside
(if within some maximum distance) in the nearest cell centre.

Change default precision from dtype='f' (single precision) to dtype='d'
(double precision). All DiagnosticCallback derived callback classes can
pass a new hdf5_dtype keyword arg to override.

### DIFF
--- a/thetis/callback.py
+++ b/thetis/callback.py
@@ -380,18 +380,19 @@ class TracerOvershootCallBack(MinMaxConservationCallback):
 
 class DetectorsCallback(DiagnosticCallback):
     """Callback that writes the specified fields interpolated in the specified detector locations"""
-    name = 'detectors'
-
     def __init__(self, solver_obj,
                  detector_locations,
                  field_names,
+                 name,
                  detector_names=None,
                  **kwargs):
         """
         :arg solver_obj: Thetis solver object
-        :list detector_locations: Locations in which fields are to be interpolated.
-        :list field_names: Fields to be interpolated.
-        :list detector_names: Names for each of the detectors. If not provided automatic names
+        :arg detector_locations: List of x, y locations in which fields are to be interpolated.
+        :arg field_names: List of fields to be interpolated.
+        :arg name: Unique name for this callback and its associated set of detectors. This determines the name
+           of the output h5 file (prefixed with diagnostic_).
+        :arg detector_names: List of names for each of the detectors. If not provided automatic names
            of the form 'detectorNNN' are created where NNN is the (0-padded) detector number.
         :arg **kwargs: any additional keyword arguments, see DiagnosticCallback
         """
@@ -415,6 +416,11 @@ class DetectorsCallback(DiagnosticCallback):
         self._variable_names = detector_names
         self.detector_locations = detector_locations
         self.field_names = field_names
+        self._name = name
+
+    @property
+    def name(self):
+        return self._name
 
     @property
     def variable_names(self):
--- a/thetis/callback.py
+++ b/thetis/callback.py
@@ -405,7 +405,7 @@ class DetectorsCallback(DiagnosticCallback):
         ndetectors = len(detector_locations)
         if detector_names is None:
             fill = len(str(ndetectors))
-            detector_names = ['detector' + str(i).zfill(fill) for i in range(ndetectors)]
+            detector_names = ['detector{:0{fill}d}'.format(i, fill=fill) for i in range(ndetectors)]
         else:
             assert ndetectors == len(detector_names), "Different number of detector locations and names"
         self._variable_names = detector_names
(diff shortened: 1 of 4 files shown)


thetis/utility.py
```diff
@@ -1577,3 +1577,63 @@ def compute_boundary_length(mesh2d):
         one_func = Function(p1).assign(1.0)
         boundary_len[i] = assemble(one_func * ds_restricted)
     return boundary_len
+
+
+def select_and_move_detectors(mesh, detector_locations, detector_names=None,
+                              maximum_distance=0.):
+    """Select those detectors that are within the domain and/or move them to the nearest cell centre within the domain."""
```
tkarna: Better docstring?

tkarna: Better docstring?

thetis/callback.py
```diff
@@ -417,8 +375,61 @@ def minmax():
...
+                }
+        super().__init__(solver_obj, array_dim=sum(field_dims), attrs=attrs, **kwargs)
+
+        ndetectors = len(detector_locations)
+        if detector_names is None:
+            fill = len(str(ndetectors))
+            detector_names = ['detector' + str(i).zfill(fill) for i in range(ndetectors)]
+        else:
+            assert ndetectors == len(detector_names), "Different number of detector locations and names"
+        self._variable_names = detector_names
```
tkarna: Hmm semantically variable names should be related to the field names? I.e. a variable is a physical quantity with a unit. (In the long run these hf5 output files should be CF compliant in some sense...)

thetis/callback.py
```diff
@@ -417,8 +375,61 @@ def minmax():
...
-                                                      solver_obj,
-                                                      outputdir=outputdir,
-                                                      export_to_hdf5=export_to_hdf5,
-                                                      append_to_log=append_to_log)
+        super(TracerOvershootCallBack, self).__init__(minmax, solver_obj, **kwargs)
+
+
+class DetectorsCallback(DiagnosticCallback):
+    """Callback that writes the specified fields interpolated in the specified detector locations"""
+    name = 'detectors'
```
tkarna: Will this work for multiple instances? Typically you do not want to extract all the variables in the same locations. E.g. in 3D you'll have different stations for temperature/salinity and elevation. If you define two instances will they try to write to the same file?

tkarna:
Looks good. Will this work in 3D out of the box? At least `select_and_move_detectors` seems to be 2D only. We can extend later. (At least the vertical coordinate typically needs some special attention in 3D).

thetis/callback.py
stephankramer: I see, good point. I suppose I could have the user specify a `name` for each detector callback?

thetis/callback.py
stephankramer: I was just trying to reuse the existing DiagnosticHDF5. I'd rather not have `f=h5py.File(); data=f['fieldname']` and `data` be a array of ntimesteps x ndetectors, i.e. not use detector names but just have them in one big array in the specified order. We typically have a large number of detectors and it gets hard to keep track of which ones, and in which order, have been used in a specific run. In particular because this tend to keep changing for subsequent runs in a specific domain as we change the mesh slightly (moving boundaries) and add more data sources. We could have `f['fieldname']['detectorname'][:]` if you prefer (or the other way around)?

tkarna: Yes, I see your point. I guess this is fine as it is. If we did `f['detectorname']['fieldname'][:]` we could attach CF metadata to the individual arrays, but at the moment `DiagnosticHDF5` class doesn't support such nesting and I'm not sure if that would be easy to implement.

thetis/callback.py
tkarna: Yes, that's probably the only way to do it. Should the name be compulsory so that users do not accidentally create two overlapping instances?

thetis/callback.py
```diff
@@ -417,8 +370,67 @@ def minmax():
...
+                 # use null-padded ascii strings, dtype='U' not supported in hdf5, see http://docs.h5py.org/en/latest/strings.html
+                 'field_names': np.array(field_names, dtype='S'),
+                 'field_dims': field_dims,
+                }
+        super().__init__(solver_obj, array_dim=sum(field_dims), attrs=attrs, **kwargs)
+
+        ndetectors = len(detector_locations)
+        if detector_names is None:
+            fill = len(str(ndetectors))
+            detector_names = ['detector' + str(i).zfill(fill) for i in range(ndetectors)]
```
wence-: AKA
```python
detector_names = ['detector{:0{fill}d}'.format(i, fill=fill) for i in range(ndetectors)]
```

PR #117, opened 2017-11-28
Time integrator convergence tests

tkarna:
Adds a convergence test that solves an ODE on a tiny mesh. As the solution only depends on time, we can check the convergence rate of all implemented time integrators.

Also removes all unused 3D time integrators.


thetis/implicitexplicit.py
```diff
@@ -94,10 +96,11 @@ def get_final_solution(self):
         """
         # set solution to u_n + sum(b*k_erk)
         self.erk.get_final_solution(additive=False)
+        # set correct reference solution in the DIRK solver
+        self.dirk.solution_old.assign(self.erk.solution)
```
stephankramer: Ah great, I see what you did: yeah, that looks fine now

thetis/timeintegrator.py
```diff
@@ -476,8 +475,11 @@ def __init__(self, equation, solution, fields, dt, bnd_conditions=None,
         self.msolution_old = Function(fs, name='dual solution')
         self.rhs_func = Function(fs, name='rhs linear form')
 
-        continuity = element_continuity(fs.ufl_element())
-        self.fs_is_dg = continuity.horizontal == 'dg' and continuity.vertical == 'dg'
+        try:
+            continuity = element_continuity(fs.ufl_element())
+            self.fs_is_dg = continuity.horizontal == 'dg' and continuity.vertical == 'dg'
+        except Exception:
```
stephankramer: What is it exactly that throws an `Exception` ? And is catching `Exception` not equivalent to a blank except, since all exceptions are derived (sub-classed) from `Exception` ? If it really is some code that throws a generic Exception (wherever that happens), could we fix that?

miklos1: Blank `except:` has been deemed "bad" by the `flake8` dictators.

stephankramer: Yes, just to clarify: I wasn't suggesting changing it to a blank except, just objecting to `except Exception` _because_ it seems equivalent to a blank except (which I agree is usually bad).

tkarna: Thanks for pointing this out. The `try` clause was there to handle cases where `element_continuity` doesn't work - admittedly bad design. I have now extended `element_continuity` function to handle the mixed function space used in the tests, so there's no need for `try` anymore.

PR #118, opened 2017-11-28
Remove Travis test setup

tkarna:
Remove Travis test setup that's now obsolete.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -6,12 +6,6 @@ def pytest_addoption(parser):
                      help="Only run tests marked for Travis")
 
 
-def pytest_runtest_setup(item):
-    not_travis = item.get_marker("not_travis")
-    if not_travis is not None and item.config.getoption("--travis"):
-        pytest.skip("Skipping test marked not for Travis")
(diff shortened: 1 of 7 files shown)


test/conftest.py
```diff
@@ -6,12 +6,6 @@ def pytest_addoption(parser):
                      help="Only run tests marked for Travis")
 
 
-def pytest_runtest_setup(item):
-    not_travis = item.get_marker("not_travis")
-    if not_travis is not None and item.config.getoption("--travis"):
-        pytest.skip("Skipping test marked not for Travis")
```
tkarna: @wence- I guess we can now remove all Travis related bits in `conftest.py`? Do we still need the `still alive` hack?

wence-: No

tkarna: Okay, I've cleaned up `conftest.py` some more.

tkarna:
I've cleaned up `conftest.py`. Any thoughts?

tkarna:
Is this good to go?

PR #119, opened 2017-11-28
Equation sign convention

tkarna:
Fixes confusion in the sign convention for `Term` and `Equation` classes.

All `Term`s have been implemented to be on the right-hand-side, but the documentation claims the opposite. This fixes the documentation to match the implementation.

Also fixes signs of some terms (3D internal pressure gradient term, 2D BathymetryDisplacementMassTerm), and simplifies the expressions for some other terms (e.g. source terms).

### DIFF
--- a/thetis/momentum_eq.py
+++ b/thetis/momentum_eq.py
@@ -152,7 +152,7 @@ class PressureGradientTerm(MomentumTerm):
 
         f = (int_pg[0]*self.test[0] + int_pg[1]*self.test[1])*self.dx
 
-        return f
+        return -f
 
 
 class HorizontalAdvectionTerm(MomentumTerm):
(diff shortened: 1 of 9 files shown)


thetis/momentum_eq.py
```diff
@@ -714,4 +707,4 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
                                   Dx(bhead, 1)*self.test[1])
             f = g_grav * grad_head_dot_test * self.dx
 
-        return f
+        return -f
```
stephankramer: Ah ok, so `int_pg` is now -g grad r ? Still not sure what the logic is...

tkarna: Yeah so my idea was that `residual` should always return a term on the RHS. But what I did indeed flips the sign of the `int_pg` field which is counter-intuitive. I have now changed this so that all the terms are on RHS and `int_pg` has the expected sign.

tkarna:
Is this good to go?

PR #122, opened 2018-01-26
Adding 2-D tracer transport in thetis

thangel:
adding the option to solve advection diffusion equation in 2D


test/tracerEq/test_h-diffusion_mes_2d.py
```diff
@@ -0,0 +1,188 @@
...
+        i += 1
+        if t >= next_export_t - 1e-8:
+            print_output('{:3d} i={:5d} t={:8.2f} s tracer={:8.2f}'.format(iexport, i, t, norm(solverobj.fields.tracer_2d)))
+            export_func()
+            next_export_t += solverobj.options.simulation_export_time
+            iexport += 1
+
+    # project analytical solultion on high order mesh
+    t_const.assign(t)
+    tracer_ana_ho.project(tracer_expr)
```
tkarna: Same comment on high order projection.

thetis/coupled_timeintegrator_2d.py
```diff
@@ -0,0 +1,147 @@
+"""
+Time integrators for solving coupled depth-averaged equations, currently shallow water equations coupled with one tracer.
```
tkarna: line too long?

thetis/tracer_eq_2d.py
```diff
@@ -0,0 +1,275 @@
+r"""
+2D advection diffusion equation for tracers.
+
+The advection-diffusion equation of tracer :math:`T` in conservative form reads
```
tkarna: is T the depth averaged tracer concentration?

thetis/tracer_eq_2d.py
```diff
@@ -0,0 +1,275 @@
...
+        elif 'un' in funcs:
+            uv_ext = funcs['un']*self.normal
+        else:
+            uv_ext = uv_in
+
+        return c_ext, uv_ext, elev_ext
+
+
+class HorizontalAdvectionTerm(TracerTerm):
+    # TODO adapt equation for 2-D case
```
tkarna: rm comment?

thetis/tracer_eq_2d.py
```diff
@@ -0,0 +1,275 @@
...
+
+    where :math:`\sigma` is a penalty parameter,
+    see Epshteyn and Riviere (2007).
+
+    Epshteyn and Riviere (2007). Estimation of penalty parameters for symmetric
+    interior penalty Galerkin methods. Journal of Computational and Applied
+    Mathematics, 206(2):843-872. http://dx.doi.org/10.1016/j.cam.2006.08.029
+
+    .. note ::
+        Note the minus sign due to :class:`.equation.Term` sign convention
```
tkarna: this note is now obsolete due to #119

thetis/tracer_eq_2d.py
```diff
@@ -0,0 +1,275 @@
...
+
+    The weak form reads
+
+    .. math::
+        F_s = \int_\Omega \sigma \phi dx
+
+    where :math:`\sigma` is a user defined scalar :class:`Function`.
+
+    .. note ::
+        Due to the sign convention of :class:`.equation.Term`, this term is assembled to the left hand side of the equation
```
tkarna: ditto

test/tracerEq/test_h-advection_mes_2d.py
```diff
@@ -0,0 +1,207 @@
...
+        i += 1
+        if t >= next_export_t - 1e-8:
+            print_output('{:3d} i={:5d} t={:8.2f} s salt={:8.2f}'.format(iexport, i, t, norm(solverobj.fields.tracer_2d)))
+            export_func()
+            next_export_t += solverobj.options.simulation_export_time
+            iexport += 1
+
+    # project analytical solultion on high order mesh
+    t_const.assign(t)
+    tracer_ana_ho.project(ana_tracer_expr)
```
tkarna: I think projection to high-order function space is no longer necessary if `tracer_expr` is pure UFL expression; `errornorm` should just work. This is a remnant of times when it didn't.

thetis/field_defs.py
```diff
@@ -64,6 +64,12 @@
     'unit': 'm s-1',
     'filename': 'Velocity2d',
 }
+field_metadata['tracer_2d'] = {
+    'name': 'Tracer',
```
tkarna: Depth averaged?

thetis/coupled_timeintegrator_2d.py
```diff
@@ -0,0 +1,147 @@
...
+                theta=self.options.timestepper_options.implicitness_theta)
+        else:
+            self.timesteppers.swe2d = self.swe_integrator(solver.eq_sw, self.fields.solution_2d,
+                                                          fields, solver.dt,
+                                                          bnd_conditions=solver.bnd_functions['shallow_water'],
+                                                          solver_parameters=self.options.timestepper_options.solver_parameters)
+
+    def _create_tracer_integrator(self):
+        """
+        Create time integrator for salinity equation
```
tkarna: tracer equation?

thetis/tracer_eq_2d.py
```diff
@@ -0,0 +1,275 @@
...
+        f = 0
+        source = fields_old.get('source')
+        if source is not None:
+            f += -inner(source, self.test)*self.dx
+        return -f
+
+
+class TracerEquation2D(Equation):
+    """
+    3D tracer advection-diffusion equation :eq:`tracer_eq` in conservative form
```
tkarna: 2D

thetis/tracer_eq_2d.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    Generic tracer term that provides commonly used members and mapping for
+    boundary functions.
+    """
+    def __init__(self, function_space,
+                 bathymetry=None, use_lax_friedrichs=True):
+        """
+        :arg function_space: :class:`FunctionSpace` where the solution belongs
+        :kwarg bathymetry: bathymetry of the domain
+        :type bathymetry: 3D :class:`Function` or :class:`Constant`
```
tkarna: rm 3D

tkarna:
One missing bit is the documentation. At least 2D model formulation should be updated. Would also be nice to have a 2D demo with tracers. Can be in another PR.

thetis/solver2d.py
```diff
@@ -212,6 +216,8 @@ def create_function_spaces(self):
             raise Exception('Unsupported finite element family {:}'.format(self.options.element_family))
         self.function_spaces.V_2d = MixedFunctionSpace([self.function_spaces.U_2d, self.function_spaces.H_2d])
 
+        self.function_spaces.Q_2d = FunctionSpace(self.mesh2d, 'DG', 1, name='Q_2d')
```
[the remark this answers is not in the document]
stephankramer: Yeah, the reason I suggested a different symbol is that I interpreted `H` as the pressure/elevation space - which in 3D has to be same as the tracer space. For depth averaged tracers, I _think_ we don't get strict conservation anyway, so perhaps we could be more flexible. Also strictly speaking I think the requirement is for the tracer space to be a subspace of the pressure space, so in the case of P1DG-P2 you could use P1 tracers.

thetis/tracer_eq_2d.py
stephankramer: It is indeed, so the tracer equation should *not* be solved in conservative form as it suggests in the comment. The implementation indeed uses the non-conservative form, and the docstring for the HorizontalAdvectionTerm does have the right formula.

thetis/solver2d.py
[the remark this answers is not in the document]
tkarna: I see, yes we already use `H_2d` for the elevation space.

test/tracerEq/test_h-advection_mes_2d.py
```diff
@@ -0,0 +1,204 @@
...
+        ti.advance(t)
+        t += solverobj.dt
+        i += 1
+        if t >= next_export_t - 1e-8:
+            print_output('{:3d} i={:5d} t={:8.2f} s salt={:8.2f}'.format(iexport, i, t, norm(solverobj.fields.tracer_2d)))
+            export_func()
+            next_export_t += solverobj.options.simulation_export_time
+            iexport += 1
+
+    # project analytical solultion on high order mesh
```
tkarna: typo: solution

test/tracerEq/test_h-diffusion_mes_2d.py
```diff
@@ -0,0 +1,182 @@
...
+        ti.advance(t)
+        t += solverobj.dt
+        i += 1
+        if t >= next_export_t - 1e-8:
+            print_output('{:3d} i={:5d} t={:8.2f} s tracer={:8.2f}'.format(iexport, i, t, norm(solverobj.fields.tracer_2d)))
+            export_func()
+            next_export_t += solverobj.options.simulation_export_time
+            iexport += 1
+
+    # project analytical solultion on high order mesh
```
tkarna: typo

test/tracerEq/test_h-advection_mes_2d.py
```diff
@@ -0,0 +1,204 @@
...
+    x0 = 0.3*lx
+    sigma = 1600.
+    xy = SpatialCoordinate(solverobj.mesh2d)
+    t_const = Constant(t)
+    ana_tracer_expr = exp(-(xy[0] - x0 - u*t_const)**2/sigma**2)
+    tracer_ana = Function(solverobj.function_spaces.H_2d, name='tracer analytical')
+
+    p1dg_ho = FunctionSpace(solverobj.mesh2d, 'DG', options.polynomial_degree + 2,
+                            vfamily='DG', vdegree=options.polynomial_degree + 2)
+    tracer_ana_ho = Function(p1dg_ho, name='tracer analytical')
```
tkarna: I guess my point on the high order projection was that `p1dg_ho` and `tracer_ana_ho` are no longer needed. You can just compute the error as `errornorm(tracer_ana_ufl_expr, tracer_sol)`. I think `/test/pressure_grad` has some examples (`degree_rise` option is also obsolete).

thangel: Oh now I see, yes I just realised that these do not make a difference! Updating it now

tkarna:
Thanks for the updates, looks good to me!

PR #127, opened 2018-04-13
Update website look

tkarna:
Some updates for the website:
- add a new full-width banner image
- add GMD paper
- move Jenkins status panel to the front page
- update fonts


tkarna:
As far as I'm concerned this is ready to go now.

PR #128, opened 2018-05-23
Tidal barrage thetis example

thangel:
Adding an example for the operation of a tidal impoundment in thetis. Let me know if I missed anything.


examples/tidal_barrage/modules/input_barrages.py
```diff
@@ -0,0 +1,55 @@
+from modules.parameterisations import *
+
+
+def initialise_barrage(Barrage_Number=1):
```
stephankramer: (Dummy) arguments should in principle be lower case. Exceptions are for instance mathematical (like Q_s) which are absolutely fine.

examples/tidal_barrage/modules/input_barrages.py
```diff
@@ -0,0 +1,55 @@
+from modules.parameterisations import *
```
stephankramer: Could you add a docstring at the top very briefly, even if just one line, describing what goes into each module?

examples/tidal_barrage/modules/input_barrages.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def initialise_barrage(Barrage_Number=1):
+    """
+    :param Barrage_Number: Number of barrages considered.
```
stephankramer: All methods and functions should have a doc string (in particular any user facing ones) with a one line description and a description of the arguments.

examples/tidal_barrage/modules/input_barrages.py
thangel: amended accordingly, added short docstrings to be more specific

PR #129, opened 2018-06-11
Callback for evaluating objective functionals

jwallwork23:
This base callback has two functionalities, as described in the following. Consider an objective functional `J = kt * ks * eta * dx * dt` for free surface displacement `eta` in the shallow water equations, where `ks` is an indicator function relating to some spatial region of interest and `kt` is an indicator function relating to a time window of interest.

Firstly, we could take the following callback to enable storage of the spatial integral values at each timestep. Temporal integration is achieved using the `quadrature` method once the simulation is complete. This provides a value for `J` on the mesh(es) used for the simulation.

```
class SWCallback(FunctionalCallback):
    """Calculates values for an objective functional which integrates free surface over a region of time and space."""
    name = 'objective functional'

    def __init__(self, solver_obj, **kwargs):
        """
        :arg solver_obj: Thetis solver object
        :arg **kwargs: any additional keyword arguments, see DiagnosticCallback
        """
        from firedrake import assemble

        def objectiveSW():
            """
            :param solver_obj: FlowSolver2d object.
            :return: objective functional value for callbacks.
            """
            mesh = solver_obj.fields.solution_2d.function_space().mesh()

            # Construct kernel function
            ks = Function(VectorFunctionSpace(mesh, "DG", 1) * FunctionSpace(mesh, "DG", 1))
            k0, k1 = ks.split()
            iA = Function(FunctionSpace(mesh, "DG", 1), name="Region of interest")
            iA.interpolate(Expression("pow(x[0] - x0, 2) + pow(x[1] - y0, 2) < r + eps ? 1 : 0",
                                      x0=0.5, y0=0.5, r=pow(0.2, 2), eps=1e-10))
            k1.assign(iA)
            kt = Constant(1.)   # In general this may not be the whole time period

            return assemble(kt * inner(ks, solver_obj.fields.solution_2d) * dx)

        super(SWCallback, self).__init__(objectiveSW, solver_obj, **kwargs)
```

The above callback will output a single float value for the objective functional. A second option is to replace the ``from firedrake import assemble`` above with a ``from firedrake_adjoint import assemble``. This has the action of instead writing the spatial integral values at each timestep to tape. Once the simulation has terminated, calling the ``quadrature`` method now assembles the objective functional object which can be used in solving time dependent adjoint equations.


thetis/callback.py
```diff
@@ -481,3 +481,49 @@ def __call__(self):
             field_vals.append(np.reshape(self._evaluate_field(field_name), (ndetectors, -1)))
 
         return np.hstack(field_vals)
+
+
+class FunctionalCallback(DiagnosticCallback):
+    """Base class for callbacks that evaluate a functional involving integrals in both time and space. This callback
```
stephankramer: It's not a base class, is it?

thetis/callback.py
```diff
@@ -481,3 +481,49 @@ def __call__(self):
...
+    can also be used to assemble time dependent objective functionals for adjoint simulations. Time integration is
+    achieved using the trapezium rule."""
+    variable_names = ['current integral', 'objective value']
+
+    def __init__(self, scalar_callback, solver_obj, **kwargs):
+        """
+        Creates error comparison check callback object
```
stephankramer: What?

thetis/callback.py
```diff
@@ -481,3 +481,49 @@ def __call__(self):
...
+
+        :arg scalar_callback: Python function that takes the solver object as an argument and
```
stephankramer: It's not actually called with the solver obj as argument

thetis/callback.py
```diff
@@ -481,3 +481,49 @@ def __call__(self):
...
+            returns a list of values of an objective functional.
+        :arg solver_obj: Thetis solver object
+        :arg **kwargs: any additional keyword arguments, see DiagnosticCallback
+        """
+        kwargs.setdefault('export_to_hdf5', False)
+        kwargs.setdefault('append_to_log', False)
+        super(FunctionalCallback, self).__init__(solver_obj, **kwargs)
+        self.scalar_callback = scalar_callback      # Evaluate functional
+        self.objective_value = [scalar_callback()]  # Store functional value
```
stephankramer: Is this necessary? The callback is called at the beginning of each timestep (see comments below)

thetis/callback.py
```diff
@@ -481,3 +481,49 @@ def __call__(self):
...
+        self.dt = solver_obj.options.timestep
+        self.t = solver_obj.simulation_time
+
+    def __call__(self):
+        value = self.scalar_callback()
+        if self.t > 0.5 * self.dt:              # Don't store the first value twice
```
stephankramer: The callback will indeed be called at the beginning of the first timestep. If this is at t=0. I would argue it should store the value of that call instead of the call when the callback is initialised, as the model is not necessarily completely initialised (initial conditions) when the callback is initialised. Also models do not necessarily start at t=0 (e.g. when starting from a checkpoint).

thetis/callback.py
```diff
@@ -481,3 +481,49 @@ def __call__(self):
...
+            self.objective_value.append(value)
+        self.t += self.dt
+
+        return value, self.objective_value
```
stephankramer: Do you have to return objective_value. I think it breaks using hdf5 output for this DiagnosticCallback, as it will be of a different length after each call.

thetis/callback.py
jwallwork23: Sorry, this is something I forgot to change from a previous version

thetis/callback.py
jwallwork23: Good point, thanks

thetis/callback.py
```diff
@@ -481,3 +481,43 @@ def __call__(self):
             field_vals.append(np.reshape(self._evaluate_field(field_name), (ndetectors, -1)))
 
         return np.hstack(field_vals)
+
+
+class FunctionalCallback(DiagnosticCallback):
```
tkarna: The name could be a bit more specific, functional sounds a bit too generic. It probably should indicate that this callback integrates (an existing) scalar callback over time.

thetis/callback.py
```diff
@@ -481,3 +481,43 @@ def __call__(self):
...
+        """
+        kwargs.setdefault('export_to_hdf5', False)
+        kwargs.setdefault('append_to_log', False)
+        super(FunctionalCallback, self).__init__(solver_obj, **kwargs)
+        self.scalar_callback = scalar_callback      # Evaluate functional
+        self.objective_value = []   # Store functional value
+        self.dt = solver_obj.options.timestep
+
+    def __call__(self):
+        value = self.scalar_callback()
```
tkarna: Here `value` is a numerical value (i.e. float), right?

thetis/callback.py
```diff
@@ -481,3 +481,43 @@ def __call__(self):
...
+        """
+        :arg scalar_callback: Python function that returns a list of values of an objective functional.
+        :arg solver_obj: Thetis solver object
+        :arg **kwargs: any additional keyword arguments, see DiagnosticCallback
+        """
+        kwargs.setdefault('export_to_hdf5', False)
+        kwargs.setdefault('append_to_log', False)
+        super(FunctionalCallback, self).__init__(solver_obj, **kwargs)
+        self.scalar_callback = scalar_callback      # Evaluate functional
+        self.objective_value = []   # Store functional value
```
tkarna: Here I'm wondering if using a python list is the best way of storing the values. It's easy to append to, but wastes some memory compared to numpy arrays.

thetis/callback.py
```diff
@@ -481,3 +481,43 @@ def __call__(self):
...
+
+    def message_str(self, *args):
+        line = '{0:s} value {1:11.4e}'.format(self.name, args[0])
+        return line
+
+    def quadrature(self):
+        func = self.objective_value
+        J = 0
+        for i in range(1, len(func)):
+            J += 0.5 * (func[i] + func[i-1]) * self.dt
```
tkarna: You could do the update in a better way. For example, store a temporary integrant value in the class, and just increment it every time:
```python
self.integrant = 0
self.old_value = None

scalar_value = self.scalar_callback()
if self.old_value is not None:
    self.integrant += 0.5 * (self.old_value + scalar_value) * self.dt
self.old_value = scalar_value
```

Then you do not need to store all the values in the `self.objective_value` list at all.

jwallwork23: Thanks very much for the suggestions. The reason I was storing the spatial integrals at each timestep is so that I could look at their time evolution afterwards. But you are right, for more general purposes, it would be better not to store them. A user could just use a custom callback if this is something they want to do.

thetis/callback.py
jwallwork23: Yes, unless adjoint is being used such as described in the example above. If we use `from firedrake_adjoint import assemble`, rather than `from firedrake import assemble` then `value` is no longer a float.

thetis/callback.py
jwallwork23: Also I suppose those values can be retrieved from the log file anyway.

thetis/callback.py
stephankramer: Even then, it's an AdjFloat - which acts like any other float (it's a subclass of). It just has additional annotation associated with it.

jwallwork23: Oh okay, I see! Thanks

thetis/callback.py
```diff
@@ -481,3 +481,39 @@ def __call__(self):
             field_vals.append(np.reshape(self._evaluate_field(field_name), (ndetectors, -1)))
 
         return np.hstack(field_vals)
+
+
+class ScalarIntegrationCallback(DiagnosticCallback):
```
tkarna: I wonder if this could be called `AccumulatorCallback`, `CumulativeCallback`, `ScalarCumulativeCallback`, or something like that? `Integration` might be misleading as it could be confused with spatial integration.

jwallwork23: Good point, thanks

PR #131, opened 2018-07-05
Advect tracer in a constant velocity field

jwallwork23:
This PR includes some minor changes enabling the user to advect a tracer without solving the shallow water equations. That is, the velocity field for advection is prescribed by the initial velocity condition.

The only extra thing that need be done by the user is to specify `solver_obj.options.tracer_only = True`.


thetis/options.py
```diff
@@ -452,6 +452,8 @@ class CommonModelOptions(FrozenConfigurable):
         None, allow_none=True, help="Source term for 2D tracer equation").tag(config=True)
     horizontal_diffusivity = FiredrakeCoefficient(
         None, allow_none=True, help="Horizontal diffusivity for tracers").tag(config=True)
+    tracer_only = Bool(
+        False, help="Hold shallow water variables in initial state and consider tracer advection in the associated (constant) velocity field").tag(config=True)
```
tkarna: Maybe it's better to add this in `ModelOptions2d` as this is not yet implemented in the 3D solver.

jwallwork23: Good point!

PR #132, opened 2018-07-13
Solve tracer equation in P1 space

jwallwork23:
This PR includes an option for choosing to solve tracer transport in P1 space, using `solver_obj.options.tracer_family = 'cg'`.

In the current DG setup, problems can arise in the presence of both high diffusivity and element anisotropy (e.g. under anisotropic mesh adaptivity).

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -524,7 +524,7 @@ class ModelOptions2d(CommonModelOptions):
 
     tracer_family = Unicode('dg', help="""
         Specify whether tracer should be continuous or discontinous
-        
+
         Choose from 'dg' and 'cg'.
         """).tag(config=True)
 
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -522,6 +522,12 @@ class ModelOptions2d(CommonModelOptions):
         Prints overshoot values that exceed the initial range to stdout.
         """).tag(config=True)
 
+    tracer_family = Unicode('dg', help="""
+        Specify whether tracer should be continuous or discontinous
+        
+        Choose from 'dg' and 'cg'.
+        """).tag(config=True)
+
 
 @attach_paired_options("timestepper_type",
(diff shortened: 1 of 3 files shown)


tkarna:
Thanks for this. If tracers are in P1 I think we would need another advection scheme. The SSPRK+slope limiter approach ensures conservation and monotonicity for P1DG tracers, but not for P1. I guess we should use SUPG or something similar, but no such scheme has been implemented yet. Maybe @stephankramer could chime in here.

tkarna:
What are the issues that arise with adaptive mesh and high diffusivity? We currently use interior penalty scheme for stabilizing diffusivity, which might not work very well for highly ansotropic mesh. Thus another approach might be to seek for an alternative method for diffusion.

jwallwork23:
I have been solving an advection-diffusion equation using the `tracer_only` branch with anisotropic mesh adaptivity and have found no problems when diffusivity is <1e-4. However, when diffusivity is greater than this (as large as 0.1 for one test problem I am considering), I often get a DIVERGED_MAX_IT error. I need to look more into to what extent increasing tolerances and max iteration counts and changing convergence criteria can help this. 

Typically, the tracer value on an (often but not always) highly anisotropic element near to the Gaussian profile being advected is significantly different from its neighbours that the mesh adaptive routine throws a lot of resolution there. In some cases this region ends up being more finely meshed than that surrounding the Gaussian. The error then comes when attempting to solve the advection-diffusion equation on this mesh.

stephankramer:
I agree, offering CG advection without any stabilisation is probably not very helpful. A better DG diffusion scheme would be useful indeed but probably requires some more work (@dham, @thomasgibson: anyone tried doing CDG/LDG with slate already?). @jwallwork23 : shall we revisit this when you're back? I can give you some pointers to implement SUPG...

jwallwork23:
@stephankramer : Yes sure, that's something I'd be interested to look at when I get back.

PR #135, opened 2018-09-05
Update 3D tests

tkarna:
Updates for 3D model test cases

- adds support for quad meshes
- change `options.use_lax_friedrichs_tracer` to `False` by default
- updates for lock exchange and baroclinic eddies test

### DIFF
--- a/examples/baroclinic_eddies/submitRuns.py
+++ b/examples/baroclinic_eddies/submitRuns.py
@@ -1,0 +1,214 @@
+"""
+Launches simulations on HPC machines.
+
+Run paramenters are inherited from the master test case. However, here all
+the parameters can take multiple values. All combinations of parameter values are
+executed:
+
+Example:
+
+python submitRuns.py -r coarse medium -dt 25 50
+
+will excecute runs
+python mytestcase.py -r coarse -dt 25
+python mytestcase.py -r coarse -dt 50
+python mytestcase.py -r medium -dt 25
+python mytestcase.py -r medium -dt 50
+
+In addition to the options defined in the test case itself this script defines
+options:
+--test    - print job submission scirpts without launching them
+--dev     - run simulation in development queue for a short time for debugging
+--verbose - print job submission scirpts and debug information
+
+Depends on batchScriptLib
+"""
+import baroclinic_eddies as thetistestcase
+from collections import OrderedDict
+import argparse
+import itertools
+import hpclauncher
+
+
+def get_timerequest(options):
+    """
+    User defined function that will return time allocation based on run options
+    """
+    default = hpclauncher.TimeRequest(2, 0, 0)
+    devel = hpclauncher.TimeRequest(0, 20, 0)
+    timedef = {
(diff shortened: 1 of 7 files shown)


thetis/options.py
```diff
@@ -372,9 +372,9 @@ class CommonModelOptions(FrozenConfigurable):
     use_lax_friedrichs_velocity = Bool(
         True, help="use Lax Friedrichs stabilisation in horizontal momentum advection.").tag(config=True)
     lax_friedrichs_velocity_scaling_factor = FiredrakeConstantTraitlet(
-        Constant(1.0), help="Scaling factor for Lax Friedrichs stabilisation term in horiozonal momentum advection.").tag(config=True)
+        Constant(1.0), help="Scaling factor for Lax Friedrichs stabilisation term in horizontal momentum advection.").tag(config=True)
     use_lax_friedrichs_tracer = Bool(
-        True, help="Use Lax Friedrichs stabilisation in tracer advection.").tag(config=True)
+        False, help="Use Lax Friedrichs stabilisation in tracer advection.").tag(config=True)
```
tkarna: My concern here is that this changes the default behavior for the 2D model as well. For 3D model, omitting the LF flux for tracers seems like the right thing to do, I'm not sure about 2D. If needed, we could specify this option separately for 2D and 3D. Any preferences?

examples/baroclinic_eddies/submitRuns.py
```diff
@@ -0,0 +1,214 @@
+"""
+Launches simulations on HPC machines.
+
+Run paramenters are inherited from the master test case. However, here all
```
stephankramer: typo

thetis/options.py
stephankramer: I think for the (limited number of) 2d cases with tracers that we've run so far, we've had LF switched off - as it's fairly diffusive. So I'm happy for this to be the default

examples/baroclinic_eddies/submitRuns.py
```diff
@@ -0,0 +1,214 @@
...
+
+will excecute runs
+python mytestcase.py -r coarse -dt 25
+python mytestcase.py -r coarse -dt 50
+python mytestcase.py -r medium -dt 25
+python mytestcase.py -r medium -dt 50
+
+In addition to the options defined in the test case itself this script defines
+options:
+--test    - print job submission scirpts without launching them
```
stephankramer: typo

examples/baroclinic_eddies/submitRuns.py
```diff
@@ -0,0 +1,214 @@
...
+--dev     - run simulation in development queue for a short time for debugging
+--verbose - print job submission scirpts and debug information
+
+Depends on batchScriptLib
```
stephankramer: What is that? To do with hpclauncher?

tkarna: yes, it should say hpclauncher instead.

PR #141, opened 2018-11-09
docs: embed animations in https

tkarna:
Fix embedded animations in webpage: all URLs should be https


wence-:
While we're here, you could also redo the Jenkins build link to https? Also, we're now serving it over 443 (so no need for the magic port):

https://jenkins.ese.ic.ac.uk

Points to the right place.

PR #142, opened 2018-11-09
Test examples

tkarna:
Add examples to test suite.

Sets an environment flag `THETIS_REGRESSION_TEST` during tests which can be used in the examples to reduce computational cost (e.g. by reducing run duration).

By default, the test routine will run all python files found under `examples/*/`. Files that should not be tested must be added in `exclude_files` list.


tkarna:
Good points, thanks.
1. Yes, we need to extend the mechanism when examples depend on forcing data. Actually columbia_plume already does which is why I've excluded the example for now. I'll deal with that later.
2. That's a good point. Moving the tidal farm test to `test_adjoint/test_adjoint_examples/` would probably be the right choice.

tkarna:
I've addressed point 2 now. Added `test_adjoint/examples/test_examples.py`.

PR #145, opened 2018-12-08
Store vtk fields to separate sub-directory

tkarna:
Storing all files in a single directory stalls the file system for large runs. For example, storing 10 output fields with 100 MPI processes and 500 time steps results in 500k files.

Now each field is stored in a separate sub-directory:
```
outputs/Salinity3d/Salinity3d.pvd
outputs/Salinity3d/Salinity3d_1.vtu
```


tkarna:
Thanks, Than!

PR #146, opened 2019-01-07
Strong residuals

jwallwork23:
Apologies for such a large PR. This is mainly for discussion as to whether what I have written seems like a good approach or not, rather than a PR to be merged soon.

For error estimation and implementation of SUPG stabilisation, we need to be able to compute strong residuals in Thetis.

This PR includes:
* Residual terms for each term in the shallow water and 2d tracer equations. These include cell residuals (which together give the strong residual for the equation) and edge residuals, which are the result of reversing the integration by parts _on each cell_. These edge residuals are needed for dual weighted error estimation, as in [Houston, Georgoulis & Hall, 2006]. The formulation is quite complicated for DG spaces, so my derivation of these expressions is ongoing work.
* Support for strong residuals in the the steady, Forward Euler and Crank Nicolson time integrators.
* Callbacks for evaluating the residuals (and error estimators) and saving associated fields to vtu.

Apologies there are so many commits, many of which I changed my mind on over time. If this PR were to be merged, I would be happy to just create a new branch with fewer, clearer commits.

Cheers!


thetis/shallowwater_eq.py
```diff
@@ -998,3 +1001,455 @@ def residual(self, label, solution, solution_old, fields, fields_old, bnd_condit
...
+
+        f = g_grav * grad(eta)
+
+        return -f
+
+    def residual_edge(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions):
+        raise NotImplementedError  # FIXME
+
+
+class HUDivResidual(ShallowWaterContinuityResidualTerm):
```
tkarna: Writing a separate residual form for each term looks like a lot of duplicated code. I wonder if there's a better way to do this (maybe not?).

thetis/shallowwater_eq.py
```diff
@@ -998,3 +1001,455 @@ def residual(self, label, solution, solution_old, fields, fields_old, bnd_condit
...
+        uv_old, eta_old = solution_old.split()
+        return self.residual_cell(label, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions)
+
+    def edge_residual(self, label, solution, solution_old, fields, fields_old, bnd_conditions):
+        uv, eta = solution.split()
+        uv_old, eta_old = solution_old.split()
+        return self.residual_edge(label, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions)
+
+
+class ShallowWaterMomentumResidual(BaseShallowWaterResidual):
```
tkarna: As this file is getting rather large it might make sense to define the strong residual equation in another file.

thetis/timeintegrator.py
```diff
@@ -5,6 +5,10 @@
 from .utility import *
 from abc import ABCMeta, abstractmethod
 
+from .shallowwater_eq import ShallowWaterMomentumResidual, FreeSurfaceResidual
+from .tracer_eq_2d import TracerResidual2D
```
tkarna: Time integrator class should not depend on any particular equation.

thetis/timeintegrator.py
```diff
@@ -68,6 +72,27 @@ def __init__(self, equation, solution, fields, dt, solver_parameters={}):
         self.solver_parameters = {}
         self.solver_parameters.update(solver_parameters)
 
+        if self.equation.__class__.__name__ == "ShallowWaterEquations":
+            self.momentum_residual = ShallowWaterMomentumResidual(
```
tkarna: The problem here is that the time integrators are aware of the equation being solved. It becomes difficult to write/maintain such if clauses for all the equations and time integrators.
Currently the time integrator classes take an `Equation` as an argument in the constructor. You could add (an optional?) residual argument as well. The correct residual object can then be created simultaneously with the equation (this happens in the solver class).

thetis/timeintegrator.py
```diff
@@ -218,6 +294,80 @@ def advance(self, t, update_forcings=None):
         for k in sorted(self.fields_old):
             self.fields_old[k].assign(self.fields[k])
 
+    def cell_residual(self, label='all'):
```
tkarna: `cell_residual` and `edge_residual` are not called anywhere. Am I missing something? They should be used in the solver, no?

thetis/callback.py
```diff
@@ -551,3 +552,197 @@ def get_val(self):
...
+        """
+        kwargs.setdefault('export_to_hdf5', False)
+        self.export_to_hdf5 = kwargs.get('export_to_hdf5')
+        super(ErrorEstimateCallback, self).__init__(solver_obj, **kwargs)
+        self.estimator = error_type
+        self.directory = solver_obj.options.output_directory
+        self.ix = 0
+        self.export_to_pvd = export_to_pvd
+        if self.export_to_pvd:
+            self.plot_file = File(self.directory + 'ErrorEstimator2d.pvd')
```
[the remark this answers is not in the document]
jwallwork23: Okay, will hold on with this part

thetis/shallowwater_eq.py
jwallwork23: Thanks, will create a new file for this

thetis/timeintegrator.py
jwallwork23: At the minute they are only used by my (external) error estimation routines, or in the callbacks

jwallwork23:
Thanks very much for the feedback, @tkarna! There is a lot to rebase and some quite major restructuring to do, so I think it would be easiest to just create a new branch. Will let you know when I have something working.

tkarna:
As strong residuals and SUPG imply major changes in the code maybe it's worthwhile to create a separate issue where we can outline the needed bits and pieces.

jwallwork23:
Okay, sounds good to me. I will close this PR. Thanks again

PR #149, opened 2019-02-01
Use 'ksponly' if linear

jwallwork23:
Just a minor thing.

I noticed that a nonlinear solver is used even when we set `options.use_nonlinear_equations = False`. By passing the solver parameter `'ksponly'` for `'snes_type'`, we can just use a linear solver.

For the '2D channel with boundary conditions' example, I did notice timing differences when this is used or not. For T=3600s, using the linear equations and averaged over 10 runs:
* Without 'ksponly': average time 1.2136s
* With 'ksponly': average time 1.0493s

### DIFF
--- a/thetis/solver.py
+++ b/thetis/solver.py
@@ -864,6 +864,7 @@ class FlowSolver(FrozenClass):
         # Avoid additional nonlinear iterations if linear equations are used
         if not self.options.use_nonlinear_equations:
             self.options.timestepper_options.solver_parameters['snes_type'] = 'ksponly'
+            self.options.timestepper_options.solver_parameters_tracer['snes_type'] = 'ksponly'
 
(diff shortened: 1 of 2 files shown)


dorugeber:
N.B. without that parameter, I assume it assembles the residual an extra time to see if a second nonlinear iteration is needed.

tkarna:
This appears to break the tests.

jwallwork23:
It seems I misunderstood. Nevermind, will close the PR. Thanks for the comments

tkarna:
In principle we could to this kind of fix for the 2D solver. I think you should just ensure that the time integrator is indeed (semi-) implicit. In 3D, the time integrator is a CoupledTimeIntegrator so the options are different.

PR #152, opened 2019-03-07
Fix boolean petsc options warnings

stephankramer:
This is to fix the warnings associated with the {'option': True} => {'option': None} transition.


tkarna:
Thanks for this. The tests should be fixed now in master so if you rebase all should go green.

tkarna:
Now `continuity3d/test_continuity_mes.py::test_setup1` fails. Also in master. Something must have changed in firedrake.

PR #154, opened 2019-03-22
Columbia plume application

tkarna:
Updates for Columbia river plume application


thetis/tracer_eq.py
```diff
@@ -179,12 +179,14 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
                     else:
                         c_in = solution
                         c_ext, uv_ext, eta_ext = self.get_bnd_functions(c_in, uv, elev, bnd_marker, bnd_conditions)
+                        # add interior tracer flux
```
stephankramer: This change makes perfect sense to me, but just out of interest: is there something that spurred this, some instability, conservation properties, etc.?

stephankramer: (Same for the change in momentum equation)

examples/columbia_plume/cre-plume.py
```diff
@@ -154,110 +187,171 @@
...
 # ocean initial conditions
-salt_roms_3d = Function(solver_obj.function_spaces.P1, name='ROMS salinity')
-temp_roms_3d = Function(solver_obj.function_spaces.P1, name='ROMS temperature')
-liveocean_interp = LiveOceanInterpolator(solver_obj.function_spaces.P1,
-                                         [salt_roms_3d, temp_roms_3d],
-                                         ['salt', 'temp'],
-                                         'forcings/liveocean/f2015.*/ocean_his_*.nc',
-                                         init_date)
-liveocean_interp.set_fields(0.0)
+# FIXME these should be CG fields
```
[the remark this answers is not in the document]
tkarna: Yeah, these forcing fields are intrinsically CG so using DG just wastes memory. When splitting the velocity to depth mean and its deviation, I'm using the same operators as in the solver. Which are for DG space... I'll remove the fixme

thetis/tracer_eq.py
tkarna: Yes, exactly, the open boundaries were really unstable with active tracers. This change makes a big difference.

tkarna:
Yes, I agree that most of the diagnostics and forcing routines are generic. Pushed them to `thetis/callback.py` and (a new module) `thetis/forcing.py`, respectively.

For the plots, I could add this example to the web page once I have the final results.

PR #155, opened 2019-04-04
SU stabilisation

jwallwork23:
I was working on stabilisation for CG tracers a couple of months ago but got sidetracked. This PR allows P1 tracers and simple SU stabilisation (with the same stabilisation coefficient as in the FEniCS demo https://github.com/FEniCS/dolfin/blob/master/demo/undocumented/advection-diffusion/python/demo_advection-diffusion.py).

Consider the following script.
```
from thetis import *

mesh = RectangleMesh(60, 10, 60, 10)
x, y = SpatialCoordinate(mesh)
P1 = FunctionSpace(mesh, "CG", 1)
P1_vec = VectorFunctionSpace(mesh, "CG", 1)

u0 = interpolate(as_vector((1., 0.)), P1_vec)
zero = Function(P1)
b = Function(P1).assign(1.)
x0 = 10.
y0 = 5.
r = 0.5
bell = conditional(
    ge((1 + cos(pi*min_value(sqrt(pow(x-x0, 2) + pow(y-y0, 2)) / r, 1.))), 0.),
    (1 + cos(pi*min_value(sqrt(pow(x-x0, 2) + pow(y-y0, 2)) / r, 1.))),
    0.)
source = interpolate(0. + bell, P1)

solver_obj = solver2d.FlowSolver2d(mesh, b)
options = solver_obj.options
options.timestep = 0.1
options.simulation_export_time = options.timestep*10
options.simulation_end_time = 50.
options.timestepper_type = 'CrankNicolson'
options.fields_to_export = ['tracer_2d']
options.solve_tracer = True
options.tracer_only = True
options.use_limiter_for_tracers = False
options.use_su_tracer = True
options.tracer_element_family = 'cg'
options.horizontal_diffusivity = Constant(0.1)
options.tracer_source_2d = source
solver_obj.assign_initial_conditions(elev=zero, uv=u0, tracer=zero)
solver_obj.bnd_functions['tracer'][1] = {'value': Constant(0.)}  # inflow condition
solver_obj.iterate()
```

At timestep 30, the tracer concentration is
su_t30
whereas with `options.use_su_tracer = False` we have
no_stab_t30

SUPG implementation to follow, if this looks okay.

### DIFF
--- a/test/tracerEq/test_h-advection_mes.py
+++ b/test/tracerEq/test_h-advection_mes.py
@@ -136,7 +136,7 @@ def run_convergence(ref_list, saveplot=False, **options):
         l2_err.append(run(r, **options))
     x_log = numpy.log10(numpy.array(ref_list, dtype=float)**-1)
     y_log = numpy.log10(numpy.array(l2_err))
-    setup_name = 'h-diffusion'
+    setup_name = 'h-advection'
 
     def check_convergence(x_log, y_log, expected_slope, field_str, saveplot):
         slope_rtol = 0.20
--- a/test/tracerEq/test_h-advection_mes_2d.py
+++ b/test/tracerEq/test_h-advection_mes_2d.py
@@ -184,14 +184,13 @@ def polynomial_degree(request):
     return request.param
 
 
-@pytest.mark.parametrize(('stepper'),
-                         [('CrankNicolson')])
-
 @pytest.fixture(params=['dg', 'cg'])
 def tracer_element_family(request):
     return request.param
 
 
+@pytest.mark.parametrize(('stepper'),
+                         [('CrankNicolson')])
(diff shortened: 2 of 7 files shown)


jwallwork23:
Thanks for the feedback! I made changes based on 1-3. In order for tests to pass, I needed to put the stabilisation checks in `create_equations`, rather than `iterate`. The modified tests above pass with `tracer_element_family = 'cg'` alone. However `test_h-advection_mes_2d` and `test_steady_adv-diff_mms_2d` both fail when and `use_su_stabilization = True`. Perhaps this indicates that SU is too diffusive?

I noticed that running `test_consistency_2d.py` fails with the default options.
Also, this test case fails under either CG alone or CG with SU.

PR #156, opened 2019-04-09
Fix 2d model Riemann fluxes

tkarna:
Minor fixes to Riemann fluxes in the 2D shallow water equations. Improves errors near curved land boundaries.

### DIFF
--- a/test/swe2d/test_thacker.py
+++ b/test/swe2d/test_thacker.py
@@ -5,7 +5,7 @@ from thetis import *
 import pytest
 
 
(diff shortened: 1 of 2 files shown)


tkarna:
This formulation follows what was done in SLIM. The fluxes are derived from the linear shallow water model (in which case you can solve the Riemann problem), see eqn (16) in Comblen et al (2010) https://doi.org/10.1002/fld.2094. The boundary terms are obtained from the same flux terms by assuming impermeable bnd conditions i.e. `eta_ext = eta_in` and `u_ext = -u_in`. We could surely experiment with other fluxes as well.

tkarna:
Okay so in the end this boiled down to removing the 0.5 factor. The test suite seems to be somewhat picky on these fluxes. I agree that the fluxes should be revisited in the future.

PR #157, opened 2019-04-11
Dockerise the Jenkins file

dham:
This PR switches the Jenkins file to use docker. This has two impacts.

1. We don't have to build firedrake every time, we just inherit the latest master build.
2. We can run tests on 12 cores instead of 4.

The combination of these two cuts the test time in half. 1/3 of the remaining test time is taken by one adjoint test. It might be worth considering if this test can be shortened.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,11 @@
 pipeline {
     agent {
-        label 'linux'
+      docker {
+        image 'firedrakeproject/firedrake-vanilla:latest'
+        label 'firedrakeproject'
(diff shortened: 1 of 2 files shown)


dham:
I must correct myself. 2/3 of the test time is the tidal farm adjoint test.

stephankramer:
On 12/04/2019 00:12, David A. Ham wrote:
> I must correct myself. 2/3 of the test time is the tidal farm adjoint test.

I'll take a look. Probably means it's gone haywire - shouldn't be that 
long a test...

Cheers
Stephan

PR #158, opened 2019-04-12
Fix/loopy changes

wence-:
Ports the custom kernels to the new kernel interface. Although we can still use string kernels, one level of indirection has been removed. I also correct some of the access descriptors.

Needs firedrakeproject/firedrake#1408 to land simultaneously.


wence-:
All but two tests passed for me:

test/barotropicChannel/test_closed_channel.py::test_closed_channel

test/continuity3d/test_continuity_mes.py::test_setup1

Any ideas?

wence-:
Huh, OK, works on Jenkins (and my laptop), but not on old (16.04) ubuntu. If everyone is happy I will hit go on Monday with everything else.

wence-:
(You needed to get firedrake-install from the relevant branch, this will go away and firedrake-update will work when I merge).

wence-:
Can you reapprove please?

PR #164, opened 2019-07-08
Tracer advection equation

mc4117:
Due to the coupled nature of Thetis, we cannot calculate the depth-averaged product **u**c required by the sediment concentration equation, but only the product of the depth-averaged **u** (from the hydrodynamic component) and the depth-averaged **c** (from the sediment transport component).

This addition allows us to add a correction factor to the tracer advection equation to account for this error. More details about this correction factor can be found in:

_Huybrechts, N., Villaret, C., & Hervouet, J. M. (2010, September). Comparison between 2D and 3D modelling of sediment transport: application to the dune evolution. In River Flow (pp. 887-893)._

### DIFF
--- a/test/tracerEq/test_h-advection_mes_2d.py
+++ b/test/tracerEq/test_h-advection_mes_2d.py
@@ -47,6 +47,9 @@ def run(refinement, **model_options):
     options.output_directory = outputdir
     options.simulation_end_time = t_end
     options.simulation_export_time = t_export
+    solverobj.create_function_spaces()
+    uv_tracer = Function(solverobj.function_spaces.U_2d, name='uv tracer')
+    options.tracer_advective_velocity = uv_tracer
     options.solve_tracer = True
     options.use_limiter_for_tracers = True
     options.fields_to_export = ['tracer_2d']
@@ -93,6 +96,8 @@ def run(refinement, **model_options):
 
     # custom time loop that solves tracer equation only
     ti = solverobj.timestepper.timesteppers.tracer
+
+    uv_tracer.assign(uv_init)
 
     i = 0
     iexport = 1
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -83,6 +83,7 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
                       'diffusivity_h': self.options.horizontal_diffusivity,
                       'source': self.options.tracer_source_2d,
                       'lax_friedrichs_tracer_scaling_factor': self.options.lax_friedrichs_tracer_scaling_factor,
+                      'tracer_advective_velocity': self.options.tracer_advective_velocity, 
                       }
             if issubclass(self.tracer_integrator, timeintegrator.CrankNicolson):
                 self.timesteppers.tracer = self.tracer_integrator(
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -65,7 +65,10 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
                 semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
                 theta=self.options.timestepper_options.implicitness_theta)
         else:
-            self.timesteppers.swe2d = self.swe_integrator(solver.eq_sw, self.fields.solution_2d, fields, solver.dt, bnd_conditions=solver.bnd_functions['shallow_water'], solver_parameters=self.options.timestepper_options.solver_parameters)
+            self.timesteppers.swe2d = self.swe_integrator(solver.eq_sw, self.fields.solution_2d,
+                                                          fields, solver.dt,
+                                            bnd_conditions=solver.bnd_functions['shallow_water'],
+                            solver_parameters=self.options.timestepper_options.solver_parameters)
 
     def _create_tracer_integrator(self):
         """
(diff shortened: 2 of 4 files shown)


thetis/options.py
```diff
@@ -466,6 +466,8 @@ class CommonModelOptions(FrozenConfigurable):
         None, allow_none=True, help="Source term for 2D continuity equation").tag(config=True)
     tracer_source_2d = FiredrakeScalarExpression(
         None, allow_none=True, help="Source term for 2D tracer equation").tag(config=True)
+    tracer_advective_velocity = FiredrakeVectorExpression(
+        None, allow_none=True, help="Velocity to used in tracer advection equation.").tag(config=True)
```
thangel: Velocity to be used in tracer advection equation

thetis/tracer_eq_2d.py
```diff
@@ -114,7 +114,9 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
         if fields_old.get('uv_2d') is None:
             return 0
         elev = fields_old['elev_2d']
-        uv = fields_old['uv_2d']
+        uv = fields_old.get('tracer_advective_velocity')
+        if uv is None:
+            uv = fields_old['uv_2d']
```
thangel: Looks good!

thetis/coupled_timeintegrator_2d.py
```diff
@@ -65,10 +65,7 @@ def _create_swe_integrator(self):
                 semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
                 theta=self.options.timestepper_options.implicitness_theta)
         else:
-            self.timesteppers.swe2d = self.swe_integrator(solver.eq_sw, self.fields.solution_2d,
-                                                          fields, solver.dt,
-                                                          bnd_conditions=solver.bnd_functions['shallow_water'],
-                                                          solver_parameters=self.options.timestepper_options.solver_parameters)
+            self.timesteppers.swe2d = self.swe_integrator(solver.eq_sw, self.fields.solution_2d, fields, solver.dt, bnd_conditions=solver.bnd_functions['shallow_water'], solver_parameters=self.options.timestepper_options.solver_parameters)
```
thangel: I'd split it visually as the lines deleted

thetis/options.py
thangel: (Correct description)

test/tracerEq/test_h-advection_mes_2d.py
```diff
@@ -47,6 +47,9 @@ def run(refinement, **model_options):
     options.output_directory = outputdir
     options.simulation_end_time = t_end
     options.simulation_export_time = t_export
+    solverobj.create_function_spaces()
```
thangel: I wonder whether it's worth adding an example on this functionality - but not essential for now 👍

thetis/options.py
tkarna: I agree that the description could be better. If I understand correctly, by default the tracer equation uses `uv_2d` field from the shallow water equations. This option allows specifying a custom velocity field instead. This should be made clear to the users here.

tkarna: Also: this option is specific to the 2D model, so I'd put it under `ModelOptions2d` class.

thetis/coupled_timeintegrator_2d.py
tkarna: I agree

thetis/coupled_timeintegrator_2d.py
```diff
@@ -83,7 +80,7 @@ def _create_tracer_integrator(self):
                       'diffusivity_h': self.options.horizontal_diffusivity,
                       'source': self.options.tracer_source_2d,
                       'lax_friedrichs_tracer_scaling_factor': self.options.lax_friedrichs_tracer_scaling_factor,
-                      }
+                      'tracer_advective_velocity': self.options.tracer_advective_velocity, }
```
tkarna: this change is not needed (I prefer the old one)

thetis/coupled_timeintegrator_2d.py
```diff
@@ -92,9 +89,7 @@ def _create_tracer_integrator(self):
                     semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
                     theta=self.options.timestepper_options.implicitness_theta)
             else:
-                self.timesteppers.tracer = self.tracer_integrator(solver.eq_tracer, solver.fields.tracer_2d, fields, solver.dt,
-                                                                  bnd_conditions=solver.bnd_functions['tracer'],
-                                                                  solver_parameters=self.options.timestepper_options.solver_parameters_tracer,)
+                self.timesteppers.tracer = self.tracer_integrator(solver.eq_tracer, solver.fields.tracer_2d, fields, solver.dt, bnd_conditions=solver.bnd_functions['tracer'], solver_parameters=self.options.timestepper_options.solver_parameters_tracer,)
```
tkarna: also here the old split is more legible

test/tracerEq/test_h-advection_mes_2d.py
```diff
@@ -47,6 +47,9 @@ def run(refinement, **model_options):
...
+    uv_tracer = Function(solverobj.function_spaces.U_2d, name='uv tracer')
+    options.tracer_advective_velocity = uv_tracer
```
tkarna: I guess it would be nice to run this test twice, with and without this option.

thetis/options.py
mc4117: I have adapted the definition and changed the option to the ModelOptions2d class. I committed these changes although it doesn't seem to have appeared in the conversation

thetis/coupled_timeintegrator_2d.py
mc4117: I have changed this and committed it to the branch

thetis/coupled_timeintegrator_2d.py
mc4117: I have changed this and committed it to the branch

PR #171, opened 2019-10-09
BCs other than free-slip for 2d tracer

jwallwork23:
As I understand it, Thetis master assumes free-slip boundary conditions for the 2d tracer equation and additional Dirichlet conditions are optional.

I am considering some test cases where I need to be able to have Dirichlet BCs on their own, as well as open boundaries where no BCs are applied.

This PR extends the 2d tracer equation so that it can account for these, along with non-trivial Neumann conditions.

My implementation isn't the most elegant for the case where the user specifies both Dirichlet and Neumann conditions (which is required for one of the tests) - it would be great if anyone has any suggestions for how to improve it.

If we are happy with this then I can do the 3d case, too.


stephankramer:
My first thoughts:
* what exactly do you want to happen for the case that "the user specified both a Dirichlet and a Neumann bc"?
* I don't like the `c_ext` in `bnd_functions` returning the Neumann value, that's not a `c` value so it shouldn't be in c_ext. I would rather just use a `funcs['neumann']` at the place where it's needed
* note that the value you specify for "neumann" is not the normal derivative, rather it's the diffusive flux \kappa\nabla c\cdot n, so perhaps we should call the boundary condition that?

jwallwork23:
Thanks for the feedback @stephankramer !
Have renamed as suggested and made it so `'value'` overrides `'diff_flux'`.

jwallwork23:
Following discussion in Thetis meeting, I am now defaulting to Neumann behaviour (to be consistent with how we have done things previously) instead of defaulting the value.

tkarna:
This looks good to me. Could you please add a test case that verifies that the diffusive flux through the boundary works correctly?

jwallwork23:
Is this simple test sufficient or do we actually want to test the value imposed by `diff_flux` BCs?

test/tracerEq/test_bcs_2d.py
```diff
@@ -0,0 +1,93 @@
...
+    solver_obj.assign_initial_conditions(uv=uv_tracer)
+    solver_obj.iterate()
+    sol = solver_obj.fields.tracer_2d
+
+    # Check boundary conditions are satisfied
+    tol = 1e-5
+    n = FacetNormal(mesh2d)
+    diff_tensor = as_matrix([[nu, 0, ],
+                             [0, nu, ]])
+    diff_flux = dot(diff_tensor, grad(sol))
```
tkarna: I guess the problem here is that you're not testing the diffusion operator implemented in Thetis.

tkarna:
I guess you could do a test with constant horizontal diffusion and zero velocity, where you just prescribe a non-zero diff flux on one boundary. Should have a simple analytical solution. We could then even do convergence tests, which would be good for further testing the integrity of the diffusion operator.

jwallwork23:
Thanks for the feedback and suggestion @tkarna. I set up a simple diffusion problem with an inhomogeneous Neumann BC as you suggested and derived the corresponding Fourier series solution. I truncate the series at 100 terms as they don't change much by that point. The convergence test checks for convergence under uniform mesh refinement for P1DG and P2DG tracers.

test/tracerEq/test_bcs_2d.py
```diff
@@ -0,0 +1,161 @@
...
+    Solving the inhomogeneous problem amounts to summing the solutions of the homogeneous problems
+    and subtracting ic.
+
+    The exact solution takes the form of a Fourier series, which we truncate appropriately.
+    """
+    lx = 10
+    x, y = SpatialCoordinate(mesh)
+
+    # Diffusivity and diffusive flux BC to impose
+    nu = 0.1
```
tkarna: These values are now defined twice,  in the fourier call and in the run method. Just use the ones in the run method?

test/tracerEq/test_bcs_2d.py
```diff
@@ -0,0 +1,161 @@
...
+    assert slope > 2, msg.format(slope)
+
+
+@pytest.fixture(params=[1, 2])
+def polynomial_degree(request):
+    return request.param
+
+
+@pytest.mark.parametrize(('stepper'),
+                         [('CrankNicolson')])
```
tkarna: Does this work with other time steppers too? If possible we should test them for better code coverage (to avoid issues like #172).

test/tracerEq/test_bcs_2d.py
```diff
@@ -0,0 +1,161 @@
+from thetis import *
+import pytest
+import numpy as np
+
+
+def fourier_series_solution(mesh, time):
+    """
+    Consider a diffusion problem with a inhomogeneous Neumann condition and zero initial condition:
+       c_t = nu * c_xx, c_x(0, t) = diff_flux, c_x(l, t) = 0, c(x, 0) = 0
```
tkarna: If you wish, you could use the math environment in the docstrings for better rendering:

```
Coefficient :math:`\alpha`:

.. math::
    \alpha = ...
```

test/tracerEq/test_bcs_2d.py
jwallwork23: As I understand it, Crank Nicolson is the only tracer timestepper currently available within the `create_timesteppers` method of `solver2d`. None of the other 2d tracer tests use anything other than Crank Nicolson. Maybe a future PR could consider extending the options and tests?

tkarna: I see, that indeed seems to be the case. Yeah, let's address this in another issue.

Issue #172, opened 2019-10-09
DIRK time integrators do not work with wetting-drying


tkarna:
Currently `test/swe2d/test_thacker.py` only tests the `CrankNicolson` time integrator.

Indeed, all `rungekutta.DIRKGeneric` based time integrators in the 2D solver (`BackwardEuler`, `DIRK22`, `DIRK33`) are not wetting-drying aware as the do not assemble the modified bathymetry term (`ShallowWaterTerm.wd_bathymetry_displacement`).

tkarna:
Fixed in #173.

PR #173, opened 2019-10-11
Fix DIRK time-steppers for wetting-drying

tkarna:
Implements `rungekutta.DIRKGenericUForm` which solves the implicit equations for each sub-solution `u^(i)` instead of the tendencies. This is compatible with the wetting-drying method. Replaces previous DIRK solvers in the 2D model. Adresses #172 .

This adds a new option to all implicit 2D schemes: `options.use_semi_implicit_linearization`. The default value is `True`, consistent with the `CrankNicolson` time stepper. Previously, the DIRK schemes did not have this feature.

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -47,8 +47,6 @@ class CrankNicolsonTimestepperOptions2d(SemiImplicitTimestepperOptions2d):
     implicitness_theta = BoundedFloat(
         default_value=0.5, bounds=[0.5, 1.0],
         help='implicitness parameter theta. Value 0.5 implies Crank-Nicolson scheme, 1.0 implies fully implicit formulation.').tag(config=True)
-    use_semi_implicit_linearization = Bool(
-        False, help="Use linearized semi-implicit time integration").tag(config=True)
 
 
 class PressureProjectionTimestepperOptions2d(TimeStepperOptions):
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -29,6 +29,8 @@ class SemiImplicitTimestepperOptions2d(TimeStepperOptions):
         'ksp_type': 'gmres',
         'pc_type': 'sor',
     }).tag(config=True)
+    use_semi_implicit_linearization = Bool(
+        False, help="Use linearized semi-implicit time integration").tag(config=True)
 
 
 class SteadyStateTimestepperOptions2d(TimeStepperOptions):
--- a/thetis/rungekutta.py
+++ b/thetis/rungekutta.py
@@ -558,6 +558,136 @@ class DIRKGeneric(RungeKuttaTimeIntegrator):
         self.update_solution(i_stage)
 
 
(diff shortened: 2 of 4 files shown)


thetis/rungekutta.py
```diff
@@ -558,6 +558,132 @@ def solve_stage(self, i_stage, t, update_forcings=None):
...
+        if i_stage == 0:
+            # NOTE solution may have changed in coupled system
+            self.solution_old.assign(self.solution)
+        if not self._initialized:
+            error('Time integrator {:} is not initialized'.format(self.name))
+        if update_forcings is not None:
+            update_forcings(t + self.c[i_stage]*self.dt)
+        self.solver[i_stage].solve()
+        if i_stage < self.n_stages - 1:
+            self.k_solver[i_stage].solve()
```
tkarna: Storing the tendency here involves another solve which is not ideal. I tried assigning `k = (u^(i) - u_old ...)/dt/a[i][i]` but that doesn't seem to work.

wence-: Mathematically doesn't work, or implementation craps out?

tkarna: The solver does not converge. I checked the norm of `k` obtained from the solver (as implemented here) and from the assignment and the fields are indeed different. I guess I'm missing something.

stephankramer: Well you need the modified mass terms in u^(i) and u_old when doing `k = (u^(i) - u_old ...)/dt/a[i][i]`, i.e. solve (but this is more of a projection):
```
<test, k> = mass_term(u^i) - mass_term(u^0) - sum_{j<i} <test, dt*k_j
```

tkarna: Ah yes, that must be it. The `\tilde h` function does not fit in the function space so it must be projected. Still the projection you proposed is probably cheaper than re-evaluating `equation.residual`.

tkarna: I've added the projection proposed by Stephan in the latest version.

thetis/options.py
```diff
@@ -29,6 +29,8 @@ class SemiImplicitTimestepperOptions2d(TimeStepperOptions):
         'ksp_type': 'gmres',
         'pc_type': 'sor',
     }).tag(config=True)
+    use_semi_implicit_linearization = Bool(
```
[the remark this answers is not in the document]
stephankramer: Also, to confirm I understand: the default is now changed to solve the full nonlinear system, right? I think this would be a reasonable (robust) default indeed...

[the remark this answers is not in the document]
tkarna: Ah, that's right: `CrankNicolsonTimestepperOptions2D` is inherited from the `SemiImplicitTimestepperOptions2d` class. I'll fix that.
Yes, in the 2D model the default behavior is now `use_semi_implicit_linearization=True` for all implicit schemes, which makes more sense.

PR #176, opened 2019-11-11
Allow user-provided SIPG parameter

jwallwork23:
I had recently been having trouble with the `SteadyState` shallow water solver failing to converge on highly anisotropic meshes. @stephankramer recommended increasing the interior penalty parameter and this turned out to work nicely.

This PR enables the user to provide their own SIPG parameter.

I have written a script to compute the parameter automatically for a given mesh and viscosity/diffusivity. Would this be useful to include, or would we prefer to just leave it to the user?


stephankramer:
> I have written a script to compute the parameter automatically for a given mesh and viscosity/diffusivity. Would this be useful to include, or would we prefer to just leave it to the user?

Yes, it could be added as utility function maybe, with the user still setting it?

thetis/tracer_eq.py
```diff
@@ -295,9 +295,11 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
             elemsize = (self.h_elem_size*(self.normal[0]**2
                                           + self.normal[1]**2)
                         + self.v_elem_size*self.normal[2]**2)
-            sigma = 5.0*degree_h*(degree_h + 1)/elemsize
-            if degree_h == 0:
-                sigma = 1.5/elemsize
+            sigma = fields_old.get('sipg_parameter')
+            if sigma is None:
+                sigma = 5.0*degree_h*(degree_h + 1)/elemsize
```
stephankramer: Note that this uses a different definition of the parameter than alpha in solver2d because here we are already dividing by a cellsize h, i.e. sigma=alpha/h (with a default of alpha=5*p*(p+1)). I think we should be setting alpha as a user parameter, as h will be spatially varying.

thetis/tracer_eq_2d.py
```diff
@@ -215,9 +215,11 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
             # sigma = 6.93 = 3.5*p*(p+1)
 
             degree_h = self.function_space.ufl_element().degree()
-            sigma = 5.0*degree_h*(degree_h + 1)/self.cellsize
-            if degree_h == 0:
-                sigma = 1.5 / self.cellsize
+            sigma = fields_old.get('sipg_parameter')
```
stephankramer: Same thing here

thetis/options.py
```diff
@@ -533,6 +533,17 @@ class ModelOptions2d(CommonModelOptions):
...
+    sipg_parameter = FiredrakeScalarExpression(None, allow_none=True, help="""
+        Parameter used in Symmetric Interior Penalty Galerkin method of Epshteyn and
+        Riviere (2007). For anisotropic meshes, it is advisable to use a larger value
+        than the default.
+
+        Epshteyn and Riviere (2007). Estimation of penalty parameters for symmetric
+        interior penalty Galerkin methods. Journal of Computational and Applied
+        Mathematics, 206(2):843-872. http://dx.doi.org/10.1016/j.cam.2006.08.029
+
+        """).tag(config=True)
```
tkarna: I'd prefer the help string to start with a short (< 1 line) description as in the case of other options. Should somehow summarize the default? Unsure if the reference should be included here. Note that these docstrings will land on https://thetisproject.org/model_options_2d.html .

tkarna:
Also, if you wish to keep the commit history cleaner, you can rebase this branch on the current master (`git rebase master`).

tkarna:
> >I have written a script to compute the parameter automatically for a given mesh and viscosity/diffusivity. Would this be useful to include, or would we prefer to just leave it to the user?

> Yes, it could be added as utility function maybe, with the user still setting it?

I agree. Could you share the function, please (in a gist or such)? In 3D I've added heuristics to set the time step which has worked quite well. Maybe we could add something similar in 2D...

jwallwork23:
Thanks both for the feedback! Will get working on your recommended changes.

For the utility function, I am currently using https://gist.github.com/jwallwork23/a927462556a297e5cea5d87f52128cb0.

Note that it assumes `nu` to be constant in each element.

tkarna:
> For the utility function, I am currently using ...

Thanks, it looks like the utility function only works for 2D triangular meshes. Thus it's probably better to indeed include it as a optional util for the user. (Is it not possible to compute the min angle without plex inteface?)

wence-:
All of that data is in the firedrake mesh. And also the plex data may be incorrect.

jwallwork23:
@wence- Oh okay. How do I associate vertex data with an element in the firedrake mesh?

jwallwork23:
@tkarna Regarding including a test, would it be sufficient to parametrise `test/tracerEq/test_h-advection_mes_2d.py` so that it checks convergence for (a) the default penalty parameter (as currently) and (b) also with the automatic penalty parameter?

tkarna:
> Regarding including a test, would it be sufficient to parametrise `test/tracerEq/test_h-advection_mes_2d.py` so that it checks convergence for (a) the default penalty parameter (as currently) and (b) also with the automatic penalty parameter?

Yeah, we should keep the current default parameter tests. You could add a test with a custom SIPG parameter and verify that a) the result is different (i.e. the parameter does something) and plausibly b) the solution still converges. I think you can wiggle the parameter a bit and still get a decent convergence.

thetis/field_defs.py
```diff
@@ -310,3 +310,27 @@
...
+    'shortname': 'Tracer SIPG',
+    'unit': '',
+    'filename': 'sipg_parameter_tracer',
+}
+field_metadata['sipg_parameter_vertical_tracer'] = {
+    'name': 'Tracer SIPG parameter in vertical',
+    'shortname': 'Tracer SIPG vertical',
+    'unit': '',
+    'filename': 'sipg_parameter_vertical_tracer',
+}
```
tkarna: It might not be very clear, but these metadata are needed for `solver.fields` Functions so that they can be exported. I don't think we need to include SIPG param in here.

thetis/coupled_timeintegrator.py
```diff
@@ -237,6 +238,8 @@ def _create_mom_integrator(self):
                   'uv_p1': self.fields.get('uv_p1_3d'),
                   'lax_friedrichs_velocity_scaling_factor': self.options.lax_friedrichs_velocity_scaling_factor,
                   'coriolis': self.fields.get('coriolis_3d'),
+                  'sipg_parameter': self.options.sipg_parameter,
```
tkarna: You are passing `solver.options.sipg_parameter` to the equations, so there's no need to add these parameters in `solver.fields`.

thetis/solver.py
```diff
@@ -551,6 +629,11 @@ def create_fields(self):
         self.fields.h_elem_size_3d = Function(self.function_spaces.P1)
         self.fields.h_elem_size_2d = Function(self.function_spaces.P1_2d)
         get_horizontal_elem_size_3d(self.fields.h_elem_size_2d, self.fields.h_elem_size_3d)
+        self.set_sipg_parameter()
+        self.fields.sipg_parameter = self.options.sipg_parameter
+        self.fields.sipg_parameter_vertical = self.options.sipg_parameter_vertical
+        self.fields.sipg_parameter_tracer = self.options.sipg_parameter_tracer
+        self.fields.sipg_parameter_vertical_tracer = self.options.sipg_parameter_vertical_tracer
```
tkarna: You don't need to add these in `self.fields` as SIPG parameter is accessed from `self.options`.

thetis/options.py
```diff
@@ -666,3 +684,11 @@ class ModelOptions3d(CommonModelOptions):
         Constant(10.0), help="Constant temperature if temperature is not solved").tag(config=True)
     constant_salinity = FiredrakeConstantTraitlet(
         Constant(0.0), help="Constant salinity if salinity is not solved").tag(config=True)
+    sipg_parameter_vertical = FiredrakeScalarExpression(
+        Constant(10.0), help="Penalty parameter used for vertical viscosity terms.").tag(config=True)
+    sipg_parameter_vertical_tracer = FiredrakeScalarExpression(
+        Constant(10.0), help="Penalty parameter used for vertical diffusivity terms.").tag(config=True)
+    sipg_parameter_turb = FiredrakeScalarExpression(
```
tkarna: I'm not sure if the turb values are really needed. If I recall correctly, this specialization did not exits before.

thetis/solver.py
```diff
@@ -442,6 +443,78 @@ def create_function_spaces(self):
...
+        """
+        degree_h, degree_v = self.function_spaces.U.ufl_element().degree()
+        alpha_h = 5.0*degree_h*(degree_h+1) if degree_h != 0 else 1.5
+        alpha_v = 5.0*degree_v*(degree_v+1) if degree_v != 0 else 1.0
+        degree_h_tracer, degree_v_tracer = self.function_spaces.H.ufl_element().degree()
+        alpha_h_tracer = 5.0*degree_h_tracer*(degree_h_tracer+1) if degree_h_tracer != 0 else 1.5
+        alpha_v_tracer = 5.0*degree_v_tracer*(degree_v_tracer+1) if degree_v_tracer != 0 else 1.0
+        degree_h_turb, degree_v_turb = self.function_spaces.turb_space.ufl_element().degree()
+        alpha_h_turb = 5.0*degree_h_turb*(degree_h_turb+1) if degree_h_turb != 0 else 1.5
+        alpha_v_turb = 5.0*degree_v_turb*(degree_v_turb+1) if degree_v_turb != 0 else 1.0
```
tkarna: So here you need separate values for the turb, tracer, and momentum, as p may be different. If, instead, set_sipg_parameter is a member of the equation class, there's no need for such specialization.

jwallwork23:
Thanks for the feedback @tkarna. I'll make the sipg parameter an attribute of the equation class.

tkarna:
Thinking about this more, as the fields may have different polynomial degree, the user must be able to set different constant values for each. So the specialization is needed. So I think this is fine.

Issue #177, opened 2019-12-12
get_facet_mask is broken


tkarna:
`utility.get_facet_mask` appears to be broken, breaking the 3D model. Below is a minimal example:

```python
from firedrake import *
mesh2d = UnitSquareMesh(10, 10)
mesh = ExtrudedMesh(mesh2d, 10, 0.1)
V = FunctionSpace(mesh, 'DG', 1)
from thetis import get_facet_mask
get_facet_mask(V, 'geometric', 'top')
# returns [0, 1, 2, 3, 4, 5]
get_facet_mask(V, 'geometric', 'bottom')
# also returns [0, 1, 2, 3, 4, 5]
# expected: the top/bottom nodes of a prism, e.g. [0, 1, 2] and [3, 4, 5]
```

tkarna:
It looks like this is a consequence of Firedrake now defaulting to spectral elements (firedrakeproject/firedrake#1537). The following works in the minimal example:
```
helem = FiniteElement('DG', mesh2d.ufl_cell(), 1, variant="equispaced")
velem = FiniteElement('DG', interval, 1, variant="equispaced")
elem = TensorProductElement(helem, velem)
V = FunctionSpace(mesh, elem)
```

tkarna:
Fixed in #178.

PR #178, opened 2019-12-13
Use equispaced elements in 3D model

tkarna:
By default, Firedrake now uses spectral elements for which the current 2D-3D function mapping does not work. All 3D function spaces are therefore set to "equispaced" variant.

### DIFF
--- a/test/firedrake/test_implicit_friction.py
+++ b/test/firedrake/test_implicit_friction.py
@@ -5,7 +5,7 @@ Tests implicit bottom friction formulation
 Intended to be executed with pytest.
 """
 from firedrake import *
-from thetis.utility import get_functionspace_3d
+from thetis.utility import get_functionspace
 import numpy as np
 import time as time_mod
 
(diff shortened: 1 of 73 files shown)


tkarna:
This PR adds the following helper functions which create equispaced function spaces by default:
- `utility.get_functionspace_2d`
- `utility.get_functionspace_3d`

tkarna:
Note that I have not updated the 2D model as it is unclear whether the default should be equispaced or spectral.

tkarna:
Fixes #177.

test/firedrake/test_implicit_friction.py
```diff
@@ -35,14 +36,14 @@ def test_implicit_friction(do_export=False, do_assert=True):
 
     # ----- define function spaces
     deg = 1
-    p1dg = FunctionSpace(mesh, 'DG', degree=1, vfamily='DG', vdegree=1)
-    p1dgv = VectorFunctionSpace(mesh, 'DG', degree=1, vfamily='DG', vdegree=1)
-    u_h_elt = FiniteElement('RT', triangle, deg + 1)
-    u_v_elt = FiniteElement('DG', interval, deg)
+    p1dg = get_functionspace(mesh, 'DG', 1)
+    p1dgv = get_functionspace(mesh, 'DG', 1, vector=True)
+    u_h_elt = FiniteElement('RT', triangle, deg + 1, variant='equispaced')
```
dham: Note that this doesn't actually do anything since we don't have spectral on triangles.

tkarna: Yeah, true.

PR #180, opened 2020-01-17
Dirk22 and norm smoother

mc4117:
An extra option was also added to the solver object to smooth the velocity norm in the friction term in the shallow water equations. This is to avoid non-differentiable functions in the continuous formulation and is discussed in section 2.1 in "Funke, S., Farrell, P., & Piggott, M. (2017). Reconstructing wave profiles from inundation data. Computer Methods in Applied Mechanics and Engineering, 322, 167–186. Retrieved from https://arxiv.org/pdf/1612.09597.pdf". The default is that this smoother = 0 unless set by the user and thus all tests should be unaffected by this change.

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -462,7 +462,8 @@ class CommonModelOptions(FrozenConfigurable):
     norm_smoother = FiredrakeConstantTraitlet(
         Constant(0.0), help=r"""
         Coefficient used to avoid non-differentiable functions in the continuous formulation of the velocity norm in
-        the drag term in the momentum equation
+        the quadratic bottom drag term in the momentum equation. This replaces the velocity norm in the quadratic
+        bottom drag term with :math:`\|u\| \approx \sqrt{\|u\|^2 + \alpha^2}`
         """).tag(config=True)
     horizontal_viscosity = FiredrakeScalarExpression(
         None, allow_none=True, help="Horizontal viscosity").tag(config=True)
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -459,6 +459,11 @@ class CommonModelOptions(FrozenConfigurable):
 
         Bottom stress is :math:`\tau_b/\rho_0 = -g \mu^2 |\mathbf{u}|\mathbf{u}/H^{1/3}`
         """).tag(config=True)
+    norm_smoother = FiredrakeConstantTraitlet(
+        Constant(0.0), help=r"""
+        Coefficient used to avoid non-differentiable functions in the continuous formulation of the velocity norm in
+        the drag term in the momentum equation
+        """).tag(config=True)
     horizontal_viscosity = FiredrakeScalarExpression(
         None, allow_none=True, help="Horizontal viscosity").tag(config=True)
     coriolis_frequency = FiredrakeScalarExpression(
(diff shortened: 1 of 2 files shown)


tkarna:
Thanks for this. It appears that the first 3 commits have already been merged to master. You should rebase your branch on current master (you probably will see some conflicts which you'll need to fix manually). In addition, it seems that you have accidentally added a file `.DS_Store` which should be removed. It might be easier to just take the current master and plug in the options.py and shallowwater_eq.py files that you have modified in the last commit.

mc4117:
Thanks for your comment! I have rebased the branch so now the only commit is the addition of the norm-smoother option and also removed the DS_Store file.

thetis/options.py
```diff
@@ -459,6 +459,11 @@ class CommonModelOptions(FrozenConfigurable):
 
         Bottom stress is :math:`\tau_b/\rho_0 = -g \mu^2 |\mathbf{u}|\mathbf{u}/H^{1/3}`
         """).tag(config=True)
+    norm_smoother = FiredrakeConstantTraitlet(
+        Constant(0.0), help=r"""
+        Coefficient used to avoid non-differentiable functions in the continuous formulation of the velocity norm in
```
stephankramer: It would be helpful to add the formula from the paper, something like:
```
This replaces the velocity norm in the quadratic bottom drag term with :math:`\|u\| \approx \sqrt{\|u\|^2 + \alpha^2}`
```
so people have an idea what sort of value to put, i.e. a *small* velocity with units of m/s.

thetis/shallowwater_eq.py
```diff
@@ -688,7 +688,7 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
             C_D = g_grav * manning_drag_coefficient**2 / total_h**(1./3.)
 
         if C_D is not None:
-            f += C_D * sqrt(dot(uv_old, uv_old)) * inner(self.u_test, uv) / total_h * self.dx
+            f += C_D * sqrt(dot(uv_old, uv_old) + (self.options.norm_smoother)**2) * inner(self.u_test, uv) / total_h * self.dx
```
stephankramer: Just to nitpick: these brackets are superfluous

mc4117:
Thanks for your comments Stephan. I have made the changes you suggested and added them as a new commit

tkarna:
This PR has 28 commits for changing 7 lines of code... But there are essentially only two commits in which you make meaningful changes. You could create a new branch instead (I'm calling it foo) and cherry-pick those two commits on top:
```
git chechout master
git checkout -b foo
git cherry-pick cf7dbf5 027db9e
```
Now, if this branch looks good, you could just override the PR (caution: this destroys the old branch!):
```
git checkout DIRK22_and_norm_smoother
git reset --hard foo
git push -f
git branch -d foo
```
I'm suggesting this because reviewing the pull requests gets quite involved if the commit history is all tangled up with commits that have already been merged/reverted. It's a good practice to learn to keep your branches tidy.

mc4117:
Thanks for the suggestion Tuomas I've changed the branch so that you only see the relevant commits

PR #181, opened 2020-01-17
Split take 2

mc4117:
Some very small changes to ensure that the split function works correctly with thetis adjoint


tkarna:
Here also the history is quite scrambled up. You are essentially only changing 4 lines in timeintegrator.py so I'd rather have those changes on top of the current master. Plus there are two `.DS_Store` files included.

thetis/timeintegrator.py
```diff
@@ -205,7 +205,10 @@ def initialize(self, solution):
         self.solution_old.assign(solution)
         # assign values to old functions
         for k in sorted(self.fields_old):
-            self.fields_old[k].assign(self.fields[k])
+            if type(self.fields_old) == Function:
+                self.fields_old[k].project(self.fields[k])
```
tkarna: I don't see why this change is necessary. By construction fields[k] and fields_old[k] live in the same function space so assign is equivalent to project. Just faster.

mc4117: Ok I noticed that the adjoint seemed to work slightly better using this approach but it may have been a coincidence based on other changes I made so given what you say I am happy to ignore this change

stephankramer: This _should_ be exactly the same indeed (also for the adjoint) - if it's not (I would redo your Taylor test to check) then we should fix it in pyadjoint rather than workaround in Thetis

mc4117: I checked the taylor test and the normalised differences between the two approaches are O(10^(-14)) so I'm guessing this is just machine precision?

stephankramer: Ah yes, when I said "exactly the same" - that is within some precision: the projection does a (fairly trivial) solve Mx = Mb where M is the mass matrix, whereas assign is a straight copy x=b.

mc4117: Ok I will close this pull request then and delete the branch as it seems like the changes are irrelevant. Most of the split block changes happened in https://github.com/mc4117/pyadjoint/tree/fixing_split

PR #189, opened 2020-03-11
Conservative tracer 2d

mc4117:
Stephan, Than and I have been built a conservative tracer equation into thetis. This was done because when the wetting and drying scheme is switched on, in the non-conservative case, the model seemed to be leaking tracer. We have shown that with the conservative scheme this effect is reduced. Any comments or suggestions would be most welcome.


thetis/callback.py
```diff
@@ -299,6 +299,53 @@ def message_str(self, *args):
         return line
 
 
+class TracerScalarConservationCallback(DiagnosticCallback):
+    """Base class for callbacks that check conservation of a scalar quantity"""
+    variable_names = ['integral', 'relative_difference']
+
```
jwallwork23: I guess you have copied and pasted this from `ScalarConservationCallback`, so it would be good if you changed the docstring.

thetis/options.py
```diff
@@ -498,13 +510,14 @@ class ModelOptions2d(CommonModelOptions):
...
+    use_tracer_conservative_form = Bool(False, help='Solve 2D tracer transport in the conservative form').tag(config=True)
     use_wetting_and_drying = Bool(
         False, help=r"""bool: Turn on wetting and drying
 
         Uses the wetting and drying scheme from Karna et al (2011).
         If ``True``, one should also set :attr:`wetting_and_drying_alpha` to control the bathymetry displacement.
         """).tag(config=True)
-    wetting_and_drying_alpha = FiredrakeConstantTraitlet(
+    wetting_and_drying_alpha = FiredrakeScalarExpression(
         Constant(0.5), help=r"""
```
jwallwork23: I guess this change is part of a different PR (#186)?

tkarna:
It looks like this PR sits on top of #180 so we should sort that out first.

thetis/callback.py
mc4117: Good spot! I've updated it now

thetis/options.py
mc4117: Sorry I forgot I'd merged that branch. I don't actually need it so I've changed it in the branch

tkarna: Instead of reverting the changes by hand it would be better to clean up the commit history. You'll need to rebase this branch, so you'll hit this issue again at least once again.

PR #190, opened 2020-03-13
Corrective vel factor

mc4117:
As part of my conservative tracer 2d I have created this separate pull request where I have changed the implementation of the corrective velocity factor (see previous pull request https://github.com/thetisproject/thetis/pull/164). This is because I realised that the correction factor also needed to be applied to the boundary conditions which made a difference in the case of the conservative tracer.

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -543,8 +543,8 @@ class ModelOptions2d(CommonModelOptions):
 
         Prints deviation from the initial mass to stdout.
         """).tag(config=True)
-    tracer_advective_velocity = FiredrakeVectorExpression(None, allow_none=True,
-                                                          help="Custom function to be used for the velocity variable in tracer advection equation").tag(config=True)
+    tracer_advective_velocity_factor = FiredrakeScalarExpression(Constant(1.0), 
+                                                          help="Custom factor which is multiplied to the velocity variable in tracer advection equation").tag(config=True)
     check_tracer_overshoot = Bool(
         False, help="""
         Compute tracer overshoots at every export
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -543,8 +543,10 @@ class ModelOptions2d(CommonModelOptions):
 
         Prints deviation from the initial mass to stdout.
         """).tag(config=True)
-    tracer_advective_velocity_factor = FiredrakeScalarExpression(Constant(1.0), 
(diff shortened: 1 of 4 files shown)


thetis/options.py
```diff
@@ -543,8 +543,10 @@ class ModelOptions2d(CommonModelOptions):
 
         Prints deviation from the initial mass to stdout.
         """).tag(config=True)
-    tracer_advective_velocity = FiredrakeVectorExpression(None, allow_none=True,
-                                                          help="Custom function to be used for the velocity variable in tracer advection equation").tag(config=True)
+    tracer_advective_velocity_factor = FiredrakeScalarExpression(
+        Constant(1.0), help="""
+        Custom factor which is multiplied to the velocity variable in tracer advection equation
+        """).tag(config=True)
```
jwallwork23: Maybe you could include a sentence on when it might be useful to use this parameter?

tkarna:
Looking at the Huybrechts et al. (2010) paper, I guess you are defining the correction factor in the application py file? And this correction factor has been derived for sediments. i.e. it's not for any generic tracer.

mc4117:
@jwallwork23  Yes that's correct. It's basically just changing a previous pull request (https://github.com/thetisproject/thetis/pull/164)
@tkarna Yes that's also correct. Unfortunately there isn't just a default corrective factor that you can use for all tracer applications and even for sediment transport it's dependent on all sorts of variables.

jwallwork23:
I don't know a lot about tracer parametrisations. However, following from my previous observation, maybe we could just have both? i.e. `uv = alpha*uv + beta`? With defaults `alpha=1`, `beta=0`. Just an idea.

thetis/options.py
tkarna: If the help string is long it's a good idea to split it into a short description + new line  + long description. E.g., in here https://github.com/thetisproject/thetis/blob/master/thetis/options.py#L405

mc4117: I have updated this with your suggestion

mc4117:
I am happy to include it if necessary but I think a factor like the beta one you're talking about would probably normally go into something like the source term. I am also rewriting how thetis does source terms in my next pull request so I think it might be best to leave this for now?

jwallwork23:
@mc4117 Fair enough, sounds good :)

PR #192, opened 2020-03-19
Update 3d bottom friction

tkarna:
Updates related to 3D model bottom friction:
- Fix implicit bottom friction formulation. Avoids velocity sign change in the bottom element.
- Adds surface/bottom Ekman layer tests and examples
- Bottom roughness is now a field in options (instead of a physical constant)
- Remove obsolete parabolic eddy viscosity option
- Remove obsolete `compute_bottom_friction` method


test/bottomFriction/test_ekman_bottom.py
```diff
@@ -0,0 +1,116 @@
...
+    solver_obj.create_equations()
+
+    v_init_2d = -0.49
+    solver_obj.assign_initial_conditions(uv_2d=Constant((0, v_init_2d)))
+
+    x, y, z = SpatialCoordinate(solver_obj.mesh)
+
+    solver_obj.iterate()
+
+    if verify and os.getenv('THETIS_REGRESSION_TEST') is None:
```
stephankramer: This is a little odd: what is this test testing that is not tested by the example (through test_examples.py)? and why do we need THETIS_REGRESSION_TEST logic here?

test/bottomFriction/test_ekman_surface.py
```diff
@@ -0,0 +1,115 @@
...
+
+    dt = 90.0
+    t_end = 4 * 3600.0  # sufficient to reach ~steady state
+    t_export = 450.0
+    u_mag = 1.0
+
+    f_coriolis = 1e-4
+    nu_v = 5e-4
+
+    if os.getenv('THETIS_REGRESSION_TEST') is not None:
```
stephankramer: Same here

test/bottomFriction/test_ekman_surface.py
```diff
@@ -0,0 +1,115 @@
...
+    uv_ana = Function(solver_obj.function_spaces.P1DGv, name='solution')
+    uv_ana.interpolate(uv_ana_expr)
+
+    # initialize with a linear v profile to speed-up convergence
+    v_init_expr = conditional(z > -d, a*(1 + z_s), 0)
+    solver_obj.assign_initial_conditions(uv_3d=as_vector((v_init_expr/3, -v_init_expr, 0)))
+
+    solver_obj.iterate()
+
+    if verify and os.getenv('THETIS_REGRESSION_TEST') is None:
```
stephankramer: And here

test/bottomFriction/test_ekman_bottom.py
tkarna: Oh, thanks for pointing that out. I must have copied that over from the examples.

PR #193, opened 2020-03-24
small fix

mc4117:
Following this pull request https://github.com/thetisproject/thetis/pull/190 I realised there was a small error in the corrective factor implementation. Previously the line change was being multiplied by self.corr_factor twice (once in line 120 and then in line 92 of thetis/tracer_eq_2d.py). This change removes this double multiplication and thus corrects the code.

### DIFF
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -92,11 +92,7 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
                                                                   semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
                                                                   theta=self.options.timestepper_options.implicitness_theta)
             else:
-                self.timesteppers.tracer = self.tracer_integrator(solver.eq_tracer,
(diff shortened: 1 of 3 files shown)


mc4117:
I have now also added an implementation error if the use tries to use a timestepping scheme which is not crank nicolson for the coupled shallow water-tracer problem

thetis/coupled_timeintegrator_2d.py
```diff
@@ -92,11 +92,7 @@ def _create_tracer_integrator(self):
                                                                   semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
                                                                   theta=self.options.timestepper_options.implicitness_theta)
             else:
-                self.timesteppers.tracer = self.tracer_integrator(solver.eq_tracer,
```
thangel: Do we have to remove this structure altogether? I assumed that this the error is a temporary fix until we expand this section. I have no strong feelings either way

mc4117: That's true. i could just comment it out instead if that would be better?

PR #194, opened 2020-03-26
Rossby soliton test case

jwallwork23:
This PR adds an extra test for the shallow water solver, taken from p.3-6 of [1].

An initial two peak Rossby soliton propagates around a non-dimensionalised periodic equatorial channel. The test is inviscid and is set up such that the solution at time T=120 should be identical to the initial condition. (See below for initial conditions presented on a high resolution mesh.)

The test case admits approximate solutions by asymptotic expansion, so we can compute error metrics at the final time to assess how accurately Thetis handles dispersion and nonlinearity. The error metrics chosen are relative mean peak height, relative mean phase speed and RMS error. These are chosen so that we can compare against results presented in [1] for FVCOM and ROMS. In order to compute the error metrics, we project onto the same high resolution mesh used for FVCOM comparisons.

I only included two low resolution runs in the pytest, so that this test case doesn't add too much time to the Thetis test suite. Pytest just verifies that all of the metrics improve upon uniform mesh refinement.

If the test is run using python (rather than pytest), it will use the same high resolution meshes as in [1], save the outputs to HDF5 and then present the error metrics in a Markdown friendly table. _[I am currently running the tests on the finest mesh and will add them here when finished.]_

There are two reasons I wanted to consider this test case:
* It is nice to compare Thetis with FVCOM and ROMS.
* I intend to implement ALE in Thetis and this would be good for testing it.

[1] H. Huang, C. Chen, G.W. Cowles, C.D. Winant, R.C. Beardsley, K.S. Hedstrom and D.B. Haidvogel, "FVCOM validation experiments: Comparisons with ROMS for three idealized barotropic test problems" (2008), Journal of Geophysical Research: Oceans, 113(C7).

### Initial elevation

boyd_init_elev

### Initial velocity

boyd_init_velocity


test/swe2d/test_rossby_wave.py
```diff
@@ -0,0 +1,342 @@
...
+        polynomials.append(2*y*polynomials[i-1] - 2*(i-1)*polynomials[i-2])
+
+    # First order terms
+    u_terms += C*φ*0.5625*(3 + 2*y*y)*Ψ
+    u_terms += φ*φ*Ψ*sum(u[i]*polynomials[i] for i in range(28))
+    v_terms += dφdx*φ*Ψ*sum(v[i]*polynomials[i] for i in range(28))
+
+    return interpolate(as_vector([u_terms, v_terms]), U_2d)
+
+def asymptotic_expansion_elev(H_2d, order=1, time=0.0, soliton_amplitude=0.395):
```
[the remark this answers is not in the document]
jwallwork23: Thanks, that's really useful. Have reformatted accordingly

test/swe2d/test_rossby_wave.py
```diff
@@ -0,0 +1,342 @@
...
+    solver_obj.assign_initial_conditions(uv=uv_a, elev=elev_a)
+
+    # Solve PDE and save to HDF5
+    solver_obj.iterate()
+    fname = "{:d}_{:d}".format(refinement_level, mesh2d.comm.size)
+    di = create_directory(os.path.join(os.path.dirname(__file__), 'outputs'))
+    with DumbCheckpoint(os.path.join(di, fname), mode=FILE_CREATE) as chk:
+        chk.store(solver_obj.fields.elev_2d)
+
+    physical_constants['g_grav'].assign(9.81)  # TODO: Is this needed?
```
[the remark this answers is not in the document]
jwallwork23: I think the reason my previous attempt (https://github.com/thetisproject/thetis/tree/rossby_soliton_test) didn't work was because I assigned `g_grav` as a global variable. I guess pytest was remembering this in later tests, causing them to fail.

Maybe it will be okay here because it is being modified within a function. We shall see in the current test, I guess!

test/swe2d/test_rossby_wave.py
```diff
@@ -0,0 +1,342 @@
...
+        solver_obj.bnd_functions['shallow_water'][tag] = {'uv': Constant(as_vector([0., 0.]))}
+
+    # Apply initial conditions (asymptotic solution at initial time)
+    uv_a = asymptotic_expansion_uv(solver_obj.function_spaces.U_2d, order=order)
+    elev_a = asymptotic_expansion_elev(solver_obj.function_spaces.H_2d, order=order)
+    solver_obj.assign_initial_conditions(uv=uv_a, elev=elev_a)
+
+    # Solve PDE and save to HDF5
+    solver_obj.iterate()
+    fname = "{:d}_{:d}".format(refinement_level, mesh2d.comm.size)
```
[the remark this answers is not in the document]
jwallwork23: So the reason I use HDF5 checkpointing is because I got to the end of one run, projected the solution onto the reference mesh and then just got `Killed`, probably because the reference space was pushing my memory limits. I just didn't want to lose the solution again.

But I agree that it is a bit of a weird thing to do in a test. Will modify now.

test/swe2d/data/FVCOM.json
```diff
@@ -0,0 +1 @@
+{"h+": [0.85, 0.967, 0.996], "h-": [0.818, 0.963, 0.996], "c+": [0.991, 0.999, 1.0], "c-": [0.991, 0.999, 1.0], "rms": [0.0035165, 0.000609, 0.00010149], "dx": [0.25, 0.125, 0.05], "dt": [0.01, 0.005, 0.002]}
```
tkarna: Maybe in these file names you could somehow indicate that these files are related to the Rossby solition test case?

jwallwork23:
I realised the parallel partitioning doesn't work for very fine reference meshes (see general on slack), so am currently modifying the test to account for this.

test/swe2d/test_rossby_wave.py
```diff
@@ -0,0 +1,400 @@
...
+    solver_obj = solver2d.FlowSolver2d(mesh2d, bathymetry2d)
+    options = solver_obj.options
+    options.timestepper_type = stepper
+    options.timestep = 0.96/refinement_level if model_comparison or stepper == 'SSPRK33' else 9.6/refinement_level
+    options.simulation_export_time = 5.0
+    options.simulation_end_time = T
+    options.use_grad_div_viscosity_term = False
+    options.use_grad_depth_viscosity_term = False
+    options.horizontal_viscosity = None
+    options.no_exports = True and refinement_level <= 48
```
tkarna, line 182: True and conditional = conditional. AFAIK you could drop the conditional as this gets overridden anyway

test/swe2d/test_rossby_wave.py
```diff
@@ -0,0 +1,400 @@
...
+    return out+'\n'
+
+
+# ---------------------------
+# standard tests for pytest
+# ---------------------------
+
+@pytest.mark.parametrize('stepper,family,rms',
+                         [
+                             ('CrankNicolson', 'dg-dg', [1.1522e-02, 4.3892e-03]),
```
tkarna, line 356: Not sure if it makes sense to test that the rmse is allclose to these values. I'd imagine those smaller digits will wiggle as firedrake/petsc etc is updated. So far we've only tested error < target + tolerance. We could change the strategy to indeed test that the errors do not change, but this could trigger a lot of test failures in the future. Don't know if @stephankramer has any thoughts on this.

stephankramer: I agree - small variations are to be expected due to solver tolerances, parallel, and in particular with unstructured meshes (not here) mesh generation. In my experience having tests fail if the error changes at all, just leads to a lot of update work which is then blindly done thus defeating the purpose. In this case the errors are all pretty much the same, so I would just check for an error< 1.3 for all - arguably dg-cg is a _little_ more accurate.

jwallwork23:
Thanks @tkarna @stephankramer for taking a look at this. I made the suggested modifications.

PR #195, opened 2020-03-26
Spatially varying SIPG parameter

jwallwork23:
This PR extends the automatic SIPG parameter generation implemented in #176 so that it is allowed to vary in space.

SIPG parameters for all equations (momentum/swe/tracer) default to `Constant`s, but can be set as P0 fields. Where the previous PR computed the minimum angle in all elements and multiplied it by the maximum ratio of viscosity or diffusivity (as appropriate) in all elements, this PR actually takes the elementwise values, in line with [Epshteyn & Riviere, 2007].

### DIFF
--- a/thetis/solver.py
+++ b/thetis/solver.py
@@ -511,14 +511,14 @@ class FlowSolver(FrozenClass):
                         max_sipg = self.options.sipg_parameter_turb.vector().gather().max()
                         print_output("Maximum turbulence SIPG value in horizontal: {:.2f}".format(max_sipg))
                 else:
-                    if self.options.solve_salinity or self.options.solve_temperature
+                    if self.options.solve_salinity or self.options.solve_temperature:
                         print_output("Tracer SIPG parameter in horizontal: {:.2f}".format(alpha_h_tracer))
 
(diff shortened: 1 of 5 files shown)


thetis/solver.py
```diff
@@ -460,9 +461,6 @@ def set_sipg_parameter(self):
             X = \frac{\max_{x\in K}(\nu(x))}{\min_{x\in K}(\nu(x))},
 
         :math:`p` the degree, and :math:`\theta_K` is the minimum angle in the element.
-
-        In practice, we take the maximum value of :math:`X` and minimum value of
-        :math:`\alpha_K` over all elements.
         """
         degree_h, degree_v = self.function_spaces.U.ufl_element().degree()
         alpha_h = 5.0*degree_h*(degree_h+1) if degree_h != 0 else 1.5
```
tkarna: In the 2d model you've made all the alphas Constants. Maybe here too?

jwallwork23:
Thanks @tkarna. Have rebased and made all alphas `Constant`s.

jwallwork23:
Fixed print statement error, good to go once approved :)

PR #197, opened 2020-03-31
Prescribed velocity test

jwallwork23:
Here's another simple test which will be useful for testing ALE in 2d (when I get round to implementing it).

A Gaussian tracer concentration profile is advected in: (a) a uniform velocity field; and (b) a velocity field whose y-component varies sinusoidally with time. The domain is periodic and the problem is inviscid, set up such that the exact solution at t=10 is identical to the initial condition (under both velocity fields).

The convergence test iteratively doubles the mesh resolution and halves the timestep, checking the resulting L2 error converges with slope 2.

Once ALE is implemented, we should see that advecting the mesh in the prescribed velocity field (i.e. the Lagrangian case) leads to much smaller errors.

### DIFF
--- a/test/tracerEq/test_bcs_2d.py
+++ b/test/tracerEq/test_bcs_2d.py
@@ -149,13 +149,15 @@ def run_convergence(**model_options):
     assert slope > 2, msg.format(slope)
 
 
-@pytest.fixture(params=[1, 2])
+@pytest.fixture(params=[1])
 def polynomial_degree(request):
     return request.param
 
 
-@pytest.mark.parametrize(('stepper'),
-                         [('CrankNicolson')])
+@pytest.fixture(params=['CrankNicolson', 'ForwardEuler', 'SSPRK33'])
+def stepper(request):
+    return request.param
+
 @pytest.mark.parametrize(('diffusivity'),
                          [(Constant(0.1))])
 def test_horizontal_advection(polynomial_degree, stepper, diffusivity):
--- a/test/tracerEq/test_bcs_2d.py
+++ b/test/tracerEq/test_bcs_2d.py
@@ -154,7 +154,7 @@ def polynomial_degree(request):
     return request.param
 
 
-@pytest.fixture(params=['CrankNicolson', 'ForwardEuler', 'SSPRK33'])
+@pytest.fixture(params=['CrankNicolson', 'SSPRK33', 'ForwardEuler', 'BackwardEuler', 'DIRK22', 'DIRK33'])
 def stepper(request):
     return request.param
 
--- a/test/tracerEq/test_prescribed_velocity.py
+++ b/test/tracerEq/test_prescribed_velocity.py
@@ -55,6 +55,8 @@ def run(refinement_level, velocity_type, **model_options):
 
 
 def run_convergence(ref_list, velocity_type, **options):
+    """Runs test for a list of refinements and computes error convergence rate."""
+    setup_name = 'prescribed-velocity'
     l2_err = []
     for r in ref_list:
         l2_err.append(run(r, velocity_type, **options))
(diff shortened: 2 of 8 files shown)


jwallwork23:
The Jenkins tests did pass but aren't reported here for some reason.

tkarna:
You could plausibly extend this test case similarly to what was done with the Rossby wave test: add more time integrators, function spaces, and also check the upper bound of L2(?) error.

tkarna:
Yeah, ok, extending the tracer model to support other time integrators could be done in this PR or later.

jwallwork23:
I seem to have successfully hooked up the explicit timesteppers `'ForwardEuler'` and `'SSPRK33'` for the 2d tracer model.

Using them, the test from this PR passes, along with `test_h-advection_2d`, `test_h-diffusion_2d` and `test_bcs_2d`. I didn't modify `test_consistency_2d`, as it would need reformatting. I found that the two extra timesteppers failed setup 2 in `test_steady_adv_diff_2d`, so I guess we'll need to look into that.

If what I did looks okay, I can extend it to whichever implicit timesteppers might be appropriate and possibly also the `'SteadyState'` solver. :)

test/tracerEq/test_bcs_2d.py
```diff
@@ -149,13 +149,15 @@ def run_convergence(**model_options):
     assert slope > 2, msg.format(slope)
 
 
-@pytest.fixture(params=[1, 2])
+@pytest.fixture(params=[1])
```
jwallwork23, line 152: The tracer space is hard-coded as P1DG so this didn't actually do anything.

thetis/options.py
```diff
@@ -16,6 +16,10 @@ class TimeStepperOptions(FrozenHasTraits):
 class ExplicitTimestepperOptions(TimeStepperOptions):
     """Options for explicit time integrator"""
     use_automatic_timestep = Bool(True, help='Set time step automatically based on local CFL conditions.').tag(config=True)
+    solver_parameters_tracer = PETScSolverParameters({
+        'ksp_type': 'gmres',
+        'pc_type': 'sor',
+    }).tag(config=True)
```
jwallwork23, line 22: This was just a naive choice, but it seems to work

test/tracerEq/test_prescribed_velocity.py
```diff
@@ -0,0 +1,122 @@
+"""
+Simple 2D tracer advection problem in a doubly periodic domain with prescribed fluid velocities.
```
tkarna, line 2: This test case advects a Gaussian bump in a periodic channel with a constant or sinusoidal (in time) velocity. In the end, L2 error is computed. Note that `test_h-advection_mes_2d.py` already advects a (1D) Gaussian and computes the L2 error and convergence rate. As in principle this test does not add to the test coverage, it might be a better idea to extend `test_h-advection_mes_2d.py` instead: add periodic mesh, 2D Gaussian source and optionally sinusoidal velocity. Our test suite is already quite heavy so we shouldn't add unnecessary tests.

thetis/coupled_timeintegrator_2d.py
```diff
@@ -78,21 +82,38 @@ def _create_tracer_integrator(self):
...
-                                                                  semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
-                                                                  theta=self.options.timestepper_options.implicitness_theta)
+                self.timesteppers.tracer = self.tracer_integrator(
+                    solver.eq_tracer, solver.fields.tracer_2d, fields, solver.dt,
+                    bnd_conditions=solver.bnd_functions['tracer'],
+                    solver_parameters=self.options.timestepper_options.solver_parameters_tracer,
+                    semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
+                    theta=self.options.timestepper_options.implicitness_theta,
+                )
+            elif issubclass(self.tracer_integrator, rungekutta.SSPRK33) or \
```
tkarna, line 101: It's somewhat unfortunate that we need these if statements. The idea of a unified timeintegrator API is that one could just directly construct `self.tracer_integrator` object without any checking... Now the optional arguments seem to cause problems. Still, there should be a more succinct way to do this.

thetis/solver2d.py
```diff
@@ -299,7 +299,7 @@ def create_equations(self):
         self.eq_sw.bnd_functions = self.bnd_functions['shallow_water']
         if self.options.solve_tracer:
             self.fields.tracer_2d = Function(self.function_spaces.Q_2d, name='tracer_2d')
-            if self.options.timestepper_type == 'CrankNicolson':
+            if self.options.timestepper_type in ('CrankNicolson', 'SSPRK33', 'ForwardEuler', 'BackwardEuler', 'DIRK22', 'DIRK33', 'CrankEuler'):
```
tkarna, line 302: In my opinion, testing for `timestepper_type` does not belong to to `create_equations` function, would fit more naturally in `create_timestepper`. Just create the equation if `solve_tracer` is set.

thetis/solver2d.py
```diff
@@ -340,69 +340,92 @@ def create_timestepper(self):
...
             'volume_source': self.options.volume_source_2d, }
         self.set_time_step()
+        if self.options.timestepper_type == 'CrankEuler' and not self.options.solve_tracer:
+            self.options.timestepper_type = 'CrankNicolson'
         if self.options.timestepper_type == 'SSPRK33':
-            self.timestepper = rungekutta.SSPRK33(self.eq_sw, self.fields.solution_2d,
-                                                  fields, self.dt,
-                                                  bnd_conditions=self.bnd_functions['shallow_water'],
-                                                  solver_parameters=self.options.timestepper_options.solver_parameters)
+            if self.options.solve_tracer:
```
tkarna, line 346: This chain of if clauses is too long. There must be a shorter way of doing this. A good rule in programming is to avoid code duplication (always!). At least it seems that transposing the `timestepper_type` and `solve_tracer` checks would make sense as the coupled integrator construction is always the same.

jwallwork23:
Thanks for the feedback @tkarna. I'll close this PR, re-think the timestepper implementation and make a new one without the additional test.

PR #198, opened 2020-04-03
Un-xfail adjoint tests

jwallwork23:
Two of the adjoint tests were xfailed as a temporary measure due to changes in Firedrake. Following recent fixes in Firedrake, the tests now pass and can be included in the test suite again.

### DIFF
--- a/examples/tidalfarm/tidalfarm.py
+++ b/examples/tidalfarm/tidalfarm.py
@@ -25,9 +25,6 @@ from pyadjoint.optimization.optimization import minimise
 import numpy
 op2.init(log_level=INFO)
 
-test_gradient = False  # whether to check the gradient computed by the adjoint (see note below)
-optimise = True
-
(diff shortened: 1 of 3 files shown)


tkarna:
Looking at #185, I think you missed one in `test_adjoint/examples/test_examples.py`.

stephankramer:
For completeness, this was fixed in https://github.com/firedrakeproject/firedrake/pull/1635

Yes, the test_adjoint/examples/test_examples.py still needs looking at. When I tried it, it seemed to take forever - not sure yet whether an adjoint bug (e.g. solver parameters), or that it just needs more trimming. Will take another look...

stephankramer:
Thanks. No, I'm doing more cleaning up of the adjoint in https://github.com/thetisproject/thetis/tree/adjoint-cleanup to take advantage of the results of the adjoint sprint. I have already shortened the example when run as a test.

PR #200, opened 2020-04-07
Automatic wetting and drying alpha

jwallwork23:
This PR allows the wetting and drying alpha parameter to be set automatically as a P1 field.

It follows the recommendation in [Karna et al. 2011]:
  \alpha \approx |L_x \nabla h|,
where L_x is the horizontal length scale of mesh elements at the wet-dry interface and h is the bathymetry.

An approximation is used here:
  element_width_x * abs(h_x) + element_width_y * abs(h_y).

The value may be capped using `options.wetting_and_drying_alpha_max`, which is set to 2 by default.

### DIFF
--- a/test/swe2d/test_thacker.py
+++ b/test/swe2d/test_thacker.py
@@ -14,16 +14,16 @@ from thetis import *
 import pytest
 
 
-@pytest.mark.parametrize("stepper,n,dt,alpha,max_err",
+@pytest.mark.parametrize("stepper,n,dt,max_err",
                          [
-                             ('BackwardEuler', 10, 600., 2., 0.33),
-                             ('BackwardEuler', 25, 300., 2., 0.19),
-                             ('CrankNicolson', 10, 600., 2., 0.26),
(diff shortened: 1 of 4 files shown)


tkarna:
Thanks for introducing this. The definition of the alpha parameter looks legit. My only concern is that the p0 alpha field can be quite noisy in the case of a real bathymetry where grad(bath) varies a lot from element to element. This could cause problems e.g. in the iterative solver. Also physically the "porosity" of the bottom should be smoothly varying in (x, y). So we'd need to somehow smooth the alpha field. Not sure if mapping to, say, p1 function space would be sufficient. Could also consider Laplacian diffusion, but that would(?) introduce another parameter. Maybe someone could try this out in a realistic test case (with rough bathymetry)?

jwallwork23:
Thanks for the feedback @tkarna. I agree that it would be good to consider a realistic test case to investigate this further. Will ask around.

tkarna (approved):
Thanks, looks all good.

PR #203, opened 2020-04-09
Reduce reference mesh refinement level for Rossby wave test

jwallwork23:
Running `test_rossby_wave.py` for just SSPRK33 and dg-dg uses
* ~1.75GB at the original reference_refinement_level=768
* ~273MB at the reduced reference_refinement_level=96
The test also runs a lot quicker at the lower refinement level.

Given that this test has recently caused Jenkins to crash, reducing unnecessarily high resolution would probably be a good idea.

### DIFF
--- a/test/swe2d/test_rossby_wave.py
+++ b/test/swe2d/test_rossby_wave.py
@@ -251,6 +251,6 @@ def family(request):
 
 
 def test_convergence(stepper, family):
-    run_convergence([12, 24], timestepper_type=stepper,
+    run_convergence([24, 48], timestepper_type=stepper,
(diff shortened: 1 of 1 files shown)


jwallwork23:
@tkarna I haven't seen anything as yet. Will rebase my other branches on top of it to see if they encounter the same Jenkins crash we have been seeing.

tkarna:
If you don't see new failures, go ahead an merge.

jwallwork23:
Tests go green on Jenkins here, for PR 200 and for `ale-tracer-only`, `passive-tracer-example` and `anisotropic-cell-measure` branches. Whilst this doesn't guarantee it's fixed, it looks promising.

PR #204, opened 2020-04-15
Update swe/test_anisotropic

jwallwork23:
When automatic SIPG parameter was first introduced in #176, it came with a steady SW test defined on an anisotropic mesh which models two tidal turbines in a channel.

The test case is an old version of the setup used in a paper which we just had accepted. I updated the setup to match the final version and realised that for some FE pairs the default SIPG parameter does work (although not for `'rt-dg'`). The test now considers all three FE pairs and checks that either (a) auto-SIPG converges and default doesn't; or (b) both converge, but auto-SIPG requires fewer nonlinear solver iterations.

For `'dg-dg'` and `'dg-cg'`, I found that using auto-SIPG means half as many nonlinear iterations.


test/swe2d/test_anisotropic.py
```diff
@@ -30,83 +50,107 @@ def run_steady_turbine(**model_options):
...
     solver_obj.create_equations()
 
-    # Apply boundary conditions
+    # apply boundary conditions
     solver_obj.bnd_functions['shallow_water'] = {
-        1: {'uv': Constant([3.0, 0.0])},
-        2: {'elev': Constant(0.0)},
-        3: {'un': Constant(0.0)},
+        1: {'uv': inflow_velocity},  # inflow condition used upstream
+        2: {'elev': Constant(0.0)},  # Dirichlet condition at outflow to close system
```
tkarna, line 78: The elev boundary condition is imposed weakly, so it's not really a Dirichlet condition.

jwallwork23:
Thanks for the review, @tkarna

PR #207, opened 2020-04-17
Timesteppers other than Crank-Nicolson for 2D tracer

jwallwork23:
Currently the 2D tracer equation only supports `'CrankNicolson'` timestepping. This PR is a rewrite of PR #197 which attempted to extend to five other schemes, but involved quite a bit of code duplication.

The updated version reduces duplication and provides a `CoupledMatchingTimeIntegrator2D` class which implies matching integrators for both shallow water and tracer solvers. The following additional schemes are considered: ForwardEuler, BackwardEuler, DIRK22, DIRK33, SSPRK33.

Statuses of 2D tracer tests following these changes:
* `test_h-advection_mes_2d.py`: all pass except BackwardEuler
* `test_h-diffusion_2d.py`: all pass
* `test_bcs_2d.py`: all pass
* `test_consistency_2d.py`: all pass if overshoot tol loosened from `1e-12` to `1e-11` and conservation tol loosened from `1e-4` to `1.2e-4`.
* `test_steady_adv-diff_mms_2d.py`: all pass except ForwardEuler and SSPRK33

### DIFF
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -37,42 +37,13 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
 
         self._create_integrators()
 
-    def _create_swe_integrator(self):
-        """
-        Create time integrator for 2D system
-        """
-        self.timesteppers.swe2d = self.solver.get_swe_timestepper(self.swe_integrator)
-
-    def _create_tracer_integrator(self):
-        """
-        Create time integrator for tracer equation
-        """
-        solver = self.solver
-
-        if self.solver.options.solve_tracer:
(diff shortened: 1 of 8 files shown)


thetis/coupled_timeintegrator_2d.py
```diff
@@ -14,12 +14,10 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
     """
     __metaclass__ = ABCMeta
 
-    @abstractproperty
     def swe_integrator(self):
         """time integrator for the shallow water equations"""
         pass
 
-    @abstractproperty
```
jwallwork23, line 22: This was required for the way I have implemented `CoupledMatchingTimeIntegrator2D`, but there might be another way.

PR #209, opened 2020-04-28
Update accumulator callback

jwallwork23:
As it currently stands, `AccumulatorCallback` time integrates a user-provided scalar callback using the trapezium rule. If the timestepper is `'CrankNicolson'` with theta being a half then this is the right thing to do. In other cases, it would be better to use the quadrature routine which matches the timestepper.

In this PR, the original `AccumulatorCallback` becomes a more general tool for summing quantities evaluated at each timestep (or export). A new callback. `TimeIntegratorCallback`, is introduced which does the quadrature for most currently implemented timesteppers.

### DIFF
--- a/thetis/callback.py
+++ b/thetis/callback.py
@@ -529,15 +529,13 @@ class AccumulatorCallback(DiagnosticCallback):
     """
     Callback that sums values of time-dependent functionals
     contributed at each timestep.
-
-    This callback can also be used to express time-dependent objective
-    functionals for adjoint simulations.
     """
-    variable_names = ['spatial integral at current timestep']
+    variable_names = ['integral over current timestep']
 
     def __init__(self, functional_callback, solver_obj, **kwargs):
         """
-        :arg functional_callback: Python function that returns an integral in time and space
+        :arg functional_callback: Python function with a single argument that evaluates an integral
+            of the prognostic solution in space
         :arg solver_obj: Thetis solver object
         :arg kwargs: any additional keyword arguments, see DiagnosticCallback
         """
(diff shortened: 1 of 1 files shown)


thetis/callback.py
```diff
@@ -523,43 +527,120 @@ def __call__(self):
...
+        try:
+            assert num_updates == num_weights
+        except AssertionError:
+            msg = "Update and weight vectors have different lengths ({:d} vs. {:d})"
+            raise ValueError(msg.format(num_updates, num_weights))
+
+        def time_integral_callback():
+            """Time integrate spatial integral over a single timestep"""
+            integral = 0
+            for i in range(num_weights):
```
tkarna, line 636: A bunch of `self.` references missing, e.g. `num_weights` is undefined.

thetis/callback.py
```diff
@@ -523,43 +527,120 @@ def __call__(self):
...
+                stored = i == 0 and hasattr(self, 'previous_update')
+                update_integral = self.previous_update if stored else spatial_integral(updates[i])
```
tkarna, line 638: Have you tested this? It should be possible to design a test where the analytical solution is known (constant in time is easy). As far as I can see, `update_integral` is wrong. In some cases `updates` holds the tendencies while, presumably, the `spatial_integral` should be evaluated using the solution itself. What if user wants to evaluate a time integral of a diagnostic field? Then it would not be sufficient to track just the solution. I wonder if it would be better to extend the time integration API to allow for user-defined callback functions do this? The time integrator could then handle time-integration if necessary(?). I would also like to create a better API for user-defined diagnostic fields. `ExportManager.add_export` already takes optional `preproc_func` argument which can be used to compute dianostics but this should be extended. We could think what the API could look like.

jwallwork23: I am using it for some tsunami simulations at the minute and it works for the integrators I have tried. But no, I haven't done any more extensive testing than that.

Good point - I suppose it would be better to have the time integrators do the quadrature themselves. I will have a think and maybe we can discuss on Friday.

PR #211, opened 2020-05-22
Use 2D Functions in 3D equations

tkarna:
Adds `ExtrudedFunction` class which allows using 2D fields in extruded 3D mesh:

```python
mesh2d = ...
mesh3d = ExtrudedMesh(mesh2d, ...)
fs_2d = FunctionSpace(mesh2d, ...)
f_2d = ExtrudedFunction(fs_2d, mesh_3d=mesh3d)  # this behaves like a 2D Function
f_3d = f_2d.view_3d  # this can be used in 3D equations, shares the data with f_2d
```
The 3D function lives in Foo x R extruded function space.

Under the hood there's `extend_function_to_3d` function:
```python
f_3d = extend_function_to_3d(f_2d, mesh_3d)
```
In this PR, `ExtrudedFunction` is used for elevation and bathymetry fields, reducing the total memory footprint of the 3D solver.

The following auxiliary fields are removed:
- `bathymetry_3d`
- `elev_3d`
- `uv_p1_3d`
- `uv_mag_3d`
- `w_mesh_surf_2d`
- `w_mesh_surf_3d`
- `elev_cg_3d` is renamed to `elev_domain_3d`

### DIFF
--- a/thetis/momentum_eq.py
+++ b/thetis/momentum_eq.py
@@ -179,8 +179,6 @@ class HorizontalAdvectionTerm(MomentumTerm):
     def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=None):
         if not self.use_nonlinear_equations:
             return 0
-        uv_p1 = fields_old.get('uv_p1')
-        uv_mag = fields_old.get('uv_mag')
         lax_friedrichs_factor = fields_old.get('lax_friedrichs_velocity_scaling_factor')
 
         uv_depth_av = fields_old.get('uv_depth_av')
@@ -207,13 +205,7 @@ class HorizontalAdvectionTerm(MomentumTerm):
                   + uv_up[1]*uv_av[1]*jump(self.test[1], self.normal[1]))*(self.dS_v + self.dS_h)
             # Lax-Friedrichs stabilization
             if self.use_lax_friedrichs:
-                if uv_p1 is not None:
-                    gamma = 0.5*abs((avg(uv_p1)[0]*self.normal('-')[0]
-                                     + avg(uv_p1)[1]*self.normal('-')[1]))*lax_friedrichs_factor
-                elif uv_mag is not None:
-                    gamma = 0.5*avg(uv_mag)*lax_friedrichs_factor
-                else:
-                    raise Exception('either uv_p1 or uv_mag must be given')
+                gamma = 0.5*abs(un_av)*lax_friedrichs_factor
                 f += gamma*(jump(self.test[0])*jump(uv[0])
                             + jump(self.test[1])*jump(uv[1]))*(self.dS_v + self.dS_h)
             for bnd_marker in self.boundary_markers:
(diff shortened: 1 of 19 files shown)


thetis/solver.py
```diff
@@ -870,37 +865,30 @@ def create_equations(self):
...
         self.copy_uv_dav_to_uv_dav_3d = ExpandFunctionTo3d(self.fields.uv_dav_2d, self.fields.uv_dav_3d,
                                                            elem_height=self.fields.v_elem_size_3d)
         self.copy_uv_to_uv_dav_3d = ExpandFunctionTo3d(self.fields.uv_2d, self.fields.uv_dav_3d,
                                                        elem_height=self.fields.v_elem_size_3d)
-        self.uv_mag_solver = VelocityMagnitudeSolver(self.fields.uv_mag_3d, u=self.fields.uv_3d)
         self.mesh_updater = ALEMeshUpdater(self)
 
         if self.options.use_smagorinsky_viscosity:
-            self.smagorinsky_diff_solver = SmagorinskyViscosity(self.fields.uv_p1_3d, self.fields.smag_visc_3d,
+            self.smagorinsky_diff_solver = SmagorinskyViscosity(self.fields.uv_3d, self.fields.smag_visc_3d,
```
stephankramer, line 885: Here, you also seem to drop the use of a P1CG-projected velocity. If I look at the `SmagorinskyViscosity` code I can believe that works with `weak_form=True` - but the logic here seems to only use the weak form for P0. For P0 it's obvious that the non-"weak form" case doesn't work, as the derivatives are zero - but for higher order DG (P1DG), the solution may still be piecewise constant, e.g. in a backwards facing step and I presume you still require viscosity at the shear interface?

thetis/momentum_eq.py
```diff
@@ -207,13 +205,7 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
...
             # Lax-Friedrichs stabilization
             if self.use_lax_friedrichs:
-                if uv_p1 is not None:
-                    gamma = 0.5*abs((avg(uv_p1)[0]*self.normal('-')[0]
-                                     + avg(uv_p1)[1]*self.normal('-')[1]))*lax_friedrichs_factor
-                elif uv_mag is not None:
-                    gamma = 0.5*avg(uv_mag)*lax_friedrichs_factor
-                else:
-                    raise Exception('either uv_p1 or uv_mag must be given')
+                gamma = 0.5*abs(un_av)*lax_friedrichs_factor
```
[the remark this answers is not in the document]
tkarna: Yeah this bit is quite old in fact. At first I kept the uv_mag field around and then also added the p1 projection at some point. But for the LF flux you only need an estimate of the velocity magnitude so the mean of the normal velocity should be sufficient. I haven't seen any significant difference in the solution.

thetis/solver.py
tkarna: Yeah, that's right. In the 3D model we're always using a discontinuous velocity field so it's indeed better to always use the weak form. I've changed this now.

PR #214, opened 2020-06-12
Fix cached coordinate functions in VTK outputs

### LINKED ISSUE #208: VTK output for fields on continuous functionspaces is broken
It seems that for any field that is defined on a continuous function space the corresponding vtk output produces broken vtu files. I think this is due to the self-proclaimed hack (line 123 of thetis/exporter.py) to avoid storing a separate dg coordinates function in every `File` object. If I comment out that line, things seem to work again. This is presumably because vtu files with only continuous fields do not have a discontinuous mesh. Is the fix to just add a `and not is_cg(function)` ?

Reproducer:
```
from thetis import *
from thetis.exporter import ExportManager
  
mesh = UnitSquareMesh(2,2)

function_spaces = [
        FunctionSpace(mesh, "DG", 0, name="P0"),
        FunctionSpace(mesh, "CG", 1, name="P1"),
        FunctionSpace(mesh, "CG", 2, name="P2"),
        FunctionSpace(mesh, "DG", 1, name="P1DG"),
        FunctionSpace(mesh, "DG", 2, name="P2DG")]

fields = {}
field_meta_data = {}
for fs in function_spaces:
    name = fs.name + "_field"
    field = Function(fs, name=name)
    fields[name] = field
    field_meta_data[name] = {'shortname': name, 'filename': name}

em = ExportManager('outputs', fields.keys(), fields, field_meta_data, export_type='vtk')
em.export()
```

tkarna:
ExportMananger now uses CoordinateFunctionCache which stores custom coordinate
function in case mesh coordinates are incompatible.

Fixes #208.

### DIFF
--- a/thetis/exporter.py
+++ b/thetis/exporter.py
@@ -64,7 +64,8 @@ class VTKExporter(ExporterBase):
     """Class that handles Paraview VTK file exports"""
     def __init__(self, fs_visu, func_name, outputdir, filename,
                  next_export_ix=0, project_output=False,
-                 coords_dg=None, verbose=False):
+                 add_subdir=True,
+                 coord_func=None, verbose=False):
         """
         :arg fs_visu: function space where input function will be cast
             before exporting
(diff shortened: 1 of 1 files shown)


tkarna:
We've decided to omit coordinate function caching both in firerake and thetis.

PR #218, opened 2020-07-24
Fix rt variant

stephankramer:
variant='equispaced' is no longer allowed with Hdiv elements (previously ignored). Should be variant='point' (old default, which under-integrates) or variant='integral' (soon to be new default). For the moment I've fixed the variant to 'point' for RT.

FWIW, I've locally run all tests that mention 'rt-dg' with variant='integral', and I get 3 failures:

2 failures in test/bottomFriction/test_ekman_bottom.py that look like they might just be a tolerance tweak

1 failure in test/swe2d/test_atmospheric_pressure.py where I get a DIVERGED_LINEAR_SOLVE

### DIFF
--- a/thetis/utility.py
+++ b/thetis/utility.py
@@ -111,9 +111,17 @@ class FieldDict(AttrDict):
 
 
 def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
-                      vector=False, hdiv=False, variant='equispaced', **kwargs):
+                      vector=False, hdiv=False, variant=None, **kwargs):
     gdim = mesh.geometric_dimension()
     assert gdim in [2, 3]
+    if variant is None:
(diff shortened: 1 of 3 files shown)


thetis/utility.py
```diff
@@ -111,17 +111,25 @@ def __setattr__(self, key, value):
 
 
 def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
-                      vector=False, hdiv=False, variant='equispaced', **kwargs):
+                      vector=False, hdiv=False, variant=None, **kwargs):
     gdim = mesh.geometric_dimension()
     assert gdim in [2, 3]
+    if variant is None:
+        if h_family.upper() == 'RT' or h_family.upper() == 'Raviart-Thomas':
```
tkarna, line 118: I think `upper` would return `'RAVIART-THOMAS'`, maybe `lower` is nicer?

PR #222, opened 2020-09-09
Custom turbine area

jwallwork23:
I noticed that the tidal turbine formulation assumes circular turbine footprints. This PR keeps that as default, but allows a user provided area. (A test case I'm considering uses rectangular turbine footprints.)

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -5,7 +5,7 @@ All options are type-checked and they are stored in traitlets Configurable
 objects.
 """
 from .configuration import *
-from firedrake import Constant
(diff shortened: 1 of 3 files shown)


stephankramer:
Extra option makes sense - the options classes should just be a place to store user configuration though, and not a place to store related code. I will just add this option to the discrete-turbines-cleanup branch I'm preparing for PR. Would need to change it there anyway...

jwallwork23:
Thanks @stephankramer, sounds good. I'll close this PR, in that case.

PR #224, opened 2020-09-18
Fix compatibility with traitlets 5

### LINKED ISSUE #223: New version of traitlets package
The new version of the python package traitlets (5.0.4) does not contain the object ObserveHandler and hence `import thetis` fails because 'ObserveHandler' in "thetis/configuration.py", line 334 is not defined. 

Note: with the older version of traitlets (4.3.3) this error does not occur.

stephankramer:
Change ObserveHandler and DefaultHandler to decorators observe and
default, which seem to do the same.
For some reason we can't overwrite the default_value attribute of
PETScSolverParameters.

Fixes #223

### DIFF
--- a/thetis/configuration.py
+++ b/thetis/configuration.py
@@ -204,7 +204,6 @@ class FiredrakeVectorExpression(TraitType):
 
 class PETScSolverParameters(Dict):
     """PETSc solver options dictionary"""
-    default_value = None
     info_text = 'a PETSc solver options dictionary'
 
     def validate(self, obj, value):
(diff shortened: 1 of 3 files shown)


tkarna:
Looks good, this works on my Ubuntu 18.04 with traitlets 4.3.3.

tkarna:
What's the firedrake issue that adjoint tests are pending on?

stephankramer:
https://github.com/firedrakeproject/firedrake/issues/1841
It's fix in https://github.com/firedrakeproject/firedrake/pull/1842 is held up by failing serendipity tests in firedrake master

stephankramer:
I've also included two changes to deal with the changed solver defaults - just so we can get master green again. We should probably have a think whether we want to add `'mat_type': 'baij'` in most?/all dg solves?

Coming back to why the traitlets issue was only seen on some systems: it turns out indeed that traitlets 5 has a requirement for python>=3.7 - so pip will only give you traitlets 5 if you have that.

There is one remaining warning with traitlets 5 (deprecation warning: `trait=` keyword should be `value_trait=`); Unfortunately fixing that gives a warning with traitlets 4, and I'm not a 100% sure it works correctly (the fix) with traitlets 4. So I left it for now...

tkarna (approved):
Thanks the updates, looks all good. If the remaining traitlet deprecation warning should be an issue we could derive another `List` and `Dictionary` classes from traitlets and implement version checking therein.

PR #228, opened 2020-12-02
Meander example

mc4117:
Adding the example of a meander to the suite of thetis test cases. In order to facilitate this test case, small changes to the sediment model have been made.


examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
+"""
+Meander Test case
+=======================
+
+Solves the test case of a meander.
```
tkarna, line 5: Maybe you could be a bit more specific, "meander" is quite a generic term. Idealised 2D sediment and morphodynamics test in a circular channel or something like that?

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+
+We use this test case to validate the implementation of the mathematical and
+numerical methods used in Thetis to model sediment transport and morphological changes.
+Specifically this test case tests the secondary current implementation.
+
+For more details, see
+[1] Clare et al. 2020. “Hydro-morphodynamics 2D Modelling Using a Discontinuous
+    Galerkin Discretisation.” EarthArXiv. January 9. doi:10.31223/osf.io/tpqvy.
+"""
+
```
tkarna, line 15: too many blank lines in the file

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+vector_dg = VectorFunctionSpace(mesh2d, 'DG', 1)
+
+# define underlying bathymetry
+
+bathymetry_2d = Function(V, name='Bathymetry')
+
+gradient = Constant(0.0035)
+
+L_function = Function(V).interpolate(conditional(x > 5, pi*4*((pi/2)-acos((x-5)/(sqrt((x-5)**2+(y-2.5)**2))))/pi, pi*4*((pi/2)-acos((-x+5)/(sqrt((x-5)**2+(y-2.5)**2))))/pi))
+bathymetry_2d1 = Function(V).interpolate(conditional(y > 2.5, conditional(x < 5, (L_function*gradient) + 9.97072, -(L_function*gradient) + 9.97072), 9.97072))
```
tkarna, line 55: Too long lines, these are hard to read as it requires an editor configured to wrap the lines. Should be easy enough to reformat. It's a good idea to stick to PEP8 79 characters whenever possible.

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+
+init = max(bathymetry_2d1.dat.data[:])
+final = min(bathymetry_2d1.dat.data[:])
+
+bathymetry_2d2 = Function(V).interpolate(conditional(x <= 5, conditional(y <= 2.5, -9.97072 + gradient*abs(y - 2.5) + init, 0), conditional(y <= 2.5, -9.97072 - gradient*abs(y - 2.5) + final, 0)))
```
tkarna, line 60: maybe bathymetry_2d1 bathymetry_2d2 could have better names?

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+bathymetry_2d = Function(V).interpolate(-bathymetry_2d1 - bathymetry_2d2)
+
+# record initial bathymetry before it evolves
+initial_bathymetry_2d = Function(V).interpolate(bathymetry_2d)
+
+# choose directory to output results
+ts = time.time()
+st = datetime.datetime.fromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
+outputdir = 'outputs' + st
```
tkarna, line 69: I think we've discussed before; is it necessary to have the time stamps in the output directory name in the examples? This convention is not used in any other examples.

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+
+# initialise velocity and elevation
+chk = DumbCheckpoint("hydrodynamics_meander/elevation", mode=FILE_READ)
```
tkarna, line 72: Maybe not needed in this PR, but would be good to find a way not to have to read the init condition from the disk, especially when running the test suite.

examples/meander/meander_hydro.py
```diff
@@ -0,0 +1,129 @@
...
+P1_2d = FunctionSpace(mesh2d, 'DG', 1)
+vectorP1_2d = VectorFunctionSpace(mesh2d, 'DG', 1)
+
+# define underlying bathymetry
+
+bathymetry_2d = Function(V, name='Bathymetry')
+
+gradient = Constant(0.0035)
+
+L_function = Function(V).interpolate(conditional(x > 5, pi*4*((pi/2)-acos((x-5)/(sqrt((x-5)**2+(y-2.5)**2))))/pi, pi*4*((pi/2)-acos((-x+5)/(sqrt((x-5)**2+(y-2.5)**2))))/pi))
```
tkarna, line 36: long lines

examples/meander/meander_hydro.py
```diff
@@ -0,0 +1,129 @@
...
+
+bathymetry_2d1 = Function(V).interpolate(conditional(y > 2.5, conditional(x < 5, (L_function*gradient) + 9.97072, -(L_function*gradient) + 9.97072), 9.97072))
```
tkarna, line 38: Are these the same fields as in the main file? Then would be better to define all shared fields/expressions in a separate file like `meander_setup.py` and import here

test/examples/test_examples.py
```diff
@@ -35,6 +35,7 @@
     'lockExchange/plotting.py',
     'lockExchange/submitRuns.py',
     'migrating_trench/trench_hydro.py',
+    'meander/meander_hydro.py',
```
tkarna, line 38: why are the hydro examples omitted from tests? the scripts do have the test switch

thetis/sediment_model.py
```diff
@@ -128,115 +128,88 @@ def __init__(self, options, mesh2d, uv, elev, depth):
...
-        elif max(self.dstar.dat.data[:] < 20):
-            self.thetacr = Constant(0.04*(self.dstar**(-0.1)))
-        elif max(self.dstar.dat.data[:] < 150):
-            self.thetacr = Constant(0.013*(self.dstar**(0.29)))
-        else:
-            self.thetacr = Constant(0.055)
+        self.thetacr = Function(self.R_1d).project(conditional(self.dstar < 4, 0.24*(self.dstar**(-1)),
+                                                               conditional(self.dstar < 10, 0.14*(self.dstar**(-0.64)),
+                                                                           conditional(self.dstar < 20, 0.04*(self.dstar**(-0.1)),
+                                                                                       conditional(self.dstar < 150, 0.013*(self.dstar**(0.29)), 0.055)))))
```
tkarna, line 139: long lines, please reformat

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+elev = Function(DG_2d, name="elevation")
+chk.load(elev)
+chk.close()
+
+chk = DumbCheckpoint('hydrodynamics_meander/velocity', mode=FILE_READ)
```
tkarna, line 77: Would be better to use a `with` clause to open a file, also in the hydro file
```
with DumbCheckpoint('hydrodynamics_meander/velocity', mode=FILE_READ) as chk:
    foo
```

examples/meander/meander_example.py
```diff
@@ -0,0 +1,157 @@
...
+# this test case only uses bedload transport but using all slope effect corrections and secondary current
+options.sediment_model_options.solve_suspended_sediment = False
+options.sediment_model_options.use_bedload = True
+options.sediment_model_options.solve_exner = True
+options.sediment_model_options.use_angle_correction = True
+options.sediment_model_options.use_slope_mag_correction = True
+options.sediment_model_options.use_secondary_current = True
+options.sediment_model_options.use_advective_velocity_correction = False
+options.sediment_model_options.morphological_viscosity = Constant(1e-6)
+
```
tkarna, line 107: skipping the blank lines would make the file shorter

examples/meander/meander_hydro.py
```diff
@@ -0,0 +1,129 @@
...
+# simulate initial hydrodynamics
+# define initial elevation
+elev_init = Function(P1_2d).interpolate(0.0544 - bathymetry_2d)
+# define initial velocity
+uv_init = Function(vectorP1_2d).interpolate(as_vector((0.001, 0.001)))
+
+# choose directory to output results
+ts = time.time()
+st = datetime.datetime.fromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
+outputdir = 'outputs' + st
```
tkarna, line 55: same comment about the date stamp in the output dir

thetis/exner_eq.py
```diff
@@ -111,8 +111,18 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
...
+
         for bnd_marker in (bnd_conditions or []):
-            f += -self.test*(fac*qbx*self.n[0] + fac*qby*self.n[1])*self.ds(bnd_marker)
+            bc = bnd_conditions[bnd_marker]
+
+            keys = [*bc.keys()]
+            values = [*bc.values()]
+
+            for i in range(len(values)):
+                if float(values[i]) != 0.0:
```
tkarna, line 122: Hmm, this looks like a strange condition. Why 0 value is special? What if it is 1e-16? Maybe the "null" case could be handled with None or something?

thetis/sediment_model.py
```diff
@@ -128,115 +128,88 @@ def __init__(self, options, mesh2d, uv, elev, depth):
...
-            self.settling_velocity = Constant(1.1*sqrt(self.g*self.average_size*self.R))
-
-        # dictionary of steps (interpolate/project) to perform in order in update()
-        self.update_steps = OrderedDict()
-        # fields updated in these steps
-        self.fields = AttrDict()
+        self.settling_velocity = Function(self.R_1d).assign(conditional(self.average_size <= 1e-04, self.g*(self.average_size**2)*self.R/(18*self.viscosity),
+                                                                        conditional(self.average_size <= 1e-03, (10*self.viscosity/self.average_size)
+                                                                                    * (sqrt(1 + 0.01*((self.R*self.g*(self.average_size**3))
+                                                                                       / (self.viscosity**2)))-1), 1.1*sqrt(self.g*self.average_size*self.R))))
```
tkarna, line 148: long lines

tkarna (changes requested):
Thanks for this, looks generally fine, I just have minor comments.

The test case is called `examples/meander` which is not very descriptive. Could be `morpho_meander_2d` or something like that to give the user a better idea what it is about. (`migrating_trench` could also be renamed `morpho_trench_2d`?)

You should probably include the mesh geo file as well for completeness.

examples/meander/meander_example.py
mc4117: Yes sorry! I've removed the timestamps

examples/meander/meander_hydro.py
mc4117: removed timestamp

test/examples/test_examples.py
mc4117: The hydro files are run via the main scripts e.g. line 65 in migrating_trench/trench_example.py

examples/meander/meander_example.py
mc4117: changed to bathymetry_curve and bathymetry_straight

examples/meander/meander_example.py
mc4117: have reformatted the lines

examples/meander/meander_hydro.py
mc4117: have reformatted the lines

examples/meander/meander_example.py
mc4117: I have added the with clause to both files

thetis/exner_eq.py
mc4117: I specifically made this 0 because the only time I don't want the boundary to be considered is if the user has specified a closed boundary. If it's 1e-16 then there is some material that is leaving the system. I agree handling it with None might be better but it seemed like a lot of unnecessary extra code and boundary options for something that could be handled more simply.

examples/meander/meander_example.py
mc4117: removed most of the blank lines

thetis/sediment_model.py
mc4117: have reformatted

examples/meander/meander_example.py
mc4117: have removed most blank lines

thetis/sediment_model.py
mc4117: have reformatted

examples/meander/meander_example.py
mc4117: I have replaced the text with "Solves the test case of flow around a 180 degree bend replicating lab experiment 4 in Yen & Lee (1995)."

examples/meander/meander_hydro.py
mc4117: I have added a meander_setup file

test/examples/test_examples.py
tkarna: ok, I see

examples/sediment_trench_2d/trench_hydro.py
```diff
@@ -8,9 +8,8 @@
 the morphodynamic simulation.
 
 For more details of the test case set-up see
-[1] Clare et al. 2020. “Hydro-morphodynamics 2D Modelling Using a Discontinuous
-    Galerkin Discretisation.” EarthArXiv. January 9. doi:10.31223/osf.io/tpqvy.
-
+[1] Clare et al. (2020). Hydro-morphodynamics 2D modelling using a discontinuous Galerkin discretisation.
+    Computers & Geosciences, 104658.
```
tkarna, line 12: maybe add a doi for the impatient?

examples/sediment_meander_2d/meander_hydro.py
```diff
@@ -0,0 +1,87 @@
...
+options.timestepper_options.implicitness_theta = 1.0
+if not hasattr(options.timestepper_options, 'use_automatic_timestep'):
+    options.timestep = 1
+
+# set boundary conditions
+elev_init_const = (-max(bathymetry_2d.dat.data[:]) + 0.05436)
+left_bnd_id = 1
+right_bnd_id = 2
+swe_bnd = {}
+swe_bnd[3] = {'un': Constant(0.0)}
```
tkarna, line 68: also here the bnd condition could be omitted?

thetis/exner_eq.py
```diff
@@ -111,8 +111,17 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions=No
...
-            f += -self.test*(fac*qbx*self.n[0] + fac*qby*self.n[1])*self.ds(bnd_marker)
+            no_contr = False
+            keys = [*bnd_conditions[bnd_marker].keys()]
+            values = [*bnd_conditions[bnd_marker].values()]
+            for i in range(len(keys)):
+                if keys[i] != 'elev' and float(values[i]) == 0.0:
+                    no_contr = True
+
+            if not no_contr:
+                f += -self.test*(fac*qbx*self.n[0] + fac*qby*self.n[1])*self.ds(bnd_marker)
```
tkarna, line 124: So if I understand this correctly, for closed boundary you want to omit the boundary term on line 124? So currently the code does this (un applies for uv and flux conditions as well):
```
1 unprescribed : open
2 elev : open
3 un != 0 : open
4 un == 0 : closed
5 elev and un != 0 : open
6 elev and un == 0 : closed
```
I'm not sure if this is right. To be consistent with SWE in Thetis, case 1 should be closed. Is it necessary to differentiate the closed condition with un==0? Can you not just do
```
1 unprescribed : closed
2 elev : open
3 un : open
4 elev and un : open
```

examples/sediment_meander_2d/meander_example.py
```diff
@@ -0,0 +1,116 @@
...
+gradient_flux = (-0.053 + 0.02)/6000
+gradient_flux2 = (-0.02+0.053)/(18000-6000)
+gradient_elev = (10.04414-9.9955)/6000
+gradient_elev2 = (9.9955-10.04414)/(18000-6000)
+elev_init_const = (-max(bathymetry_2d.dat.data[:]) + 0.05436)
+flux_constant = Constant(-0.02)
+elev_constant = Constant(elev_init_const)
+
+swe_bnd = {}
+swe_bnd[3] = {'un': Constant(0.0)}
```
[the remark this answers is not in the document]
mc4117: There is a slight difference in the formulation in the equations if you enforce a zero boundary condition rather than not stating anything. This test case has issues at the boundaries (see section 5.2.1 of the paper https://www.sciencedirect.com/science/article/pii/S0098300420306324?casa_token=yW9-p0xX-WIAAAAA:pb6fwk5QcNZf6KXF0q6Ol451HIN6dFo-0kLQUrzB8z0tdhuBcfUz6lmHrPWaVJM5aqARHH-LXTA3#sec5) and I found that enforcing the strong boundary condition helped with this issue a bit.

examples/sediment_meander_2d/meander_hydro.py
mc4117: The implementation of this boundary condition has more of an impact when resolving the coupled hydro-morphodynamic model so you're right this could definitely be omitted. However, seeing as it is going to be used in meander_example.py I thought that for consistency between the two scripts it would be better to include it here.

thetis/exner_eq.py
mc4117: I'm not sure I entirely understand but this is what I think is happening in these lines of code:
1 unprescribed - closed - no exner bc
2 un = 0 - closed - no exner bc (irrespective of other conditions)
3 un != 0 - open - exner bc
4 elev and no un/un !=0 - open - exner bc
It is necessary to create a catch for cases 1 and 2 because there are advantages to using case 2 in the other equations. You are right that because of this "if clause" from the perspective of the exner equation cases 1 and 2 are the same

tkarna: Hmm, so line 124 means that you evaluate the boundary flux with the vector (qbx, qby). Thus this is an "open" boundary. If the boundary term is omitted, it is identically zero, and therefore we have a "closed" boundary.
So the logic tree is
```
1 un prescribed and un == 0  -> no_contr = True -> omit bnd term -> closed
2 all other cases: no_contr = False -> include bnd term -> open
```
Do you agree with this?

mc4117: Yes I agree with this, apart from case 1 hold for both un and flux (ie. if flux exists and flux==0 then closed)

thetis/exner_eq.py
tkarna: I think we can revisit the boundary conditions once bendy elements work

examples/sediment_meander_2d/meander_example.py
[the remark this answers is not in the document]
tkarna: I think we can revisit the boundary conditions once bendy elements work

thetis/exner_eq.py
tkarna: Also here let's redo the boundaries once bendy elements work

examples/sediment_meander_2d/meander.geo
```diff
@@ -0,0 +1,46 @@
...
+Physical Line(2) = {38, 39};
+Physical Line(3) = {3, 4, 5, 6, 7, 8 ,9, 10};
+
+Line Loop (20) = {35, 5, 9, 3, -8, 39, -37, 11, 12, -34};
+Plane Surface(16) = {20};
+Physical Surface(200) = {20};
+
+Line Loop (30) = {-36, 6, 10, 4, -7, 38, -37, 11, 12, -34};
+Plane Surface(20) = {30};
+Physical Surface(100) = {16};
```
tkarna, line 46: I noticed that this geo file does not generate the mesh :) The element size defs are missing.

thetis/sediment_model.py
```diff
@@ -106,15 +105,16 @@ def __init__(self, options, mesh2d, uv, elev, depth):
         # define function spaces
         self.P1DG_2d = get_functionspace(mesh2d, "DG", 1)
         self.P1_2d = get_functionspace(mesh2d, "CG", 1)
+        self.R_1d = get_functionspace(mesh2d, "R", 0)
```
tkarna, line 108: I think R_1d is now unused. And you need to `make lint` ;)

tkarna (approved):
Thanks for the updates, this looks good to me

PR #229, opened 2020-12-04
Shallow water on the sphere

tkarna:
Generalizes the 2D shallow water model to spherical meshes. Adds convergence tests for Williamson 2 and Läuter 3 test cases. Williamson 5 test is included without a correctness check.

To run the 2D model on the sphere, create a mesh on spherical surface, and call `init_cell_orientations`:
```python
mesh2d = IcosahedralSphereMesh(radius=r_earth, refinement_level=refinement, degree=3)
mesh2d.init_cell_orientations(SpatialCoordinate(mesh2d))
```
Note that in spherical geometry the curvature of the elements affects the model accuracy; For coarse grids, it's recommended to use coordinates with degree 2 or larger.

Spherical geometry only works with HDiv velocity spaces, i.e. the `'rt-dg'` and `'bdm-dg'` element families.

### DIFF
--- a/test/sphere/test_williamson.py
+++ b/test/sphere/test_williamson.py
@@ -1,0 +1,321 @@
+"""
+Shallow water test cases on the sphere by Willamson et al. (1992) and
+Läuter et al (2005).
+
+[1] Williamson et al., 1992. A standard test set for numerical approximations
+    to the shallow water equations in spherical geometry. Journal of
+    Computational Physics, (1):211–224.
+    https://doi.org/10.1016/S0021-9991(05)80016-6
+[2] Läuter et al., 2005. Unsteady analytical solutions of the spherical shallow
+    water equations. Journal of Computational Physics, (2):535–553.
+    https://doi.org/10.1016/j.jcp.2005.04.022
+"""
+from thetis import *
+from scipy import stats
+import numpy
+import pytest
+
+r_earth = 6371220.  # radius of Earth
+omega = 7.292e-5  # Earth's angular velocity
+
+
+def coords_xyz_to_lonlat(mesh):
+    """
+    Convert Earth-centered Cartesian coordinates to (longitude, latitude)
+    """
+    x, y, z = SpatialCoordinate(mesh)
+    z_norm = z / sqrt(x**2 + y**2 + z**2)
+    z_norm = Min(Max(z_norm, -1.0), 1.0)  # avoid silly roundoff errors
+    lat = asin(z_norm)
+    lon = atan_2(y, x)
+    return lon, lat
+
+
+def vector_enu_to_xyz(mesh, uvw_enu_expr):
+    """
+    Convert vector from local tanget plane to Earth-centered Cartesian system.
+
+    :arg x, y, z: spatial coordinates of the mesh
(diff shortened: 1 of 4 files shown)


thetis/shallowwater_eq.py
```diff
@@ -601,7 +591,13 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
         coriolis = fields_old.get('coriolis')
         f = 0
         if coriolis is not None:
-            f += coriolis*(-uv[1]*self.u_test[0] + uv[0]*self.u_test[1])*self.dx
+            if self.on_the_sphere:
+                outward_normals = CellNormal(self.mesh)
+                u = cross(outward_normals, uv)
+                f = coriolis*inner(self.u_test, u)*dx
```
stephankramer, line 597: `self.dx`

test/sphere/test_williamson.py
```diff
@@ -0,0 +1,321 @@
...
+    epsilon = Constant(1e-3)
+    r_h = sqrt(x**2 + y**2 + epsilon)
+    # local tangent plane coordinate system unit vectors
+    ne = as_vector((-y, x, 0)) * 1 / r_h  # east
+    nn = as_vector((-x * z, -y * z, x**2 + y**2)) * 1 / r_h / r_earth  # north
+    nu = as_vector((x, y, z)) / r_earth  # up
+    # map vectors from local ENU coordinates to ECEF
+    M = as_tensor((ne, nn, nu)).T
+    uvw_expr = M * uvw_enu_expr
+    return uvw_expr
```
stephankramer, line 52: These might be good to have as generic utility, if we get more tests like this (not nec. in this PR) - and have them sepaately tested, as in my experience they always end up having bugs

stephankramer (dismissed):
Looks good! I actually prefer `div`s and `grad`s over explicit component expressions with `Dx`

thetis/shallowwater_eq.py
tkarna: Good catch, thanks!

test/sphere/test_williamson.py
tkarna: Yeah, I totally agree. Users are probably going to need these in their applications as well. Maybe we can design the API when we have a couple of use cases.

thetis/shallowwater_eq.py
```diff
@@ -601,7 +591,13 @@ def residual(self, uv, eta, uv_old, eta_old, fields, fields_old, bnd_conditions=
         coriolis = fields_old.get('coriolis')
         f = 0
         if coriolis is not None:
-            f += coriolis*(-uv[1]*self.u_test[0] + uv[0]*self.u_test[1])*self.dx
+            if self.on_the_sphere:
+                outward_normals = CellNormal(self.mesh)
+                u = cross(outward_normals, uv)
+                f = coriolis*inner(self.u_test, u)*self.dx
+            else:
+                ez_x_uv = -uv[1]*self.u_test[0] + uv[0]*self.u_test[1]
```
tkarna, line 599: I guess there's no short `cross` expression for this one?

stephankramer (approved):

Issue #188, opened 2020-02-28
Negative tracer values


jwallwork23:
Apparently a few people are looking for a steady state passive tracer example in Thetis. So I thought I would code up the TELEMAC-2D 'point discharge with diffusion' test case which I considered using a standalone SUPG P1 solver in this paper: https://zenodo.org/record/3653373#.Xlkdaumnw5k

I made the new branch `passive-tracer-example`, where there is a coupled `'SteadyState'` solver. If I run `examples/passiveTracer/passiveTracer.py`, it gives a very encouraging plot:

![negative_tracer](https://user-images.githubusercontent.com/22053413/75554574-84bc0280-5a32-11ea-9413-25bef7f6acff.png)

Visually, it agrees nicely with the asymptotic solution and SUPG approximation as plotted in the paper (Figure 1, p.92). However, the values are significantly negative! Does anyone have an idea what is going wrong?

tkarna:
Hmm, no idea. Is the (elevation, velocity) solution correct? One way to debug is to run the time dependent solution and make sure it works correctly.

tkarna:
@jwallwork23 is this issue still active/relevant?

jwallwork23:
Thanks for the reminder @tkarna! I just merged master into the branch and the negative tracer values have miraculously gone away (to order 1e-07).

![fixed](https://user-images.githubusercontent.com/22053413/105686851-a4cc9e00-5eef-11eb-8e63-2a4022106e6b.png)

I'll clean up the branch and make a PR to enable the coupled 2D steady state solver (which currently gets a `NotImplementedError`) because it isn't tested.

PR #230, opened 2021-01-25
Point discharge with diffusion example

jwallwork23:
Following the closing of issue #188, this PR acts to check that the 2D steady state tracer solver works.

Convergence of the error due to evaluating two linear functionals is assessed under uniform mesh refinement. Convergence rates for the case `hydrodynamics=False` below.

convergence_aligned
(Aligned source and receiver)

convergence_offset
(Offset source and receiver)

This could be a test, if that would be useful / there aren't too many similar tests.

### DIFF
--- a/examples/point_discharge/point_discharge.py
+++ b/examples/point_discharge/point_discharge.py
@@ -1,6 +1,6 @@
 """
-TELEMAC-2D point discharge with diffusion test case
-===================================================
+TELEMAC-2D `Point Discharge with Diffusion' test case
+=====================================================
 
 Solves tracer advection equation in a rectangular domain with
 uniform fluid velocity, constant diffusivity and a constant
@@ -66,22 +66,17 @@ def bessk0(x):
     return conditional(ge(x, 2), expr2, expr1)
 
 
-class PointDischargeParameters():
-    def __init__(self, setup=1):
-        assert setup in (1, 2)
-        self.setup = setup
+class PointDischargeParameters(object):
+    """
+    Problem parameter class, including point source representation.
 
-        # Parametrisation of point source
-        # NOTE: Delta functions are not in H1 and hence do not live in the FunctionSpaces we seek
-        #       to use. The idea here is to approximate the delta function using a disc with a very
-        #       small radius. This works well for practical applications, but is not quite right for
-        #       analytical test cases such as this. As such, we have calibrated the disc radius so
-        #       that solving on a sequence of increasingly refined uniform meshes leads to
-        #       convergence of the uniform mesh solution to the analytical solution.
-        self.source_x, self.source_y = 2.0, 5.0
-        self.source_r = 0.07980 if setup == 1 else 0.07972
-        # TODO: Use Gaussian representation from [3]
-        self.source_value = 100.0
+    Delta functions are not in H1 and hence do not live in the FunctionSpaces we
+    seek to use. Here we use a Gaussian approximation with a small radius. The
+    small radius has been calibrated against the analytical solution. See [3]
+    for details.
+    """
+    def __init__(self, offset):
+        self.offset = offset
 
         # Physical parameters
         self.diffusivity = Constant(0.1)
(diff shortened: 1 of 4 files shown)


examples/point_discharge/point_discharge.py
```diff
@@ -0,0 +1,261 @@
...
+        self.source_r = 0.05606298 if self.use_lax_friedrichs_tracer else 0.05606298
+        self.source_value = 100.0
+
+        # Boundary conditions
+        self.boundary_conditions = {
+            'tracer': {
+                1: {'value': Constant(0.0)},      # inflow
+                2: {'open': None},                # outflow
+                3: {'diff_flux': Constant(0.0)},  # Neumann
+                4: {'diff_flux': Constant(0.0)},  # Neumann
```
tkarna, line 98: I think this is equivalent to omitting the boundary condition declaration (which implies that the boundary flux term is omitted).

examples/point_discharge/point_discharge.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def quantity_of_interest(self, sol):
+        kernel = self.quantity_of_interest_kernel(sol.function_space().mesh())
+        return assemble(inner(kernel, sol)*dx(degree=12))
+
+    def analytical_quantity_of_interest(self, mesh):
+        """
+        The analytical solution can be found in [1]. Due to the modified
+        Bessel function, it cannot be evaluated exactly and instead must
+        be computed using quadrature.
```
tkarna, line 146: a quadrature

examples/point_discharge/point_discharge.py
```diff
@@ -0,0 +1,261 @@
...
+    expr2 = exp(-x)/sqrt(x)*(1.25331414 + y2*(-0.7832358e-1 + y2*(0.2189568e-1 + y2*(
+        -0.1062446e-1 + y2*(0.587872e-2 + y2*(-0.251540e-2 + y2*0.53208e-3))))))
+    return conditional(ge(x, 2), expr2, expr1)
+
+
+class PointDischargeParameters(object):
+    """
+    Problem parameter class, including point source representation.
+
+    Delta functions are not in H1 and hence do not live in the FunctionSpaces we
```
tkarna, line 69: Could mention that the analytical solution involves a point source (dirac delta). And the analytical solution gives the tracer value at an arbitrary "receiver" location (if I got that right). Also the offset parameter is not explained, it moves the receiver y location to be on off-axis?

examples/point_discharge/point_discharge.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def source(self, fs):
+        triple = (self.source_x, self.source_y, self.source_r)
+        return self.gaussian(fs.mesh(), triple, scaling=self.source_value)
+
+    def bathymetry(self, fs):
+        return Function(fs).assign(5.0)
+
+    def quantity_of_interest_kernel(self, mesh):
+        triple = (20.0, 7.5 if self.offset else 5.0, 0.5)
```
tkarna, line 132: minor detail: took me a while to figure out what triple is, you could define a x,y,radius dict and pass **dict to ball.

examples/point_discharge/point_discharge.py
```diff
@@ -0,0 +1,261 @@
+"""
+TELEMAC-2D `Point Discharge with Diffusion' test case
+=====================================================
+
+Solves tracer advection equation in a rectangular domain with
+uniform fluid velocity, constant diffusivity and a constant
+tracer source term. Neumann conditions are imposed on the
+channel walls, an inflow condition is imposed on the
+left-hand boundary, and the right-hand boundary remains open.
```
tkarna, line 9: maybe mention here that this is a steady-state problem for which an analytical solution exists

tkarna (changes requested):
Thanks for this! Looks good. As the script runs a convergence test this is more a test rather than an example? So would better fit under `test/tracerEq/`(?) rather than `examples/`. Otherwise only minor comments.

jwallwork23:
Thanks for the review, @tkarna. I think I addressed all your comments.

tkarna (approved):
Thanks, looks good!

tkarna (approved):

PR #231, opened 2021-02-08
Revised interior penalty coefficient

tkarna:
Implement a simpler form for symmetric interior penalty coefficient following Koen Hillewaert's [PhD thesis](https://dial.uclouvain.be/pr/boreal/object/boreal:128254/datastream/PDF_01/view).

In this method the "element size" length scale is defined as `CellVolume(mesh)/FacetArea(mesh)` making it robust for skewed elements as well. Compared to the theoretical bound, the IP penalty term has been multiplied by 2 as it seems to stable in practice.

This PR:
- Removes the now unnecessary `use_automatic_sipg_parameter` option
- Replaces the `sipg_penalty` options by `sipg_factor` which just scales the IP penalty term (defaults to 1.0)

### DIFF
--- a/test/swe2d/test_anisotropic.py
+++ b/test/swe2d/test_anisotropic.py
@@ -11,10 +11,6 @@ of [1]. The main contributions of [1] are the derivation of a goal-oriented erro
 shallow water modelling and subsequent implementation of mesh adaptation algorithms. An adapted mesh
 resulting from this process is used in this test. The mesh is anisotropic in the flow direction.
 
-If the default SIPG parameter is used, this steady state problem fails to converge under some
-finite element pairs. The automatic SIPG parameter functionality provided by Thetis should ensure
-converge where it was not attained and reduce the number of nonlinear solver iterations otherwise.
-
 [1] J.G. Wallwork, N. Barral, S.C. Kramer, D.A. Ham, M.D. Piggott, "Goal-Oriented Error Estimation
     and Mesh Adaptation in Shallow Water Modelling" (2020), Springer Nature Applied Sciences (to
     appear).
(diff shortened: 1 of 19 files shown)


tkarna:
In some tests, I had to increase the tolerance of convergence rate. But the actual error is generally lower (especially with coarser meshes).

test/swe2d/test_anisotropic.py
```diff
@@ -137,21 +133,15 @@ def family(request):
...
-        assert snes_it_auto < snes_it_default, msg.format(snes_it_auto, snes_it_default)
-        msg = "auto-sipg: snes iterations reduced from {:d} to {:d} for {:s} PASSED"
-        print_output(msg.format(snes_it_default, snes_it_auto, family))
-    except ConvergenceError:
-        msg = "auto-sipg: snes converged where it diverged under default for {:s} PASSED"
-        print_output(msg.format(family))
+    snes_it = run(**options)
+    expected = 3
+    msg = f'snes iterations exceed expected: {snes_it} > {expected}'
+    assert snes_it <= expected, msg
```
jwallwork23, line 139: If I remember correctly, the elements in this mesh are only moderately anisotropic. It would be good to extend the test for some extreme cases (such as you get in r-adaptation), but I can do this in a separate PR.

test/tracerEq/test_v-diffusion_mes.py
```diff
@@ -177,36 +167,35 @@ def check_convergence(x_log, y_log, expected_slope, field_str, saveplot):
...
-                    use_ale_moving_mesh=use_ale)
+@pytest.mark.parametrize(
+    ('stepper', 'polynomial_degree', 'implicit', 'expected_rate'),
+    [
+        ('SSPRK22', 0, False, 1.0),
+        ('SSPRK22', 1, False, 1.7),
+        ('SSPRK22', 0, True, 1.0),
+        ('SSPRK22', 1, True, 2.2),
+    ]
+)
```
jwallwork23, line 193: Out of interest, how come you aren't testing leap frog any more?

jwallwork23 (approved):
Looks good to me! Presumably this cuts down the number of tests, since there is no longer auto-sipg.

test/swe2d/test_anisotropic.py
tkarna: Yeah, sounds like a good idea!

test/tracerEq/test_v-diffusion_mes.py
tkarna: Primarily to cut down the testing. Also because leap-frog time integrator is not really being used and I think we can remove it soon.

thetis/momentum_eq.py
```diff
@@ -358,18 +362,19 @@ def grad_h(a):
...
-            assert alpha is not None
-            sigma = avg(alpha/elemsize)
+            h_cell = self.mesh.ufl_cell().sub_cells()[0]
+            p, q = self.function_space.ufl_element().degree()
+            cp = (p + 1) * (p + 2) / 2 if h_cell == triangle else (p + 1)**2
+            # by default the factor is multiplied by 2 to ensure convergence
+            sigma = cp * FacetArea(self.mesh) / CellVolume(self.mesh)
+            sp = sigma('+')
+            sm = sigma('-')
+            sigma_max = sipg_factor * conditional(sp > sm, sp, sm)
```
stephankramer, line 372: Not that I mind but there's a `Max` as well in ufl: `Max(sp, sm)`

stephankramer (approved):
All looks good - and a lot simpler, nice

jwallwork23 (approved):

PR #233, opened 2021-03-02
Reduce test suite memory consumption

tkarna:
3D examples are run as a part of the test suite to check for syntax errors. They tend to consume quite a lot of memory which has recently caused out-of-memory errors on Jenkins. This PR reduces the max memory consumption from 6GB to 1GB. Wallclock time is also reduced by a factor ~4 by restricting the number of time steps.


jwallwork23 (approved):
I guess for a lot of these tests we are just checking that the simulation actually runs, right? If the number of timesteps being tested is cut down by a factor of 5, are we still testing enough to see any possible issues that could arise? Hard to say! Otherwise, looks fine to me.

tkarna:
> If the number of timesteps being tested is cut down by a factor of 5, are we still testing enough to see any possible issues that could arise?

Yeah, we are essentially testing for syntax errors in case options change etc. If we produce one export it also checks that diagnostics run without fail. To ensure that the simulation result is somehow meaningful we should check the outputs. But then we'd need to run much longer (and figure out meaningful fitness metrics).

More broadly, we should to run a suite of realistic simulations where we check for correctness. But that's too heavy for Jenkins, so we'd need another system.

PR #234, opened 2021-03-03
Fixes in two swe2d tests

jwallwork23:
Couple of things:
* `test_rossby_wave` wasn't always properly assessing convergence, because the error metrics can approach 1 from above or below.
* `test_anisotropic` had turbines which didn't align with the underlying adapted mesh.

The BDM-DG pair is also added to the finite element pair list on the 2D solver doc page.


tkarna (approved):
Thanks for the updates, everything looks good to me!

PR #236, opened 2021-03-03
P1+SUPG for 2D tracer model

jwallwork23:
I've been using P1+SUPG for various things in my PhD work, with the SUPG parameter advocated in [Micheletti et al. 2003], which accounts for element anisotropy. I thought maybe this would be useful for other people using Thetis?

This PR allows for CG tracers in 2D. The default is still DG. If SUPG is used then the test function of the tracer equation is modified, including the modified cell size.

In doing this, I thought it made sense to pass `options` to the tracer solver (in the same way as is done for the SW solver), instead of passing individual parameters. Is that okay? Also, I needed to make the tracer equation aware of the SW velocity field.

I didn't want to add too much to the test suite, so am just testing steady state and for BC implementation. Note that the CG tracer equation inherits the same weak BC implementation from the DG one. I also noticed there were some old SIPG parameter tests we can now drop.

[Micheletti et al. 2003] Micheletti, Stefano, Simona Perotto, and Marco Picasso. "Stabilized finite elements on anisotropic meshes: a priori error estimates for the advection-diffusion and the Stokes problems." SIAM Journal on Numerical Analysis 41.3 (2003): 1131-1162.

### DIFF
--- a/test/tracerEq/test_bcs_2d.py
+++ b/test/tracerEq/test_bcs_2d.py
@@ -120,9 +120,10 @@ def run(refinement, **model_options):
     options.horizontal_diffusivity = nu
     options.horizontal_diffusivity_scale = nu
     options.horizontal_velocity_scale = Constant(0.0)
-    options.use_limiter_for_tracers = True
     options.fields_to_export = ['tracer_2d']
     options.update(model_options)
+    options.use_limiter_for_tracers = options.tracer_element_family == 'dg'
+    options.use_supg_tracer = options.tracer_element_family == 'cg'
     options.simulation_end_time = t_end - 0.5*dt
 
     # Boundary conditions
--- a/test/tracerEq/test_bcs_2d.py
+++ b/test/tracerEq/test_bcs_2d.py
@@ -52,7 +52,9 @@ def fourier_series_solution(mesh, lx, diff_flux, **model_options):
         return assemble(2/lx*source*phi(n)*dx)
 
     def source_term(n):
-        """Simple quadrature is used to approximate time integral."""
+        """
+        A simple quadrature routine is used to approximate the time integral.
+        """
         I = 0
         tau = 0
         dt = 0.05
@@ -116,6 +118,8 @@ def run(refinement, **model_options):
     options.solve_tracer = True
     options.tracer_only = True
     options.horizontal_diffusivity = nu
+    options.horizontal_diffusivity_scale = nu
+    options.horizontal_velocity_scale = Constant(0.0)
     options.use_limiter_for_tracers = True
     options.fields_to_export = ['tracer_2d']
     options.update(model_options)
@@ -149,6 +153,11 @@ def run_convergence(**model_options):
     assert slope > 2, msg.format(slope)
 
 
+@pytest.fixture(params=['dg', 'cg'])
+def family(request):
+    return request.param
+
+
 @pytest.fixture(params=[1])
 def polynomial_degree(request):
     return request.param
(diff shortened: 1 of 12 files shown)


thetis/utility.py
```diff
@@ -1271,54 +1271,85 @@ def get_cell_widths_2d(mesh2d):
...
     return cell_widths
 
 
-def get_sipg_ratio(nu):
+def anisotropic_cell_size(mesh):
     """
-    Compute the ratio between the maximum of `nu` and the minimum of `nu` in each element. If `nu`
-    is P0 or a `Constant` then the resulting ratio is unity in each element. If `nu` varies linearly
-    in each element then the ratios are outputted as a P0 field.
+    Measure of cell size for anisotropic meshes, as described in [Micheletti, Perotto & Picasso 2003].
```
tkarna, line 1279: A full reference? (and keep 1st docstring line short, followed with blank line).

thetis/utility.py
```diff
@@ -1271,54 +1271,85 @@ def get_cell_widths_2d(mesh2d):
...
+    EVals(0) = D(1);
+    EVals(1) = D(0);
+  }
+}
+"""
+    kernel = op2.Kernel(kernel_str, 'reordered_eigendecomposition', cpp=True, include_dirs=include_dir)
+    op2.par_loop(kernel, P0_ten.node_set, evectors.dat(op2.RW), evalues.dat(op2.RW), B.dat(op2.READ))
+
+    # Return minimum eigenvalue
+    return interpolate(evalues[-1], FunctionSpace(mesh, "DG", 0))
```
tkarna, line 1352: Nice!

test/tracerEq/test_bcs_2d.py
```diff
@@ -116,9 +118,12 @@ def run(refinement, **model_options):
     options.solve_tracer = True
     options.tracer_only = True
     options.horizontal_diffusivity = nu
-    options.use_limiter_for_tracers = True
+    options.horizontal_diffusivity_scale = nu
+    options.horizontal_velocity_scale = Constant(0.0)
```
tkarna, line 122: Just curious: is there a reason to set this to 0?

tkarna (approved):
Thanks for this! Looks good to me, only minor comments.

Good to see that generalizing tracers to CG and SUPG is actually quite quite straightforward, not intrusive at all.

thetis/tracer_eq_2d.py
```diff
@@ -262,18 +271,33 @@ class TracerEquation2D(Equation):
     """
     2D tracer advection-diffusion equation :eq:`tracer_eq` in conservative form
     """
-    def __init__(self, function_space, depth,
-                 use_lax_friedrichs=False,
-                 sipg_factor=Constant(1.0)):
+    def __init__(self, function_space, depth, options, velocity):
```
tkarna, line 274: Ah, I see, you need to have access to the velocity field here. I guess we are now passing it twice, in the eq constructor (here) and then again in the time integrator (via fields). Potential pitfall is that user might define these inconsistently. But I do not see immediately see a better solution.

test/tracerEq/test_bcs_2d.py
jwallwork23: Well just because I figured that was a characteristic velocity for that test case, which is just a diffusion problem. Setting it to zero will switch SUPG off.

jwallwork23:
Thanks! Will modify as recommended and have also added P1+SUPG to the new tracer manual page.

thetis/tracer_eq_2d.py
jwallwork23: Good point. So long as the user doesn't tamper with the equation and timestepper then I guess they should stay the same, though.

stephankramer: Yeah I agree, this doesn't feel quite right - but I can't see a proper solution that isn't much more intrusive. This would be much more of an issue if we would do SUPG for momentum advection....

stephankramer (approved):
Very nice, all looks good to me.

PR #237, opened 2021-03-08
Update 2D model manual pages

jwallwork23:
* Fix doc build
* Add DIRK22 to available time integrators
* Add a note on wetting and drying parameters
* Add a page on the 2D tracer model
* Add a page on the 2D sediment model

@mc4117: am I right in saying that the Exner equation should be solved in P1 space? It looks to me like it is solved in whatever function space the user defines the bathymetry in, but maybe I misread.

### DIFF
--- a/docs/source/sediment_formulation_2d.rst
+++ b/docs/source/sediment_formulation_2d.rst
@@ -41,6 +41,14 @@ Spatial discretization
 ----------------------
 
 Thetis currently only supports suspended sediment in P1DG space.
+Lax-Friedrichs stabilization is used by default and may be
+controlled using the
+:ref:`ModelOptions2d<model_options_2d>`.\ :py:attr:`.use_lax_friedrichs_tracer`
+option. The scaling parameter used by this scheme may be controlled
+using the
+:ref:`ModelOptions2d<model_options_2d>`.\ :py:attr:`.lax_friedrichs_tracer_scaling_factor`
+option.
+
 The function space used for the bedlevel is determined by that
 used for the bathymetry. Typically, this is P1.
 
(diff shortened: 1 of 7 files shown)


docs/source/sediment_formulation_2d.rst
```diff
@@ -0,0 +1,66 @@
...
+:math:`H` is the total water depth.
+The conservative tracer model is specified using the
+:ref:`ModelOptions2d<model_options_2d>`.\ :py:attr:`.use_sediment_conservative_form`
+option.
+
+Bedload transport is modelled in two dimensions using the
+Exner equation
+:eq:`exner_eq`.
+It is solved for the bedlevel,
+:math:`z_b`, which has the effect of modifying the bathymetry.
```
mc4117, line 29: I would say ":math:`z_b` which is the bathymetry" because if you look at the code z_b is the bathymetry

mc4117:
@jwallwork23 I have only ever tried solving the exner equation in P1 but I don't see any reason in either the physics or the mathematics why it couldn't be P2 etc. In theory in order to avoid grid scale noise and unstable oscillations using CG is better than DG but in reality I think it makes very little difference. I like the way you have it in the documentation currently - "typically this is P1"

jwallwork23:
Okay cool, I'll leave it how it is. I was just wondering because if you input the bathymetry as DG then you won't have any flux terms, so you might not get what you expect?

mc4117:
I'm pretty sure I went through this once and checked that it was ok but perhaps it would be best to say it should be cg. I will add an if clause to the code as well

tkarna (approved):
Thanks a lot for these changes, looks good to me!

PR #238, opened 2021-03-19
2D tracer demo

jwallwork23:
I created this demo by translating the Firedrake [DG advection](https://firedrakeproject.org/demos/DG_advection.py.html) demo into Thetis.

There were a few tweaks required to import the bibliography and also I made sure `'DQ'` is specified on quads, to avoid getting lots of `UFL:WARNING Discontinuous Lagrange element requested on quadrilateral, creating DQ element.` warnings.

### DIFF
--- a/demos/demo_2d_tracer.py
+++ b/demos/demo_2d_tracer.py
@@ -73,7 +73,7 @@ vP1_2d = VectorFunctionSpace(mesh2d, "CG", 1)
 x, y = SpatialCoordinate(mesh2d)
 uv_init = interpolate(as_vector([0.5 - y, x - 0.5]), vP1_2d)
 
-# Now, we set up the cosine-bell--cone--slotted-cylinder initial coniditon. The
+# Now, we set up the cosine-bell--cone--slotted-cylinder initial condition. The
 # first four lines declare various parameters relating to the positions of these
 # objects, while the analytic expressions appear in the last three lines. This
 # code is simply copied from the Firedrake version of the demo. ::
--- a/demos/demo_2d_tracer.py
+++ b/demos/demo_2d_tracer.py
@@ -40,37 +40,32 @@ options.fields_to_export = ['tracer_2d']
 
 # We will run for time :math:`2\pi` -- a full rotation -- using a
 # strong stability preserving third order Runge-Kutta method (SSPRK33).
-# We take 600 steps, giving a timestep close to the CFL limit. For
-# consistency with the Firedrake demo, Thetis' automatic timestep
-# computation functionality is switched off and the same solver
-# parameters are specified. ::
+# For consistency with the Firedrake demo, Thetis' automatic timestep
+# computation functionality is switched off and the simulation time is
+# split into 600 steps, giving a timestep close to the CFL limit. ::
 
 options.timestepper_type = 'SSPRK33'
 options.timestep = pi/300.0
 options.simulation_end_time = 2*pi
 options.simulation_export_time = pi/15.0
 options.timestepper_options.use_automatic_timestep = False
-options.timestepper_options.solver_parameters_tracer = {
-    'ksp_type': 'preonly',
-    'pc_type': 'bjacobi',
-    'sub_pc_type': 'ilu',
-}
 
 # We have a pure advection problem with no diffusivity or source terms. However,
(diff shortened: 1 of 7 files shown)


demos/demo_2d_tracer.py
```diff
@@ -0,0 +1,123 @@
...
+options.timestepper_type = 'SSPRK33'
+options.timestep = pi/300.0
+options.simulation_end_time = 2*pi
+options.simulation_export_time = pi/15.0
+options.timestepper_options.use_automatic_timestep = False
+options.timestepper_options.solver_parameters_tracer = {
+    'ksp_type': 'preonly',
+    'pc_type': 'bjacobi',
+    'sub_pc_type': 'ilu',
+}
```
tkarna, line 57: Why do we need to specify the solver options here? Do the default ones not work? Or consistency with Firedrake example? For a demo it might be good if we could omit such details...

docs/source/conf.py
```diff
@@ -170,3 +171,6 @@
 }
 
 autoclass_content = "both"
+
+#  -- Options for sphinxcontrib.bibtex ------------------------------------
+bibtex_bibfiles = ['demos/demo_references.bib']
```
tkarna, line 176: Nice. So far we haven't used proper links for the references but probably should.

thetis/solver2d.py
```diff
@@ -263,11 +263,12 @@ def create_function_spaces(self):
                 'Spherical mesh requires \'rt-dg\' or \'bdm-dg\' ' \
                 'element family.'
         # ----- function spaces: elev in H, uv in U, mixed is W
-        self.function_spaces.P0_2d = get_functionspace(self.mesh2d, 'DG', 0, name='P0_2d')
+        DG = 'DG' if self.mesh2d.ufl_cell().cellname() == 'triangle' else 'DQ'
```
tkarna, line 266: Thanks for fixing this. I always get confused here though. Is there a difference between DQ and dP on quads? Which is the one we want?

demos/demo_2d_tracer.py
```diff
@@ -0,0 +1,123 @@
...
+
+# We have a pure advection problem with no diffusivity or source terms. However,
+# such terms can be specified by replacing the ``None`` values below. For
+# consistency with the Firedrake demo, we do not use stabilization or slope
+# limiters, both of which are used by default in Thetis. ::
+
+options.use_lax_friedrichs_tracer = False
+options.horizontal_diffusivity = None
+options.tracer_source_2d = None
+options.use_limiter_for_tracers = False
```
tkarna, line 67: It might be worth mentioning that one can use the limiter to obtain a non-oscillatory solution.

demos/demo_2d_tracer.py
```diff
@@ -0,0 +1,123 @@
...
+
+# The background tracer value is imposed as an upwind inflow condition.
+# In general, this would be a ``Function``, but here we just use a ``Constant``
+# value. ::
+
+solver_obj.bnd_functions['tracer'] = {'value': {'on_boundary': Constant(1.0)}}
```
tkarna, line 73: This does not look right. The dict should be
```
{bnd_id: {'value': Constant(1.0)}, ...}
```
So in this case the bc is not applied at all, implying a closed boundary condition. Which seems to result in a reasonable solution. Using the 1.0 value gives slightly smaller error.

tkarna (approved):
Thanks a lot! Looks good to me, only minor comments

demos/demo_2d_tracer.py
jwallwork23: Good point!

thetis/solver2d.py
jwallwork23: Good point. I just put DQ because it was what Colin put in the original demo and also what Firedrake automatically changed DG to.

jwallwork23: There is a difference because I think DQ1 has four DoFs on a quad and dPc1 has three.

demos/demo_2d_tracer.py
jwallwork23: Good point, have removed them.

thetis/solver2d.py
tkarna: OK then I think we indeed want DQ1

tkarna (approved):

PR #240, opened 2021-05-05
Multiple tracers

jwallwork23:
A while ago it was mentioned that it would be useful to be able to have multiple 2D tracer fields.

This PR includes a proposed approach, as well as:
* various fixes for sphinx warnings
* make conservative tracer a subclass of 2D tracer
* allow P1+SUPG for conservative tracer

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -670,6 +670,19 @@ class ModelOptions2d(CommonModelOptions):
     use_supg_tracer = Bool(
         False, help="Use SUPG stabilisation in tracer advection").tag(config=True)
 
+    def __init__(self, *args, **kwargs):
+        self.tracer_sources_2d = {}
+        super(ModelOptions2d).__init__(*args, **kwargs)
+
+    def add_tracer_2d(self, label, source=None):
+        """
+        Define a 2D tracer field
+
+        :arg label: name for the tracer field
+        :kwarg source: associated source term
+        """
+        self.tracer_sources_2d[label] = source  # TODO: Validate input
+
 
 @attach_paired_options("timestepper_type",
                        PairedEnum([('LeapFrog', ExplicitTimestepperOptions3d),
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -685,6 +685,8 @@ class ModelOptions2d(CommonModelOptions):
         :kwarg unit: units for field, e.g. '-'
         :kwarg source: associated source term
         """
+        assert ' ' not in label, "Labels cannot contain spaces"
+        assert ' ' not in filename, "Filenames cannot contain spaces"
         self.tracer_metadata[label] = {
             'name': name,
             'shortname': shortname or name,
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -671,20 +671,21 @@ class ModelOptions2d(CommonModelOptions):
         False, help="Use SUPG stabilisation in tracer advection").tag(config=True)
 
     def __init__(self, *args, **kwargs):
-        self.tracer_data_2d = {}
-        super(ModelOptions2d).__init__(*args, **kwargs)
+        self.tracer_metadata = {}
+        super(ModelOptions2d, self).__init__(*args, **kwargs)
 
-    def add_tracer_2d(self, name, filename, shortname=None, unit='-', source=None):
+    def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=None):
         """
         Define a 2D tracer field
 
-        :arg name: name for the tracer field
-        :arg filename: file name for outputs
-        :kwarg shortname: short version of name
-        :kwarg unit: units for field
+        :arg label: field label used internally by Thetis, e.g. 'tracer_2d'
+        :arg name: human readable name for the tracer field, e.g. 'Tracer concentration'
+        :arg filename: file name for outputs, e.g. 'Tracer2d'
+        :kwarg shortname: short version of name, e.g. 'Tracer'
+        :kwarg unit: units for field, e.g. '-'
         :kwarg source: associated source term
         """
-        self.tracer_data_2d[filename] = {
+        self.tracer_metadata[label] = {
             'name': name,
             'shortname': shortname or name,
             'unit': unit,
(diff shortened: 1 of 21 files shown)


tkarna (dismissed):
Thanks a lot for implementing this, looks great! Only minor comments.

- Should the tracer labels also end with `_2d` like all other fields? Or should we remove the `_2d` suffix from all the fields?
- Handling the default `tracer_2d` case looks a bit cumbersome as it needs special checking
- Now that we have unknown number of equations they could be stored in a `FieldDict` as well. But that could be too intrusive of a change for this PR

tkarna:
To clarify the tracer naming business, I think we define 4 different strings for a field:
```
field_metadata['elev_2d'] = {
    'name': 'Water elevation',
    'shortname': 'Elevation',
    'unit': 'm',
    'filename': 'Elevation2d',
}
```
- the internal label (`elev_2d`) is lowercase and ends with `_2d`. This is used as the key in the `solver.fields` dict
- the filename is CamelCase and ends with `2d`. This is used in the file names
- name is human readable name
- the shortname is used as the field name in vtk output files
Maybe we should use the same conventions for the tracers as well?
(these are subject to change if we implement CF compliant outputs)

thetis/solver2d.py
```diff
@@ -321,23 +321,30 @@ def create_equations(self):
...
+                eq = conservative_tracer_eq_2d.ConservativeTracerEquation2D
             else:
-                uv_2d, elev_2d = self.fields.solution_2d.split()
-                self.eq_tracer = tracer_eq_2d.TracerEquation2D(
-                    self.function_spaces.Q_2d, self.depth, self.options, uv_2d)
+                eq = tracer_eq_2d.TracerEquation2D
+            for label in self.options.tracer_data_2d:
+                field_metadata[label] = self.options.tracer_data_2d[label].copy()
+                field_metadata[label].pop('source')
+                self.fields[label] = Function(self.function_spaces.Q_2d, name=label)
```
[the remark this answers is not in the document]
jwallwork23: Could do, yes. The simplest case would just be to assume that the added field lives in `'Q_2d'` for now, I guess.

jwallwork23:
If the default tracer name becomes 'tracer_2d' (and the special case which converts that to just 'tracer' is avoided) we lose backwards compatibility when assigning initial conditions (which previously expected 'tracer=...'). Is that okay?

tkarna:
> If the default tracer name becomes 'tracer_2d' (and the special case which converts that to just 'tracer' is avoided) we lose backwards compatibility when assigning initial conditions (which previously expected 'tracer=...'). Is that okay?

Yeah, `assign_initial_conditions` uses the field names without the `_2d` suffix as kwargs which is a bit weird. Maybe breaking the bw compatibility in this case is not a big issue? Or in `assign_initial_conditions` you could append `_2d` in the kwarg key when looking for the field?

thetis/solver2d.py
```diff
@@ -297,6 +297,18 @@ def create_function_spaces(self):
 
         self._isfrozen = True
 
+    def add_new_field(self, label):
```
tkarna, line 300: I guess I was thinking about something more generic such as
```python
def add_new_field(self, function, label, name, filename, shortname=None, unit='-'):
    metadata[label] = {
        'name': name, ...
    }
    field_metadata[label] = metadata
    self.fields[label] = function
```
That way users could add any e.g. diagnostic function and it could be exported via `options.fields_to_export`. In the special case of tracers you would just populate the args from `options.tracer_metadata`.

thetis/solver2d.py
```diff
@@ -315,35 +327,40 @@ def create_equations(self):
...
-                    self.function_spaces.Q_2d, self.depth, self.options)
+                eq = conservative_tracer_eq_2d.ConservativeTracerEquation2D
             else:
-                uv_2d, elev_2d = self.fields.solution_2d.split()
-                self.eq_tracer = tracer_eq_2d.TracerEquation2D(
-                    self.function_spaces.Q_2d, self.depth, self.options, uv_2d)
+                eq = tracer_eq_2d.TracerEquation2D
+            for label in self.options.tracer_metadata:
+                self.add_new_field(label)
+                setattr(self.equations, label, eq(*args))
```
tkarna, line 350: here you could use the dict interface:
```python
self.equations[label] = eq(*args)
```

thetis/solver2d.py
```diff
@@ -603,8 +621,11 @@ def assign_initial_conditions(self, elev=None, uv=None, tracer=None, sediment=No
             elev_2d.project(elev)
         if uv is not None:
             uv_2d.project(uv)
-        if tracer is not None and self.options.solve_tracer:
-            self.fields.tracer_2d.project(tracer)
+        if self.options.solve_tracer:
+            for label in self.options.tracer_metadata:
+                tracer = tracers.get(label)
+                if tracer is not None:
+                    self.fields[label].project(tracer)
```
tkarna, line 628: I guess you could loop the other way:
```python
for label, func in tracers.items():
    assert label in self.options.tracer_metadata, f'Unknown tracer label {label}'  # it's good to warn users
    self.fields[label].project(func)
```
Now if we wish to keep things consistent, i.e. the args are  `uv`, `elev`, `tracer` etc you could do
```
for l, func in tracers.items():
    label = l + '_2d'
    ...
```

tkarna (changes requested):
Thanks for the updates, looks really nice! a couple more comments

thetis/options.py
```diff
@@ -670,6 +670,31 @@ class ModelOptions2d(CommonModelOptions):
...
+        """
+        assert ' ' not in label, "Labels cannot contain spaces"
+        assert ' ' not in filename, "Filenames cannot contain spaces"
+        self.tracer_metadata[label] = {
+            'name': name,
+            'shortname': shortname or name,
+            'unit': unit,
+            'filename': filename,
+            'source': source,  # TODO: Validate input
+        }
```
tkarna, line 696: I guess another way would be to use nested options (like we do for timestepper_options for example). Then each entry would be a traitlet and we could validate the input (e.g. source). The API could look like:
```python
options.add_tracer(label)
options.tracer_options[label].source = foo
```
But the nested options can be a bit cumbersome to use in practice. I'd be happy to use this as it is for now. Maybe @stephankramer has some thoughts?

tkarna (approved):
Thanks for the updates, looks good!

thetis/options.py
jwallwork23: That's a good point. I'll merge for now and leave it as a `TODO`. Thanks for the review!

PR #241, opened 2021-05-17
remove geometric boundary mask

tkarna:
Firedrake boundary method `'geometric'` has been removed in firedrakeproject/firedrake#2007

~For most (all?) Thetis use cases there's actually no difference between _geometric_ and _topological_ boundary masks (e.g. P1DG).~

The top/bottom facet mask indices have now been implemented manually.

### DIFF
--- a/thetis/utility.py
+++ b/thetis/utility.py
@@ -147,6 +147,23 @@ def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
     return constructor(mesh, elt, **kwargs)
 
 
+def get_extruded_base_element(ufl_element):
+    """
+    Return UFL TensorProductElement of an extruded UFL element.
+
+    In case of a non-extruded mesh, returns the element itself.
+    """
+    if isinstance(ufl_element, ufl.HDivElement):
+        ufl_element = ufl_element._element
+    if isinstance(ufl_element, ufl.MixedElement):
+        ufl_element = ufl_element.sub_elements()[0]
+    if isinstance(ufl_element, ufl.VectorElement):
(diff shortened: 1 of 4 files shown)


wence-:
> (e.g. P1DG).

I'm guessing there isn't a difference because you use equispaced DG nodes?

tkarna:
Yes, we need equispaced for e.g. the slope limiters to work

thetis/utility.py
```diff
@@ -884,18 +903,11 @@ def __init__(self, input_2d, output_3d, elem_height=None):
...
-        if ufl_elem.family() == 'TensorProductElement':
-            # a normal tensorproduct element
-            family_3dh = ufl_elem.sub_elements()[0].family()
-            if family_2d != family_3dh:
-                raise Exception('2D and 3D spaces do not match: {0:s} {1:s}'.format(family_2d, family_3dh))
+        base_element_3d = get_extruded_base_element(self.fs_3d.ufl_element())
+        assert isinstance(base_element_3d, ufl.TensorProductElement)
+        family_3dh = base_element_3d.sub_elements()[0].family()
+        if family_2d != family_3dh:
+            raise Exception('2D and 3D spaces do not match: {0:s} {1:s}'.format(family_2d, family_3dh))
```
jwallwork23, line 910: Could maybe write another utility function called something like `check_element_consistency_2d-3d`, since this is repeated a couple of times later?

jwallwork23 (approved):
Looks good to me! Just one minor suggestion.

thetis/utility.py
tkarna: Yeah, sure. Will do in another PR.

PR #242, opened 2021-05-21
Recover vorticity in P1 space

jwallwork23:
I use a similar script all the time and thought maybe it'd be of use to others? You were right @tkarna , `add_new_field` turns out to be useful for things other than just adding tracers.

I wondered whether this should be a callback (like the total vorticity one in `examples/freshwaterCylinder`), but Thetis callbacks have HDF5 output, whereas I'm after VTK here.


tkarna:
Thanks for this update, handling user-defined diagnostic fields indeed needs more work. At the moment, solver has the `add_new_field` method and `ExportManager.add_export` accepts an optional `preproc_func` argument which defines a function that will be evaluated before each export. When defining a new diagnostic field, you can make a new class for computing the diagnostic variable, say `MyVorticityCalculator`. It has a `solve` method which updates the field. In `utility.py` there are a bunch of such solvers defined (side note: the file is too long and should be split at some point). So at the moment you can do

```python
vorticity_2d = Function(...)
vorticity_calculator = MyVorticityCalculator(vorticity_2d, ...)
solver_obj.add_new_field(vorticity_2d, 'vorticity_2d', ...)
...
solver_obj.exporters['vtk'].add_export(
    'vorticity_2d', vorticity_2d, export_type='vtk', preproc_func=vorticity_calculator.solve)
```

Some thoughts:
1. This process could be made smoother by adding the `preproc_func` argument to `add_new_field` and propagating the info to the ExportManager. Then the last step could be omitted.
2. So far the diagnostic fields have not been included in the options. Frequently used ones could be included I think. I guess in this case the user still needs to add the field to `options.fields_to_export` list?
3. I'd prefer to keep the diagnostic solvers outside the FlowSolver2d object to keep it from bloating too much.

jwallwork23:
I think this is more in line with what you were thinking.

I moved the 3D utility solvers to a new file as suggested by the TODO, maybe more could be moved, too.

jwallwork23:
I think the errors are due to `add_new_field` jumping ahead and creating the exporters (and timesteppers) from within `create_equations` when the `'tracer_2d'` field is added.

One solution could be to reorder the solver initialisation so that `create_exporters` is called before `create_equations`. Although maybe that would imply users' scripts might need reordering, too?

tkarna:
Thanks for the update.

> I think the errors are due to add_new_field jumping ahead and creating the exporters (and timesteppers) from within create_equations when the 'tracer_2d' field is added.

Yeah, `add_new_field` should just add the metadata and store the `preproc_func` somewhere (if provided; could be stored in an internal dict). Creating the exporters should be dealt with later. So you should remove the `export` kwarg. Now if the users really want to export the function, they just add the key in `options.fields_to_export` like for any other field. Then the user can decide whether to export in vtk or hdf5 (or some new format we'll be adding). So the new API would look like:

```python
vorticity_2d = Function(...)
vorticity_calculator = MyVorticityCalculator(vorticity_2d, ...)
solver_obj.add_new_field(vorticity_2d, 'vorticity_2d', ..., preproc_func=vorticity_calculator.solve)
solver_obj.options.fields_to_export = ['vorticity_2d', ...]
```
Does that make sense?

jwallwork23:
Makes sense! Thanks

tkarna (approved):
Thanks for the updates, all good now!

PR #244, opened 2021-06-17
Picard iteration for multiple tracers

jwallwork23:
At the minute, a system of multiple 2D tracers can be solved sequentially in any order and it won't make a difference. There is some interest in extending the 2D tracer model to include reaction terms. In that case, the order does matter and it will probably be worthwhile to iterate at each timestep.

This PR introduces a `CoupledTracerPicard` time integrator which does this. For some specified number of iterations, it solves each tracer equation sequentially.

### DIFF
--- a/thetis/configuration.py
+++ b/thetis/configuration.py
@@ -213,6 +213,31 @@ class PETScSolverParameters(Dict):
         self.error(obj, value)
 
 
+class CoupledTracerImplicitness(Dict):
+    info_text = 'a dictionary of implicness values for CoupledTracerPicard'
+
+    def __init__(self, default_value=Undefined, bounds=None, **kwargs):
+        self.minval = bounds[0]
+        self.maxval = bounds[1]
+        super(CoupledTracerImplicitness, self).__init__(default_value=default_value, **kwargs)
+
+    def validate(self, obj, value):
+        if (isinstance(value, dict)
+                and all(self.minval <= theta <= self.maxval for key, theta in value.items())):
+            return value
+        self.error(obj, value)
+
+
+class CoupledTracerSemiImplicitness(Dict):
+    info_text = 'a dictionary of bools toggling semi-implicitness for CoupledTracerPicard'
+
+    def validate(self, obj, value):
+        if (isinstance(value, dict)
+                and all(isinstance(semi_implicit, bool) for key, semi_implicit in value.items())):
+            return value
+        self.error(obj, value)
+
+
 class PairedEnum(Enum):
     """A enum whose value must be in a given sequence.
 
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -100,7 +100,8 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
         # compatible with the 2d coupled timeintegrator
         assert solution2d == self.fields.solution_2d
 
-        self.timesteppers.swe2d.initialize(self.fields.solution_2d)
+        if not self.options.tracer_only:
+            self.timesteppers.swe2d.initialize(self.fields.solution_2d)
         if self.options.solve_tracer:
             if self.coupled:
                 solutions = {
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -35,7 +35,8 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
         self.fields = solver.fields
         self.timesteppers = AttrDict()
         print_output('Coupled time integrator: {:}'.format(self.__class__.__name__))
-        print_output('  Shallow Water time integrator: {:}'.format(self.swe_integrator.__name__))
+        if not self.options.tracer_only:
+            print_output('  Shallow Water time integrator: {:}'.format(self.swe_integrator.__name__))
         if self.options.solve_tracer:
             print_output('  Tracer time integrator: {:}'.format(self.tracer_integrator.__name__))
         if self.options.sediment_model_options.solve_suspended_sediment:
@@ -51,12 +52,27 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
         Creates all time integrators with the correct arguments
         """
         self.timesteppers.swe2d = self.solver.get_swe_timestepper(self.swe_integrator)
-        if self.solver.options.solve_tracer:
-            for label in self.options.tracer_metadata:
-                self.timesteppers[label] = self.solver.get_tracer_timestepper(self.tracer_integrator, label)
-        if self.solver.options.sediment_model_options.solve_suspended_sediment:
+        if self.options.solve_tracer:
+            self.coupled = self.options.timestepper_type == 'CoupledTracerPicard'
+            if self.coupled:
+                equations, solutions, fields, bnd_conditions = {}, {}, {}, {}
+                for label in self.options.tracer_metadata:
+                    args, kwargs = self.solver_get_tracer_timestepper(self.tracer_integrator, label, coupled=True)
+                    equations[label], solutions[label], fields[label], dt = args
+                    bnd_conditions[label] = kwargs['bnd_conditions']
+                self.timesteppers.coupled_tracer = self.tracer_integrator(
+                    equations, solutions, fields, dt, bnd_conditions=bnd_conditions,
+                    solver_parameters=self.options.timestepper_options.solver_parameters_tracer,
+                    semi_implicit=self.options.timestepper_options.use_semi_implicit_linearization,
+                    theta=self.options.timestepper_options.implicitness_theta,
+                    iterations=self.options.timestepper_options.picard_iterations,
+                )
+            else:
+                for label in self.options.tracer_metadata:
+                    self.timesteppers[label] = self.solver.get_tracer_timestepper(self.tracer_integrator, label)
+        if self.options.sediment_model_options.solve_suspended_sediment:
             self.timesteppers.sediment = self.solver.get_sediment_timestepper(self.sediment_integrator)
-        if self.solver.options.sediment_model_options.solve_exner:
+        if self.options.sediment_model_options.solve_exner:
             self.timesteppers.exner = self.solver.get_exner_timestepper(self.exner_integrator)
 
     def set_dt(self, dt):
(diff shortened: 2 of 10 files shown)


thetis/timeintegrator.py
```diff
@@ -433,6 +433,113 @@ def advance(self, t, update_forcings=None):
...
+        """Advances equations for one time step."""
+        if update_forcings is not None:
+            update_forcings(t + self.dt)
+        for tracer, solution in self.solutions.items():
+            self.solutions_old[tracer].assign(solution)
+
+        for it in range(self.iterations):
+            for tracer, solver in self.solvers.items():
+                with timed_stage(f"{tracer} solve"):
+                    solver.solve()
```
jwallwork23, line 535: Just realised that the order cannot be trusted here. I guess we could use an `OrderedDict`?

tkarna: Sure, sounds like a good solution as the order of tracers matter.

jwallwork23:
I guess in principle this could also be used to solve the swe-sediment-exner coupled system. What do you think @mc4117 ? If so, I can change to a more general name and hook it up. Could also be a subsequent PR.

thetis/timeintegrator.py
```diff
@@ -433,6 +433,127 @@ def advance(self, t, update_forcings=None):
...
+        :kwarg theta: Implicitness parameter, which can be given as either a single value or
+            a dictionary corresponding to the different tracer fields
+        :kwarg bool semi_implicit: If True use a linearized semi-implicit scheme
+        :kwarg int iterations: Number of Picard iterations
+        """
+        super(CoupledTracerPicard, self).__init__(equations, solutions, fields, dt, solver_parameters)
+        self.equations = equations
+        self.solutions = solutions
+        delattr(self, 'equation')
+        delattr(self, 'solution')
```
tkarna, line 473: Hmm, what you have done here suggest that this is not a TimeIntegrator, but rather a CoupledTimeintegrator (as it handles multiple equations). Can this not be designed as one? That is, is there any difference between the present implementation and bundling up a a few CrNi Timeintegrators and stepping them in the right order? You've also re-implemented the Cr-Ni forms which is redundant (unless I'm missing something). So I'd move this to `coupled_timeintegrator_2d.py`. Need to think about the names I guess. This class only handles tracer equations, whereas `CoupledTimeIntegrator2D` and offspring also include swe (and sediments).

thetis/timeintegrator.py
```diff
@@ -433,6 +433,127 @@ def advance(self, t, update_forcings=None):
...
+        """Advances equations for one time step."""
+        if update_forcings is not None:
+            update_forcings(t + self.dt)
+        for tracer, solution in self.solutions.items():
+            self.solutions_old[tracer].assign(solution)
+
+        for it in range(self.iterations):
+            for tracer, solver in self.solvers.items():
+                with timed_stage(f"{tracer} solve"):
+                    solver.solve()
```
tkarna, line 549: As we've declared theta>=0.5 in options, this indeed always means a Picard iteration? Do you think it would be possible to somehow monitor convergence? PETSc has these nice monitors which as quite useful in practice.

tkarna (changes requested):
Thanks, this looks nice! My main comment is that we could indeed use the new tracer integrator for all cases: remove the old unordered version, and also include sediments here. But it might be easier to do that in a separate PR once this has landed.

thetis/coupled_timeintegrator_2d.py
```diff
@@ -50,14 +51,36 @@ def _create_integrators(self):
...
-                self.timesteppers[label] = self.solver.get_tracer_timestepper(self.tracer_integrator, label)
-        if self.solver.options.sediment_model_options.solve_suspended_sediment:
-            self.timesteppers.sediment = self.solver.get_sediment_timestepper(self.sediment_integrator)
-        if self.solver.options.sediment_model_options.solve_exner:
-            self.timesteppers.exner = self.solver.get_exner_timestepper(self.exner_integrator)
+        if not self.options.tracer_only:
+            args, kwargs = self.solver.setup_swe_timestepper()
+            self.timesteppers.swe2d = self.swe_integrator(*args, **kwargs)
+        if self.options.solve_tracer:
+            self.coupled = self.options.timestepper_type == 'CoupledTracerPicard'
```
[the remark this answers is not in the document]
jwallwork23: Good point!

thetis/coupled_timeintegrator_2d.py
```diff
@@ -50,14 +51,36 @@ def _create_integrators(self):
...
+                self.timesteppers.coupled_tracer = self.tracer_integrator(
+                    equations, solutions, fields, dt, bnd_conditions=bnd_conditions,
+                    solver_parameters=self.options.timestepper_options.solver_parameters_tracer,
+                    semi_implicit=semi_implicit, theta=theta,
+                    iterations=self.options.timestepper_options.picard_iterations,
+                )
+            else:
+                for label in self.options.tracer_metadata:
+                    args, kwargs = self.solver.setup_tracer_timestepper(label)
+                    self.timesteppers[label] = self.tracer_integrator(*args, **kwargs)
```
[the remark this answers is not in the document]
jwallwork23: True, that would be equivalent

thetis/timeintegrator.py
jwallwork23: I think the only difference is making sure that the solves are interleaved during `advance`. i.e.
```
for it in range(self.iterations):
    for solver in self.solvers:
        solver.solve()
```
rather than
```
for solver in self.solvers:
    for it in range(self.iterations):
        solver.solve()
```

thetis/coupled_timeintegrator_2d.py
```diff
@@ -121,7 +165,35 @@ def __init__(self, solver, integrator):
         super(CoupledMatchingTimeIntegrator2D, self).__init__(solver)
 
 
+class CoupledTracerTimeIntegrator2D(CoupledTimeIntegrator2D):
```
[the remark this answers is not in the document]
jwallwork23: Good point. I do find it difficult naming things :)

thetis/configuration.py
```diff
@@ -213,6 +213,31 @@ def validate(self, obj, value):
...
+class CoupledTracerSemiImplicitness(Dict):
+    info_text = 'a dictionary of bools toggling semi-implicitness for CoupledTracerPicard'
+
+    def validate(self, obj, value):
+        if (isinstance(value, dict)
+                and all(isinstance(semi_implicit, bool) for key, semi_implicit in value.items())):
+            return value
+        self.error(obj, value)
+
+
```
stephankramer, line 240: I think this is way too specific for a new Traitlet type. A Traitlet type here should be a type that implements a type-checking validate method, so it should only be as specific as required to describe what sort of data is allowed. Here you seem to be implementing a `BoundedFloatDict` and a `BoolDict`.

Note however that traitlets allows you to compose the `Dict` type with a specified trait:
```
class MyOptions(HasFrozenTraits):
    x = Dict(value_trait=BoundedFloat(bounds=[0,1]), default_value={})
    y = CoupledTracerImplicitness(bounds=[0,1])
```
Both work as expected, but the former gives a slightly more useful error message:
```
z = MyOptions()
z.x = {'a': 1.5}
...
TraitError: Values of the 'x' trait of a MyOptions instance must be a float between 0 and 1, but a value of 1.5 <class 'float'> was specified.

z.y = {'b': -0.5}

TraitError: The 'y' trait of a MyOptions instance expected a dict, not the dict {'b': -0.5}
```

However there is a more fundamental problem with both approaches: the values only get type checked when assigning a whole new dictionary to the "bounded float dict" attribute, *not* when the dictionary is subsequently modified:
```
z = MyOptions()
z.x['a'] = 1.5
z.y['b'] = -0.5
```

Finally I'm not sure (up for discussion really) whether we want to go the route of having a series of dictionaries all with the same keys, for all the various options we want to set per field as the number of these options is inevitably going to increase. To me it would make more sense to have a hierarchy, where you only have one dictionary whose values are entire `FrozenHasTraits` objects with all the options for every field grouped together at that level. So something like:
```
class TracerFieldOptions(FrozenHasTraits):
    theta = BoundedFloat(bounds=[0,1])
    semi_implicit = Bool()

class MyOptions:
    tracer_options = Dict(value_trait=Instance(TracerFieldOptions), default_value={})
```
Then the user would
```
tracer_options_a = TracerFieldOptions()
tracer_options_a.theta = 0.5
tracer_options_a.semi_implicit = True
tracer_options_b = TracerFieldOptions()
tracer_options_b.theta = 0.5
tracer_options_b.semi_implicit = True
options = MyOptions() # this would be done by Thetis of course
options.tracer_options = {'a': tracer_options_a, 'b': tracer_options_b}
```
This has the advantage that type checking happens not only when the `TracerFieldOptions` are first filled in, but if the options are subsequently changed:
```
options.tracer_options['a'].theta = 1.5
```
gives the appropriate error. The user can however still shoot themselves in the foot  (in the same way as before) by doing:
```
options.tracer_options['a'] = 'foo' # or anything that's not a TracerFieldOptions instance
```
A disadvantage is that it exposes the user to this `TracerFieldOptions` class that the user has to instantiate so it adds some complexity...

I would really like to hear other people's opinions as the same issue applies to specifying multiple turbine farms, where we have a dictionary that maps region ids (2d gmsh physical surface ids) to options for the different farms. Soon to be expanded with a lot more options.

jwallwork23: Great point, using separate options seems much nicer, indeed. I get the point about user friendliness, though.

I guess in this case, the `ModelOptions2d.add_tracer_2d` interface could pass some of this through, but then we might end up with lots of different kwargs to account for all the timesteppers.

tkarna: Good point, Stephan! I agree that option hierarchy is probably the right way to go. As Joe says `add_tracer_2d` could indeed be used to reduce boilerplate e.g. by populating the parameters with default values. In passing I mention that we do currently have to use quite a bit of `hasattr` checks to sniff whether certain option exists which isn't very nice. I think it's the only downside of having more dynamic options structure.

stephankramer: Ah yes, forgot we already have `add_tracer_2d` indeed. So maybe `tracer_field_options` (or however we'd call it) should be some sort of frozen dict that the user can't add entries to other than through calling add_tracer_2d, so like tracer_metadata but with the values being a TracerFieldOptions instance in which all the options are set for each field including name/shortname/filename. So the TracerFieldOptions instance gets created by add_tracer_2d (so the creation isn't exposed to the user) and added as an entry in the `tracer_field_options` dictionary. That way we enforce that the user sets the minimum for a tracer field to be functional, but any further options like e.g. `source`, take a default and can be set as attributes of the TracerFieldOptions instance: `options.tracer_field_options['foo'].source = Function(...)`

On the topic (slightly off) of `hasattr` - I think this is mainly an issue of feeding in the options at the right level to the corresponding class in the object hierarchy. In particular looking at the many `hasattr`s in the various `get_..._timestepper` methods of the solver objects - it probably makes more sense to just pass in the entire `timestepper_options` (a TimeStepperOptions (or subclass) instance) to the time integrator, rather than unpacking it via various keyword arguments.

jwallwork23: 1. Sounds like a good solution to me.
2. I agree that would neaten things up. That way the `get_..._timestepper` methods just serve to feed in the appropriate parameters for the equation, rather than the timestepper (and therefore should probably be renamed).

Thanks both for the feedback, starting to think this will be easier to implement on a new branch, so will do that.

jwallwork23: [This ](https://github.com/thetisproject/thetis/pull/247) should do the trick for reducing boilerplate and standardising args/kwargs for the timesteppers.

jwallwork23:
Closing because this will be done in a sequence of PRs, starting with #247.

PR #245, opened 2021-06-28
Transect callback

tkarna:
Adds a callback for storing structured grid transects of 3D fields. Transect is defined by user-specified (x,y) coordinates and number of points in the vertical direction. The vertical coordinate of the transect adapts to the moving grid. The callback is included in the lock exchange example so that it'll be executed in the test suite.

### DIFF
--- a/thetis/callback.py
+++ b/thetis/callback.py
@@ -881,7 +881,7 @@ class TransectCallback(DiagnosticCallback):
     name = 'transect'
     variable_names = ['z_coord', 'value']
 
-    def __init__(self, solver_obj, fieldname, x, y,
+    def __init__(self, solver_obj, fieldnames, x, y,
                  location_name,
                  n_points_z=48,
                  z_min=None, z_max=None,
@@ -889,14 +889,16 @@ class TransectCallback(DiagnosticCallback):
                  append_to_log=True):
         """
         :arg solver_obj: Thetis :class:`FlowSolver` object
-        :arg fieldname: Field to extract
-        :arg x, y: list of location coordinates in model coordinate system
+        :arg fieldnames: List of fields to extract
+        :arg x, y: location coordinates in model coordinate system.
         :arg location_name: Unique name for this location. This
             determines the name of the output h5 file (prefixed with
             `diagnostic_`).
-        :arg int npoints: Number of points along the vertical axis. The 3D
+        :arg int n_points_z: Number of points along the vertical axis. The 3D
             field will be interpolated on these points, ranging from the bottom
             to the (time dependent) free surface.
+        :kwarg float z_min, zmax: Force min/max value of z coordinate extent.
+            By default, transect will cover entire depth from bed to surface.
         :kwarg str outputdir: Custom directory where hdf5 files will be stored.
             By default solver's output directory is used.
         :kwarg bool export_to_hdf5: If True, diagnostics will be stored in hdf5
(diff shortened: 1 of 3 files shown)


thetis/callback.py
```diff
@@ -861,3 +870,151 @@ def message_str(self, *args):
...
+                error('{:}: Cannot evaluate data on transect {:}'.format(self.__class__.__name__, self.location_name))
+                raise e
+            shape = list(self.value_shape) + [field_dim]
+            arr = np.array(vals).reshape(shape)
+            if field_dim == 3:
+                outvals.extend([arr[..., 0], arr[..., 1], arr[..., 2]])
+            elif field_dim == 2:
+                outvals.extend([arr[..., 0], arr[..., 1]])
+            else:
+                outvals.extend([arr[..., 0]])
```
stephankramer, line 1009: I think this could be just one line:
```
outvals.extend(arr.T)
```

thetis/callback.py
```diff
@@ -861,3 +870,151 @@ def message_str(self, *args):
...
+            to the (time dependent) free surface.
+        :kwarg float z_min, zmax: Force min/max value of z coordinate extent.
+            By default, transect will cover entire depth from bed to surface.
+        :kwarg str outputdir: Custom directory where hdf5 files will be stored.
+            By default solver's output directory is used.
+        :kwarg bool export_to_hdf5: If True, diagnostics will be stored in hdf5
+            format
+        :kwarg bool append_to_log: If True, callback output messages will be
+            printed in log
+        """
```
stephankramer, line 908: Might be nice to explicitly say it only prints a min max of each field - as I would be worried as a user this would dump a whole lot of data in the log.

thetis/callback.py
```diff
@@ -89,20 +90,27 @@ def __init__(self, filename, varnames, array_dim=1, attrs=None,
...
                 for var in self.varnames:
-                    hdf5file.create_dataset(var, (0, array_dim),
-                                            maxshape=(None, array_dim), dtype=dtype)
+                    dim_list = array_dim
+                    if isinstance(dim_list, tuple):
+                        dim_list = list(array_dim)
+                    elif not isinstance(dim_list, list):
+                        dim_list = list([array_dim])
+                    shape = tuple([0] + dim_list)
+                    max_shape = tuple([None] + dim_list)
```
stephankramer, line 99: Why is this inside the loop - none of it depends on `var` - unless I'm missing something...

thetis/callback.py
```diff
@@ -861,3 +870,151 @@ def message_str(self, *args):
...
+        return tuple(outvals)
+
+    def message_str(self, *args):
+        out = 'Evaluated transect "{:}":\n'.format(self.location_name)
+        for fieldname, arr in zip(self.variable_names[1:], args[1:]):
+            minval = arr.min()
+            maxval = arr.max()
+            out += '  {:} range: {:.3g} - {:.3g}\n'.format(
+                fieldname, minval, maxval)
+        out = out[:-1]  # suppress last line break
```
stephankramer, line 1019: It's probably more pythonic to do something like:
```
'\n'.join(list of output strings)
```

stephankramer (dismissed):
Some minor comments - looks great and useful!

thetis/callback.py
tkarna: `outvals` is actually a list in callback objects. But I can remove the if tree with
```python
            components = [arr[..., i] for i in range(arr.shape[-1])]
            outvals.extend(components)
```

stephankramer: Yeah - actually it doesn't matter that it's a list, extend takes any iterable, so you can just iterate over `a.T` but on second thought that would actually flip the remaining dimensions (nxy and nz) as well which you probably don't want - so your solution looks good

stephankramer (approved):

PR #247, opened 2021-06-29
Pass TimeStepperOptions to timesteppers

jwallwork23:
Also add some checks to `add_new_tracer`, as discussed in last Thetis meeting.

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -8,6 +8,7 @@ from .configuration import *
 from firedrake import Constant
 from .sediment_model import SedimentModel
 from .utility import AttrDict
+from collections import OrderedDict
 
 
 class TimeStepperOptions(FrozenHasTraits):
@@ -389,6 +390,13 @@ class TracerFieldOptions(FrozenHasTraits):
         None, allow_none=True, help='Source term for the tracer equation')
     diffusivity = FiredrakeScalarExpression(
         None, allow_none=True, help='Diffusion coefficient for the tracer equation')
+    metadata = Dict({
+        'label': 'tracer_2d',
+        'name': 'Tracer concentration',
+        'filename': 'Tracer2d',
+        'shortname': 'Tracer',
+        'unit': '-',
+    }, help='Dictionary of metadata for the tracer field')
 
 
 @attach_paired_options("free_surface_timestepper_type",
@@ -738,7 +746,7 @@ class ModelOptions2d(CommonModelOptions):
         False, help="Use SUPG stabilisation in tracer advection").tag(config=True)
 
     def __init__(self, *args, **kwargs):
-        self.tracer_metadata = {}
+        self.tracer = OrderedDict()
         super(ModelOptions2d, self).__init__(*args, **kwargs)
 
     def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=None, diffusivity=None):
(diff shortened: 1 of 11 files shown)


thetis/options.py
```diff
@@ -726,16 +751,27 @@ def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=
...
+        options.source = source
+        options.diffusivity = diffusivity
+        self.tracer_metadata[label] = AttrDict({
             'name': name,
             'shortname': shortname or name,
             'unit': unit,
             'filename': filename,
-            'source': source,  # TODO: Validate input
-        }
+            'options': options,
```
tkarna, line 773: So now we can access tracer options like this
```python
options.tracer_metadata[label].name
options.tracer_metadata[label].options.diffusivity
```
maybe this would be more logical
```python
options.tracer[label].metadata.name
options.tracer[label].diffusivity
```
i.e. the fields are not inside the metadata dict?

thetis/solver2d.py
```diff
@@ -331,6 +331,15 @@ def add_new_field(self, function, label, name, filename, shortname=None, unit='-
...
         """
+        assert isinstance(function, Function)
+        assert isinstance(label, str)
+        assert isinstance(name, str)
+        assert isinstance(filename, str)
+        assert shortname is None or isinstance(shortname, str)
+        assert isinstance(unit, str)
+        assert preproc_func is None or callable(preproc_func)
+        if label in self.fields:
+            print_output(f"Field '{label}' already exists. It will be overwritten.")
```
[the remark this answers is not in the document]
tkarna: While at it could you please add the equivalent to solver 3d as well?

tkarna (commented):
Thanks, this is neat! Only have one comment on the tracer options API

thetis/solver2d.py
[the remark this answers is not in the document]
jwallwork23: I did initially put it in as an assertion, but some of the tests crashed because they initialise twice (I think without meaning to). Happy to switch it and then modify those tests. Can do for 3D, too.

thetis/options.py
jwallwork23: Good idea, makes sense

thetis/solver2d.py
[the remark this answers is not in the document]
jwallwork23: Got it, the reason it was already in there is because the default 'tracer_2d' is hard-coded into `field_metadata`. Have removed it and made sure that it isn't in there when a solver object is created.

thetis/solver2d.py
```diff
@@ -136,6 +136,8 @@ def __init__(self, mesh2d, bathymetry_2d, options=None):
 
         self.bnd_functions = {'shallow_water': {}, 'tracer': {}, 'sediment': {}}
 
+        if 'tracer_2d' in field_metadata:
+            field_metadata.pop('tracer_2d')
```
tkarna, line 140: why is this needed?

thetis/solver2d.py
```diff
@@ -374,28 +384,22 @@ def create_equations(self):
...
-                self.add_new_field(Function(self.function_spaces.Q_2d, name='tracer_2d'),
-                                   'tracer_2d',
-                                   md['name'],
-                                   md['filename'],
-                                   shortname=md['shortname'],
-                                   unit=md['unit'])
+            if self.options.tracer == {}:
+                self.options.add_tracer_2d('tracer_2d', 'Depth averaged tracer', 'Tracer2d', 'Tracer', '-',
+                                           source=self.options.tracer_source_2d,
+                                           diffusivity=self.options.horizontal_diffusivity)
```
tkarna, line 390: I guess this is for backwards compatibility. Should we deprecate this code path in the future? That is, use always has to define tracers via `options.add_tracer_2d`? That would imply removing `options.solve_tracer`.

thetis/solver2d.py
```diff
@@ -828,7 +776,7 @@ def print_state(self, cputime):
         :arg float cputime: Measured CPU time
         """
         if self.options.tracer_only:
-            for l in self.options.tracer_metadata:
+            for l in self.options.tracer:
```
tkarna, line 779: `l` -> `label`? I think PEP8 suggests not to use variable name `l`

tkarna (approved):
Looks good! Only minor comments

thetis/solver2d.py
jwallwork23: Sure. We would then also be able to drop `options.horizontal_diffusivity` and `options.tracer_source_2d`. I'll add in a deprecation warning.

thetis/solver2d.py
jwallwork23: It is needed in the tests which do convergence studies. They create a sequence of solver objects and solve on each one. If this isn't included then the 2nd solver object tries to add a tracer field that already exists when it is instantiated.

tkarna: Ok, so this could plausibly be removed when we deprecate `options.solve_tracer`. We still might need a workaround in the test but it's better to have it there rather than the solver object.

tkarna (approved):

PR #251, opened 2021-07-02
Different timesteppers in 2D model

jwallwork23:
Whilst this introduces a few extra options (e.g. `tracer_timestepper_type`, `sediment_timestepper_type`, ...), it will make the 2D coupled time integrators much more flexible.


tkarna (commented):
Looks good. It's a bit strange that solver options can be passed to time integrator either via the options object or the solver_parameters kwarg.

thetis/solver2d.py
```diff
@@ -580,29 +585,30 @@ def create_timestepper(self):
...
             'SSPIMEX': implicitexplicit.IMEXLPUM2,
         }
-        if self.options.timestepper_type not in steppers:
-            raise Exception('Unknown time integrator type: {:s}'.format(self.options.timestepper_type))
         if self.options.nh_model_options.solve_nonhydrostatic_pressure:
             self.poisson_solver = DepthIntegratedPoissonSolver(
                 self.fields.q_2d, self.fields.uv_2d, self.fields.w_2d,
                 self.fields.elev_2d, self.depth, self.dt, self.bnd_functions,
-                solver_parameters=self.options.nh_model_options.solver_parameters
+                solver_parameters=self.options.nh_model_options.free_surface_timestepper_options.solver_parameters
```
stephankramer, line 592: This is not right I think? Why are we reusing the solver parameters of the nh free surface timestepper?

thetis/options.py
```diff
@@ -417,14 +443,17 @@ class NonhydrostaticModelOptions(FrozenHasTraits):
     q_degree = NonNegativeInteger(
         None, allow_none=True, help="Polynomial degree of the non-hydrostatic pressure space").tag(config=True)
     update_free_surface = Bool(True, help='Update free surface elevation after pressure projection/correction step').tag(config=True)
-    solver_parameters = PETScSolverParameters({
```
stephankramer, line 420: Why are `solver_parameters` disappearing? There should be two independent sets of `solver_parameters` here: 1) this one for the NH Poisson solve and 2) the one under the free surface timestepper. 1) is not associated with any timestepper because it's not got a time derivative

thetis/timeintegrator.py
```diff
@@ -67,7 +67,10 @@ def __init__(self, equation, solution, fields, dt, options, solver_parameters=No
         self.name = '-'.join([self.__class__.__name__,
                               self.equation.__class__.__name__])
         self.solver_parameters = {}
-        if solver_parameters is not None:
+        if solver_parameters is None:
+            if hasattr(options, 'solver_parameters'):
+                self.solver_parameters.update(options.solver_parameters)
+        else:
             self.solver_parameters.update(solver_parameters)
```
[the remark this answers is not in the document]
stephankramer: Yes I agree - I'd say we get rid of the optional `solver_parameters` kwarg. That is, hopefully we don't have to "override" it anywhere...

thetis/options.py
```diff
@@ -119,14 +135,24 @@ class ExplicitTimestepperOptions2d(ExplicitTimestepperOptions):
         'sub_pc_type': 'ilu',
         'mat_type': 'aij',
     }).tag(config=True)
-    solver_parameters_tracer = PETScSolverParameters({
+
+
+class ExplicitTracerTimestepperOptions2d(ExplicitTimestepperOptions2d):
```
stephankramer, line 140: Should these be subclassed from `ExplicitTimestepperOptions2d` and not directly from `ExplicitTimestepperOptions`. Of course it doesn't matter at the moment, but if this sets the ts options for tracer, and `ExplicitTimestepperOptions2d` for the swe then they are in my mind created equal - and if in the future we need to add an option that is specific to swe, we don't necessarily want to inherit it.

thetis/options.py
```diff
@@ -110,7 +123,10 @@ class PressureProjectionTimestepperOptions2d(TimeStepperOptions):
 
 
 class ExplicitTimestepperOptions2d(ExplicitTimestepperOptions):
-    """Options for 2d explicit time integrator"""
+    """
```
stephankramer, line 126: should we rename it `ExplicitSWETimestepperOptions2d` then? And similar for cases below? Also given my next comment on inheritance, I think it's clearer to explain what the relation between the two are.

thetis/timeintegrator.py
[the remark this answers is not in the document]
jwallwork23: Good point. I guess it makes sense to just pass the `solver_parameters` in the timestepper options. Then, even if someone wants to build their own `TimeIntegrator`, they still get to pass their own timestepper options anyway.

stephankramer (changes requested):

jwallwork23:
Okay I _think_ the tests should pass now.

`ModelOptions2d.timestepper_type` is dropped in favour of the separate types for each component. However, you can set the timestepper type across all of them using the `ModelOptions2d.set_timestepper_type` method. A nice bonus is that you can pass kwargs with that and avoid the `if hasattr(options.timestepper_options, 'use_automatic_timestep'): ...` kind of checking.

Whilst the 3D model components all get their own timestepper options, too, I decided to just keep the two LeapFrog and SSPRK22 coupled time integrators for now, rather than allowing full customisation. This could be done later. It means that `ModelOptions3d.timestepper_type` still exists and defines the coupled timeintegrator.

thetis/options.py
```diff
@@ -119,56 +144,106 @@ class ExplicitTimestepperOptions2d(ExplicitTimestepperOptions):
...
-        'pc_type': 'fieldsplit',
-        'pc_fieldsplit_type': 'multiplicative',
-    }).tag(config=True)
 
 
-class ExplicitTimestepperOptions3d(ExplicitTimestepperOptions):
-    """Base class for all 3d time stepper options"""
-    solver_parameters_2d_swe = PETScSolverParameters({
+class TimeStepperOptions3d(TimeStepperOptions):
+    """Options for 3d explicit time integrator"""
```
tkarna, line 174: `TimeStepperOptions3d` inherits empty `solver_parameters` dict from `TimeStepperOptions`. This is a bit confusing as it is not used (could actually cause bugs). Maybe better to inherit from `FrozenHasTraits` directly? The reason for the differentiation is that `TimeStepperOptions3d` is a _collection_ of timestepper options, not really an instance of that kind itself.

thetis/solver.py
```diff
@@ -840,7 +840,7 @@ def create_timestepper(self):
         elif self.options.timestepper_type == 'SSPRK22':
             self.timestepper = coupled_timeintegrator.CoupledTwoStageRK(weakref.proxy(self))
         else:
-            raise Exception('Unknown time integrator type: '+str(self.options.timestepper_type))
+            raise NotImplementedError("Custom coupled time integrators not yet accounted for")  # TODO
```
tkarna, line 843: `NotImplementedError(f'Unsupported time integrator type: {str(self.options.timestepper_type)}')`

tkarna (dismissed):
Thanks for the updates, looks good to me. Maybe @stephankramer wants to have another look?

thetis/options.py
```diff
@@ -778,14 +898,40 @@ def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=
...
 @attach_paired_options("timestepper_type",
-                       PairedEnum([('LeapFrog', ExplicitTimestepperOptions3d),
-                                   ('SSPRK22', ExplicitTimestepperOptions3d),
+                       PairedEnum([('LeapFrog', LeapFrogTimeStepperOptions3d),
+                                   ('SSPRK22', SSPRKTimeStepperOptions3d),
                                    ],
                                   "timestepper_options",
                                   default_value='SSPRK22',
-                                  help='Name of the time integrator').tag(config=True),
+                                  help='Name of the shallow water time integrator').tag(config=True),
```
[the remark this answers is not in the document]
jwallwork23: Thanks

thetis/options.py
```diff
@@ -778,14 +898,40 @@ def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=
...
+            self.sediment_model_options.sediment_timestepper_type = timestepper_type
+        if self.sediment_model_options.solve_exner:
+            options += (self.sediment_model_options.exner_timestepper_options,)
+            self.sediment_model_options.exner_timestepper_type = timestepper_type
+        if self.nh_model_options.solve_nonhydrostatic_pressure:
+            options += (self.nh_model_options.free_surface_timestepper_options,)
+        for option in options:
+            for key, value in kwargs.items():
+                if hasattr(option, key):
+                    setattr(option, key, value)
```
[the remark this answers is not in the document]
jwallwork23: Good point. I did this to avoid repeatedly setting the same parameter for the different models, but I'm probably just being lazy.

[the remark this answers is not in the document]
tkarna: Yeah, I do like the fact that you can set the parameters more easily this way, just pointing out that now the user will not be warned if the try to set a non-existent parameter. As the time stepper is the same in all cases, can we assume that all the `options` must have the user-specified attribute (`key`)? That would be easy to check.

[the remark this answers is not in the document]
jwallwork23: I think if we drop the `hasattr` then the fact that the timestepper options are frozen should sort this out.

thetis/options.py
```diff
@@ -746,11 +864,13 @@ class ModelOptions2d(CommonModelOptions):
 
     def __init__(self, *args, **kwargs):
         self.tracer = OrderedDict()
-        super(ModelOptions2d, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
 
     def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=None, diffusivity=None):
         """
-        Define a 2D tracer field
+        Add a 2D tracer field to :attr:`tracer_metadata`.
```
stephankramer, line 871: I believe this is no longer called tracer_metadata

thetis/options.py
```diff
@@ -778,14 +898,37 @@ def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=
...
+            options += (self.sediment_model_options.sediment_timestepper_options,)
+        if self.sediment_model_options.solve_exner:
+            self.sediment_model_options.exner_timestepper_type = timestepper_type
+            options += (self.sediment_model_options.exner_timestepper_options,)
+        if self.nh_model_options.solve_nonhydrostatic_pressure:
+            self.nh_model_options.free_surface_timestepper_type = timestepper_type
+            options += (self.nh_model_options.free_surface_timestepper_options,)
+        for key, value in kwargs.items():
+            for option in options:
+                setattr(option, key, value)
```
stephankramer, line 922: :+1:

stephankramer:
Just had another look - assuming the tests are fixed - all looks good to me now. Thanks for this @jwallwork23

jwallwork23:
Thanks @stephankramer!

tkarna (approved):
Great, thanks for the updates, Joe!

Issue #250, opened 2021-07-02
Is `bnd_conditions=None` really a valid choice?


jwallwork23:
I noticed that the default values for the `bnd_conditions` kwarg in all equation terms is `None`. However, only in the tracer equations is this actually a valid input. In the other cases, `bnd_conditions.get` is called without checking it is a dictionary.

Would it be safer to default to `bnd_conditions={}` throughout? We could then drop the `if bnd_conditions is not None` statements used in the tracer equations.

stephankramer:
Yeah good catch, that's bad - we shouldn't have a default value that isn't actually valid. 

However, using `{}` as the default for a keyword argument is generally considered unsafe and seen as an anti-pattern in python. The reason is that the default value is only created once, when the function is defined, and not as you might expect, when the function is called. This means that all calls to that function share the same default value, i.e. the same instance of that empty dictionary, and if anyone, inside or outside of that function changes it, say by insertion, that changed value will be the default in any subsequent calls. Thus it is widely advised for mutable (like dictionaries and lists) defaults, to instead use kwarg=None and assign the default value inside the function:
```
   if bnd_conditions is None:
      bnd_conditions = {}
```

Now in principle, since we don't actually modify `bnd_conditions` inside the residual methods, and don't give the caller access to it (i.e. we don't return it in some form) - it should be fine to use `bnd_conditions={}` here, but since this is such a common pitfall in cases where it does matter it would kind of stand out like a sore thumb, where you'd constantly be thinking "is this actually safe?".

So since it seems in most cases not providing `bnd_conditions` doesn't actually work - I would much prefer `bnd_conditions` to not be optional, i.e. for it not to have a default value. If there are cases where `residual()` is called without it (couldn't find one in a quick grep, except for calling `BathymetryDisplacementMassTerm.residual()` which doesn't even have it) we can just slot in a `{}` there.

tkarna:
Good catch indeed! I think the idea in the API has been that in case of no boundary conditions, you'd use `bnd_conditions=None` to avoid the mutable default value issue Stephan mentioned. But indeed in this case we could just drop the default value. In `residual` we are already assuming that `fields` is a dict without any type checking so we can do the same for `bnd_conditions`.

jwallwork23:
Thanks for that, I wasn't aware of the issue with `{}`. Changing from kwarg to arg seems like a good solution.

jwallwork23:
Resolved

PR #252, opened 2021-07-03
Make `bnd_conditions` an argument

jwallwork23:
See [this issue](https://github.com/thetisproject/thetis/issues/250)


thetis/solver2d.py
```diff
@@ -509,14 +509,14 @@ def get_tracer_timestepper(self, integrator, label):
...
         kwargs = dict(solver_parameters=self.options.timestepper_options.solver_parameters_tracer)
         if label in self.bnd_functions:
-            kwargs['bnd_conditions'] = self.bnd_functions[label]
+            args += (self.bnd_functions[label],)
         elif label[:-3] in self.bnd_functions:
-            kwargs['bnd_conditions'] = self.bnd_functions[label[:-3]]
+            args += (self.bnd_functions[label[:-3]],)
         else:
-            kwargs['bnd_conditions'] = {}
+            args += ({},)
```
tkarna, line 519: maybe cleaner to avoid appending to `args`, so something like:
```python
bcs = {}
if label in self.bnd_functions:
    bcs = self.bnd_functions[label]
args = (self.equations[label], self.fields[label], fields, self.dt, bcs)
```

tkarna (approved):
Thanks for this fix, looks good! I just have a minor suggestion on syntax

tkarna (approved):

tkarna (approved):

PR #254, opened 2021-07-20
Fix passing of timestepper options.

stephankramer:
Since #247 TimeStepperOptions are passed directly to the
timeintegrators. For this an options argument was added to __init__ of
TimeIntegrator and all its subclasses, however the subclasses didn't
actually pass on the new argument when calling super().__init__ I think
the reason this didn't break anything was because the _optional_
solver_parameters was substituted for the required options argument -
because the solver_parameters weren't passed through a keyword in the
call. This means that _all_ solver_parameters will have been assumed
not-provided which probably only worked because it then uses Firedrake's
default of using mumps.
Better to always use the keyword in the call to a function for keyword
arguments - as the convention is that the order of these may change.
Also it masks any mismatch in required arguments like here.

### DIFF
--- a/thetis/implicitexplicit.py
+++ b/thetis/implicitexplicit.py
@@ -31,7 +31,7 @@ class IMEXGeneric(TimeIntegrator):
         pass
 
     def __init__(self, equation, solution, fields, dt, options, bnd_conditions, solver_parameters=None):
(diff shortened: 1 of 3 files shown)


jwallwork23 (approved):
Good spot, thanks Stephan!

Issue #253, opened 2021-07-14
Deprecate `solve_tracer`


jwallwork23:
Following the discussion of PR #247, we decided to deprecate the `solve_tracer` option for `ModelOptions2d`. Tracer fields should now be specified by the user using `ModelOptions2d.add_tracer_2d`.

To facilitate the deprecation, we need to

- [x] Update all of the 2D tracer tests to use this format;
- [x] Update all of the 2D tracer demos to use this format;
- [x] Summarise changes to API and publicise;
- [x] Drop `ModelOptions2d.tracer_source_2d`;
- [x] Move `ModelOptions2d.horizontal_diffusivity` to `sediment_model_options`;
- [x] (Tangentially related) Allow for conservative/non-conservative tracer in different equations.

jwallwork23:
Done! See API changes [here](https://github.com/thetisproject/thetis/wiki/API-change-2021-08-06).

PR #258, opened 2021-08-05
Deprecate solve tracer

jwallwork23:
Addresses most of the points in #253, as well as updating the tracer and sediment model docs.

If these changes look okay then I will summarise the API changes in this PR and #251, put them in a page on the wiki and then publicise on Slack, etc.

### DIFF
--- a/thetis/coupled_timeintegrator_2d.py
+++ b/thetis/coupled_timeintegrator_2d.py
@@ -37,7 +37,7 @@ class CoupledTimeIntegrator2D(timeintegrator.TimeIntegratorBase):
         print_output('Coupled time integrator: {:}'.format(self.__class__.__name__))
         if not self.options.tracer_only:
             print_output('  Shallow Water time integrator: {:}'.format(self.swe_integrator.__name__))
-        if self.options.solve_tracer:
+        if self.options.tracer != {}:
             print_output('  Tracer time integrator: {:}'.format(self.tracer_integrator.__name__))
         if self.options.sediment_model_options.solve_suspended_sediment:
             print_output('  Sediment time integrator: {:}'.format(self.sediment_integrator.__name__))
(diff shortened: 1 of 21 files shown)


thetis/coupled_timeintegrator_2d.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, solver):
         print_output('Coupled time integrator: {:}'.format(self.__class__.__name__))
         if not self.options.tracer_only:
             print_output('  Shallow Water time integrator: {:}'.format(self.swe_integrator.__name__))
-        if self.options.solve_tracer:
+        if self.options.tracer != {}:
```
tkarna, line 40: I wonder if there's a more succinct way to do this check, it's repeated in a few places. Maybe in `Solver2d.create_equations` you could add a boolean `self.solve_tracers = self.options.tracer != {}` to have it easily available? (Caveat is that this bool would not exist before `create_equations` is called but it's probably fine if it's used only internally. An alternative is to add something to the options itself).

thetis/options.py
```diff
@@ -472,6 +472,8 @@ class TracerFieldOptions(FrozenHasTraits):
         None, allow_none=True, help='Source term for the tracer equation')
     diffusivity = FiredrakeScalarExpression(
         None, allow_none=True, help='Diffusion coefficient for the tracer equation')
+    conservative = Bool(
+        False, help='Should the tracer equation be solved in conservative form?')
```
tkarna, line 476: maybe use a more descriptive name e.g. `use_conservative_form`, similar to what we use for other boolean options

tkarna (approved):
Thanks, this is nice. Only a couple minor suggestions

tkarna (approved):
Thanks, looks good to me!

stephankramer (approved):
Looks perfect!

jwallwork23:
Thanks! Will write a wiki page shortly.

PR #259, opened 2021-08-17
Advection-diffusion-reaction

jwallwork23:
This PR allows arbitrary reaction terms to be included in the tracer model, as well as an example to show how it can be used.

It also allows for Picard iterations in the tracer model.


thetis/utility.py
```diff
@@ -683,7 +683,8 @@ def anisotropic_cell_size(mesh):
     op2.par_loop(kernel, P0_ten.node_set, evectors.dat(op2.RW), evalues.dat(op2.RW), B.dat(op2.READ))
 
     # Return minimum eigenvalue
-    return interpolate(evalues[-1], FunctionSpace(mesh, "DG", 0))
+    dim = mesh.topological_dimension()
```
stephankramer, line 686: Just curious: why did you hit this? Also the kernel above seems to be hard-coded to assume evalues has dim 2, so if the evalues vector function provided isn't dim 2 then I don't think that's safe. Would it not be simpler in any case to have the kernel just return the largest eigenvalue, since that's the only thing we're actually using?

thetis/tracer_eq_2d.py
```diff
@@ -268,6 +268,25 @@ def residual(self, solution, solution_old, fields, fields_old, bnd_conditions):
...
+    where :math:`R` is a user defined reaction term.
+    """
+    def residual(self, solution, solution_old, fields, fields_old, bnd_conditions):
+        f = 0
+        R = fields_old.get('reaction_terms')
+        if R != 0 and R is not None:
+            f += -inner(R, self.test)*self.dx
+        return -f
+
+
```
stephankramer, line 289: What is a reaction term? Not trying to be pedantic, I know within an application area it may have a well defined meaning, I'm just asking in terms of its functional meaning - which I guess is asking for mathematical rather than a physical definition. Functionally at the moment it looks like a reaction term is exactly the same as a source term. There's many source/right-hand side terms that may occur in applications and just in general I don't think we want to be adding new terms for every bit of physics if there's no functional difference wrt how Thetis handles them: the user can just provide a sum of these as the source.
Now you may say that a reaction term is a term that depends on other tracers but then we need some idea on how this coupling would work more generally - if we're just doing Picard iterations, i.e. we're treating the tracer equations one by one, then within the context of each tracer equation it's just a source term.

thetis/options.py
```diff
@@ -888,16 +899,23 @@ def add_tracer_2d(self, label, name, filename, shortname=None, unit='-', source=
         assert label not in self.tracer, f"Field '{label}' already exists."
         assert ' ' not in label, "Labels cannot contain spaces"
         assert ' ' not in filename, "Filenames cannot contain spaces"
+        if priority is not None:
+            assert isinstance(priority, int)
+            if priority in self.tracer_priorities:
```
stephankramer, line 904: Would it not be simpler to just ask the user to add the tracer in the order they want them solved rather than specifying a priority? Or is there a separate meaning to the order in `self.tracer` that we might want to be different from the solve order?

thetis/solver2d.py
```diff
@@ -592,6 +599,16 @@ def create_timestepper(self):
...
+            if self.options.tracer_priorities != {}:
+                reordered = OrderedDict()
+                priorities = list(self.options.tracer_priorities.keys())
+                priorities.sort()
+                for priority in priorities:
+                    label = self.options.tracer_priorities[priority]
+                    reordered[label] = self.options.tracer.pop(label)
+                for label in list(self.options.tracer.keys()):
+                    reordered[label] = self.options.tracer.pop(label)
+                self.options.tracer = reordered
```
stephankramer, line 611: Ah no (there's no separate ordering) - overwriting user options is never a good idea though: `add_tracer_2d` is a nice helper function that makes it easier to add tracer fields, but we should still treat the options dictionaries (FrozenHasTraits) as user provided input data that we're not allowed to clobber.

stephankramer (commented):
I think we need to think some more on the interaction with timesteppers. At the moment it only works for CrankNicolson but there it doesn't actually do crank nicolson if you have the user specify a reaction term with the tracer solution functions in them, as these solution functions contain the end-of-timestep value not the theta-weighted value that you'd need for CN, i.e. it is only correct for theta=1.

jwallwork23:
> I think we need to think some more on the interaction with timesteppers. At the moment it only works for CrankNicolson but there it doesn't actually do crank nicolson if you have the user specify a reaction term with the tracer solution functions in them, as these solution functions contain the end-of-timestep value not the theta-weighted value that you'd need for CN, i.e. it is only correct for theta=1.

That's true. Let's discuss at the meeting tomorrow.

test/examples/test_examples.py
```diff
@@ -34,6 +34,7 @@
     'lockExchange/diagnostics.py',
     'lockExchange/plotting.py',
     'lockExchange/submitRuns.py',
+    'reaction/gray_scott.py',
```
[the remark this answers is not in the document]
jwallwork23: Good point, I had misremembered that it was opt-in, rather than opt-out.

thetis/utility.py
jwallwork23: I think it was a UFL update (unrelated to this PR). But, yes, it would better to update as you suggest.

thetis/tracer_eq_2d.py
jwallwork23: I agree that it is identical to `SourceTerm`, except that the name of the field is different and it gets added as 'implicit', rather than 'source', at the end. I think the same can be achieved with `SourceTerm`, with a couple of tweaks.

thetis/solver2d.py
jwallwork23: Yeah good point. I was mainly thinking of the (not yet supported) case where you might want to solve a subset of the tracer equations as a mixed system. I'll drop this `priority` business, for now.

jwallwork23:
Oh I accidentally reversed some changes - will come back to this.

jwallwork23:
Accidental changes now fixed.

examples/reaction/gray_scott.py
```diff
@@ -0,0 +1,107 @@
...
+# Spin up timestep
+dt = 1.0e-04
+end_time = 0.0
+for i in range(4):
+    dt *= 10
+    end_time += 10*dt if i == 0 else 9*dt
+    options.timestep = dt
+    solver_obj.export_initial_state = i == 0
+    solver_obj.timestepper.timesteppers["a_2d"].set_dt(dt)
+    solver_obj.timestepper.timesteppers["b_2d"].set_dt(dt)
```
tkarna, line 98: The `set_dt` calls have no effect as you are re-creating the time steppers below? Maybe (in a future PR) we could add a mechanism to change dt. Should be OK for all Runge-Kutta type integrators.

tkarna (dismissed):
Thanks for this, looks good to me!

examples/reaction/gray_scott.py
jwallwork23: Good point. I'll make an issue for this.

PR #263, opened 2021-09-02
Remove reference to firedrake.Expression

wence-:
We've long deprecated `Expression` objects, https://github.com/firedrakeproject/firedrake/pull/2202 finally removes them. Cull the only reference in thetis to them.


tkarna (approved):
Thanks!

PR #265, opened 2021-09-03
Test notebooks

jwallwork23:
Recent experience with Firedrake notebooks bit rotting suggests it is a good idea to test them.


stephankramer (approved):
Great, useful indeed. All looks good

PR #267, opened 2021-09-10
YAML specification of diffusion-reaction models

LawrenceDior:
`read_tracer_from_yml` reads tracer diffusivities and reaction terms from a .yml file and returns an `OrderedDict` of Firedrake and UFL objects representing this information.

The `ADR_Model` class is responsible for model validation and the parsing of reaction terms.

The `load_tracers_2d` method of the `FlowSolver2d` class calls `read_tracer_from_yml` and adds the corresponding tracers to the `FlowSolver2d` object.

`examples/reaction/gray_scott_io.py` contains a working example of a diffusion-reaction model loaded from disk.

See `reaction_models/gray-scott.yml` for an example of a .yml file in the expected format.


thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+import networkx as nx
+import sympy
+from sympy.parsing.sympy_parser import \
+    convert_xor, \
+    parse_expr, \
+    standard_transformations
+from sympy.utilities.lambdify import lambdify
+
+
+class ADR_Model:
```
LawrenceDior, line 11: This class was originally created for my MSc project. As such, it may need some modification to integrate better with Thetis. I have already adjusted the format of the docstrings, but the functionality is mostly untouched.

examples/reaction/gray_scott_io.py
```diff
@@ -0,0 +1,89 @@
...
+experience different diffusivities and react with one
+another. The problem is interesting from a numerical
+modelling perspective because of the nonlinear coupling
+between the two equations in the system. Instead of
+solving the equations as a monolithic system, we opt to
+solve them alternately using a Picard iteration.
+
+[Hundsdorf & Vermer 2003] W. Hundsdorf & J.G. Verwer
+    (2003). "Numerical Solution of Time-Dependent
+    Advection-Diffusion-Reaction Equations", Springer.
```
jwallwork23, line 19: Probably best to refer to the original demo version for this and just say how this particular script differs from it.

examples/reaction/gray_scott_io.py
```diff
@@ -0,0 +1,89 @@
...
+sim_end_time = 2000.0
+
+# Specify iterative method
+options.tracer_picard_iterations = 2
+options.set_timestepper_type('CrankNicolson', implicitness_theta=1.0)
+
+# Load tracers
+solver_obj.create_function_spaces()
+solver_obj.load_tracers_2d(
+    "gray-scott", input_directory="reaction_models",
```
jwallwork23, line 48: If I run locally, it can't find the yaml file. Maybe move it to `examples/reaction`?

thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+
+    # convert_xor causes "x^2" to be parsed as "x**2"
+    transformations = (
+        standard_transformations + (convert_xor,))
+
+    reserved_symbols = {
+        "pi", "e",
+        "x", "y", "z", "t",
+        "u", "u_x", "u_y", "u_z"}
```
jwallwork23, line 38: I guess a discussion point is whether we want all of these (and any more) as reserved symbols. I can imagine pi, e, x, y, t being useful for if you have a source term that depends on them. Probably not z in a 2D model.

requirements.txt
```diff
@@ -3,3 +3,6 @@ netCDF4
 pyproj
 uptide
 pytz
+networkx
+pyyaml
+sympy
```
jwallwork23, line 8: At the minute the implementation converts yml->dict->sympy->UFL. Maybe it is possible to do yml->dict->UFL? If so, it's probably best to leave for future work.

thetis/adr_model.py
LawrenceDior, line 38: These symbols are reserved to allow reaction terms to contain pi, e, spatial coordinates, time, the flow field, and individual components of the flow field.

I'm not aware of any issues with pi or e, because these are just constants. However, at present, `parse_tracers_from_dict` will not work properly if it encounters a reaction term containing any of these other symbols. It expects all symbols to correspond to either constants or species.

The simplest way to solve this problem would be to remove everything except pi and e from `reserved_symbols`, but it might be worth considering alternatives it if people would like to be able to include spatial coordinates, etc. in reaction terms.

thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+        return expression
+
+    def species_name(self, k):
+        return self.species[k]["name"]
+
+    def diffusion(self, k):
+        return self.species[k]["diffusion"]
+
+    def reaction_term(self, k):
+        return self.species[k]["reaction"]["expression"]
```
LawrenceDior, line 188: I don't think I've used this anywhere

jwallwork23: I guess it can be dropped then?

thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def reaction_function(self, k):
+        return self.species[k]["reaction"]["function"]
+
+    def list_species_keys(self):
+        return sorted(
+            list(self.species_keys),
+            key=lambda k: self.species[k]['index'])
+
+    def dependency_graph(self):
```
LawrenceDior, line 201: This method isn't used anywhere, and it's the only reason the NetworkX import is there. I've kept it in to allow for discussion in this PR, in case developers decide to make use of it somehow.

thetis/utility.py
```diff
@@ -1004,3 +1007,162 @@ def solve(self, solve_w=True):
...
+    # Load dictionary
+    model_dict = None
+    with open(filename, 'r') as f_in:
+        model_dict = yaml.safe_load(f_in)
+
+    key_order = None
+    if preserve_order:
+        # Ensure that parse_tracers_from_dict adds tracers in the same order
+        # as the corresponding species appear in the input file.
+        key_order = get_species_key_order(filename)
```
LawrenceDior, line 1084: I believe this part isn't strictly needed, because from Python 3.6 onwards an ordinary `dict` should preserve order by default. But I could be wrong.

examples/reaction/gray_scott_io.py
LawrenceDior: Done.

thetis/adr_model.py
LawrenceDior: Agreed -- see separate comment. For z, I suppose the question is whether or not this class might also be used in tandem with 3D solvers.

examples/reaction/gray_scott_io.py
LawrenceDior: Updated.

thetis/adr_model.py
LawrenceDior: Probably, yes. I'll leave it in for now to give other developers a chance to see it, as with `dependency_graph`.

thetis/utility.py
```diff
@@ -1004,3 +1007,162 @@ def solve(self, solve_w=True):
...
+    key_suffix = '_%id' % function_space.mesh().topological_dimension() \
+                 if append_dimension else ''
+    mdl = ADR_Model(model_dict, lambdify_modules=lambdify_modules)
+
+    species = key_order
+    if species is None:
+        # Get tracer labels
+        species = mdl.list_species_keys()
+
+    adr_model = OrderedDict({s + key_suffix: {} for s in species})
```
LawrenceDior, line 1155: Another point for discussion: should the `ADR_Model` instance be used internally to set up an `OrderedDict` as it is now, or should the `ADR_Model` class be updated to allow it to be used instead of an `OrderedDict`? (e.g. by passing a `FunctionSpace` to the constructor and having it set up `Functions`, `Constant`s and UFL expressions internally)

jwallwork23:
Note: this could also be used for Thetis' sediment model.

thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+        self.constant_keys = set(model_dict["constants"].keys())
+        self.species_keys = set(model_dict["species"].keys())
+        self._validate_keys()
+        for k, v in model_dict["constants"].items():
+            self.constants[k] = float(v)
+        for i, (k, v) in enumerate(model_dict["species"].items()):
+            # Parse reaction and diffusion terms
+            reaction_term = parse_expr(
+                str(v["reaction"]),
+                transformations=ADR_Model.transformations)
```
jwallwork23, line 102: Can this not be `self.transformations`?

thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+            diffusion_term = parse_expr(
+                str(v["diffusion"]),
```
jwallwork23, line 104: Maybe better to check this is a string above, rather than converting it? Or is there a case where you would actually want to convert from a different type?

thetis/adr_model.py
jwallwork23: I think let's drop it if it's unused.

thetis/adr_model.py
jwallwork23: I think it's useful functionality. Presumably pip installing networkx is pretty quick?

thetis/adr_model.py
```diff
@@ -0,0 +1,220 @@
...
+            # Obtain a list of arguments to be passed to the
+            # reaction function
+            reaction_args = list(reaction_term.free_symbols)
+            # Store species-specific data.
+            # The reaction expression will now be converted into
+            # a callable function.
+            self.species[k] = {
+                "name": v["name"],
+                "index": i,
+                "diffusion": float(diffusion_term),
```
jwallwork23, line 122: So at the minute we are only supporting constant diffusion terms, which is fine. To support spatially varying diffusivity and coefficients, I guess we would need the functionality to load from file or something. Probably best to just leave for now until someone needs to be able to do it.

thetis/adr_model.py
jwallwork23: Please could you use this in the new version of the Gray-Scott example?

thetis/utility.py
jwallwork23: Firedrake requires Python 3.6 so I think we can drop `preserve_order`.

jwallwork23: Presumably that also means we can drop `get_species_key_order`?

thetis/utility.py
jwallwork23: I think it would be neater if `ADR_Model` had an `OrderedDict`-like structure itself. Would that be feasible?

joewallwork (changes requested):
Thanks, Lawrence! This is great. A few comments and requests. e.g. it'd be nice if you could demonstrate the networkx usage in the Gray-Scott example.

Sorry it took a while to get round to reviewing.

jwallwork23:
With @LawrenceDior's permission, I made the requested changes.

PR #268, opened 2021-09-13
Monolithic tracer systems

jwallwork23:
The main contribution of this PR is to allow tracers to be added and then solved as mixed systems. All the user needs to do is to call `add_tracers_2d`, rather than `add_tracer_2d`. This is achieved by introducing a mixed time integrator class and allowing trial and test functions to be passed through to the tracer equations.

Other changes in this PR:
* Allow control of the degree of the `Q_2d` space.
* Tracer function space is inherited from the `function` kwarg to `add_tracer_2d`/`add_tracers_2d`, otherwise  `Q_2d` is used by default. Maybe the tracer function space and discretisation parameters should be controlled by `add_tracer_2d`/`add_tracers_2d`, to ensure consistency?

### DIFF
--- a/examples/reaction/gray_scott_mixed.py
+++ b/examples/reaction/gray_scott_mixed.py
@@ -1,0 +1,95 @@
+"""
+In this version of the Gray-Scott demo, we solve the
+two equations as a monolithic system.
+
+The main difference is the use of
+`ModelOptions2d.add_tracers_2d`, rather than
+`ModelOptions2d.add_tracer_2d`.
+"""
+from thetis import *
+
+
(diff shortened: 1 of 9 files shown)


jwallwork23:
Note from @stephankramer: we should make sure the solver parameters are sensible for the mixed system.

jwallwork23:
Suggestion by @tkarna: use a coupled/mixed *equation* class, rather than timestepper.

jwallwork23:
I'll redo this using Tuomas' suggestion and open a new PR later.

Issue #269, opened 2021-09-13
Can we remove pylit?


jwallwork23:
I wonder if we can remove pylit and cut down on 45k lines of code? A version is already maintained in Firedrake and can be accessed via `$VIRTUAL_ENV/src/firedrake/pylit/pylit.py`. Maybe I am misunderstanding something. [This branch](https://github.com/thetisproject/thetis/tree/remove-pylit) would remove Thetis' pylit and use the Firedrake version to build docs.

Here is the `diff -r` between the pylit in Thetis and Firedrake: 
```
Only in thetis/pylit/doc: Makefile
Only in thetis/pylit/doc: build
Only in thetis/pylit/doc/examples: listings-python-options.sty.txt
Only in thetis/pylit/doc/examples: listings-tex-options.sty.txt
Only in thetis/pylit/doc/examples: palatino-helvetica-courier.sty.txt
diff -r thetis/pylit/doc/examples/pylit_test.py.txt firedrake/pylit/doc/examples/pylit_test.py.txt
275,278c275,278
<           directives = ['.. code-block:: python',
<                        '  .. code-block:: python',
<                        '.. code-block:: python listings',
<                        '  .. code-block:: python listings']
---
>           directives = ['.. code-block:: python3',
>                        '  .. code-block:: python3',
>                        '.. code-block:: python3 listings',
>                        '  .. code-block:: python3 listings']
789c789
<           data = ["# .. code-block:: python\n",
---
>           data = ["# .. code-block:: python3\n",
794c794
<           soll = ['.. code-block:: python\n',
---
>           soll = ['.. code-block:: python3\n',
Only in firedrake/pylit/doc/logo: pyLit-bold-framed.svg.bak
Only in thetis/pylit/doc/logo: pylit-bold-framed.png
diff -r thetis/pylit/doc/usage.txt firedrake/pylit/doc/usage.txt
87c87
<   pylit.defaults.code_block_marker = '.. code-block:: python'
---
>   pylit.defaults.code_block_marker = '.. code-block:: python3'
diff -r thetis/pylit/pylit.py firedrake/pylit/pylit.py
1c1
< #!/usr/bin/env python
---
> #!/usr/bin/env python3
diff -r thetis/pylit/test/pylit_test.py firedrake/pylit/test/pylit_test.py
307,310c307,310
<         directives = ['.. code-block:: python',
<                      '  .. code-block:: python',
<                      '.. code-block:: python listings',
<                      '  .. code-block:: python listings']
---
[...]

PR #270, opened 2021-09-14
Remove pylit

jwallwork23:
See issue #269.

Also includes a minor doc fix to reduce sphinx warnings.


tkarna (approved):
Thanks, looks good to me

PR #272, opened 2021-09-20
Fix numpy import

pbrubeck:
Firedrake was mistakenly importing numpy as np. This PR ensures that numpy is imported properly.


tkarna:
Thanks, numpy is used quite extensively in Thetis, both in the model and examples, so this will hit quite a few files. We'll need to figure out the right way to fix this.

tkarna:
Here's a list of affected files:
```
> grep -rl "np\." thetis/
thetis/solver.py
thetis/solver2d.py
thetis/timeintegrator.py
thetis/optimisation.py
thetis/coordsys.py
thetis/forcing.py
thetis/turbulence.py
thetis/callback.py
thetis/interpolation.py
thetis/limiter.py
thetis/rungekutta.py
thetis/stability_functions.py
thetis/turbines.py
thetis/utility3d.py
thetis/utility.py
thetis/coupled_timeintegrator.py
```
and more under `examples` and `tests`.

jwallwork23 (approved):
Looks good to me (assuming tests pass). Thanks!

I think it is worth adding an entry on the [API change wiki page](https://github.com/thetisproject/thetis/wiki/API-changes) and mentioning on Slack once merged.

jwallwork23:
Oh they passed while I was reviewing, nice.

tkarna:
> I think it is worth adding an entry on the [API change wiki page](https://github.com/thetisproject/thetis/wiki/API-changes) and mentioning on Slack once merged.

Yes, that's a good idea. I'll do that.

stephankramer (approved):
Looks good to me. No real preference for `numpy` vs `import numpy as np` (maybe slightly prefer the first). Personally I'm not a fan of having `import numpy`  (and similarly `import os` and `import datetime`) in `thetis/__init__.py` and would prefer just an explicit `import numpy` at the top of each file that uses it. I just feel it's not something that should come (or the user would expect to come) with `from thetis import *`

tkarna:
> Personally I'm not a fan of having import numpy (and similarly import os and import datetime) in thetis/__init__.py

Yes, I agree. We should remove those and also remove star imports (at least not use them within the Thetis package). We can fix those in another PR.

PR #273, opened 2021-10-12
Tweaks for 2D tracer

joewallwork:
A few little things:
* The outflow condition in `tracerEq/test_point_discharge.py` should be a natural BC, which you get by default.
* Would be nice to plot tracer stats, as well as SWE ones.
* Alternative method for computing anisotropic cell size, which avoids computing the SVD.


stephankramer:
Attempted fix in #275

PR #275, opened 2021-10-12
Fix JITModule error that breaks all CI tests

stephankramer:
JITModule seems to have moved from pyop2.base to pyop2.parloop


jwallwork23 (approved):
Looks good, assuming tests pass.

Thanks, @stephankramer!

PR #276, opened 2021-10-15
Event logging

jwallwork23:
I applied PETSc event logging throughout Thetis and found it really useful for profiling Thetis-based codes.

One thing I noticed was that a decent chunk (of an admittedly small application) was spent computing horizontal element sizes and facet areas. This is because they were using the default solver parameters (full LU as preconditioner). To avoid this, I've used Firedrake's `MassInvPC` for the former and an application of Jacobi for the latter.

### DIFF
--- a/thetis/solver.py
+++ b/thetis/solver.py
@@ -77,6 +77,7 @@ class FlowSolver(FrozenClass):
 
     See the manual for more complex examples.
     """
+    @PETSc.Log.EventDecorator("thetis.FlowSolver.__init__")
     def __init__(self, mesh2d, bathymetry_2d, n_layers,
                  options=None, extrude_options=None):
         """
@@ -187,6 +188,7 @@ class FlowSolver(FrozenClass):
         factor = 0.5*0.25/l_r
         return factor
 
+    @PETSc.Log.EventDecorator("thetis.FlowSolver.compute_dt_2d")
     def compute_dt_2d(self, u_scale):
         r"""
         Computes maximum explicit time step from CFL condition.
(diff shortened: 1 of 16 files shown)


tkarna (approved):
Thanks, this is great!

tkarna:
Do these events just show up in PETSc log_view?

jwallwork23:
> Do these events just show up in PETSc log_view?

Yes. If you print the log_view to a text file then you can generate a flame graph from it, as described here: https://firedrakeproject.org/optimising.html#automatic-flame-graph-generation-with-petsc.

thetis/solver.py
```diff
@@ -217,10 +217,8 @@ def compute_dt_2d(self, u_scale):
         l = inner(test, csize / u) * dx
         sp = {
             "snes_type": "ksponly",
-            "mat_type": "matfree",
-            "ksp_type": "preonly",
-            "pc_type": "python",
-            "pc_python_type": "firedrake.MassInvPC",
+            "ksp_type": "gmres",
+            "pc_type": "ilu",
```
wence-, line 221: `ksp_type` should be `cg` (the operator is SPD), and you probably want `-pc_type bjacobi -sub_pc_type ilu` to work right in parallel.

thetis/utility.py
```diff
@@ -540,7 +541,12 @@ def get_horizontal_elem_size_2d(sol2d):
     tri = TrialFunction(p1_2d)
     a = inner(test, tri) * dx
     l = inner(test, sqrt(CellVolume(mesh))) * dx
-    solve(a == l, sol2d)
+    sp = {
+        "snes_type": "ksponly",
+        "ksp_type": "gmres",
+        "pc_type": "ilu",
```
wence-, line 547: Same idea here.

PR #278, opened 2021-11-26
Fix error using ILU in parallel

jwallwork23:
A bug was introduced in #276 and reported by @jhill1. Fix suggested by @wence-.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,3 +1,4 @@
+from subprocess import check_call
 
 
 def pytest_runtest_teardown(item, nextitem):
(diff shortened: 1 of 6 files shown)


stephankramer (dismissed):
Makes sense

wence-:
Is that all of them? 4a3d90d0845 seemed to change solver parameters in four places, this only does it in two.

jwallwork23:
Good point. Okay that should do it. `grep -r ilu` only returns `'sub_pc_type': 'ilu'` type entries now.

jwallwork23:
Following discussion in today's meeting, I copied over the infrastructure for running tests in parallel from Firedrake and applied it to some tests. They ran fine and passed for me locally, hopefully they work on the CI, too.

tkarna (approved):
Thanks! Looks good to me

jhill1:
Pulled this branch and can confirm this works on the local HPC system.

PR #277, opened 2021-11-12
Simple channel inversion modeling example

tkarna:
Simple 2D channel test case to demonstrate Thetis adjoint capabilities. Forward run is used to store elevation time series at some locations. Inverse model can then be used to find the original model fields, i.e. bathymetry, bottom friction coefficient, and initial elevation field.

PR #279, opened 2021-11-26
HessianRecoverer2D

jwallwork23:
Building upon PR #242, maybe this will be useful for other people (e.g. in #277).

To use it in practice, do something like
```
from thetis.recovery import HessianRecoverer2D

P1_ten = TensorFunctionSpace(mesh2d, 'CG', 1)
hessian_2d = Function(P1_ten)
hessian_calculator = HessianRecoverer2D(hessian_2d, solver_obj, 'elev_2d')
solver_obj.add_new_field(hessian_2d, 'hessian_2d', 'Hessian of elevation', 'Hessian2d', preproc_func=hessian_calculator.solve)
solver_obj.options.fields_to_export = ['hessian_2d']
```

### DIFF
--- a/test/operations/test_diagnostics.py
+++ b/test/operations/test_diagnostics.py
@@ -46,3 +46,33 @@ def test_hessian_recovery2d(interp, tol=1.0e-08):
     assert err_g < tol, f'Gradient approximation error {err_g:.4e}'
     err_H = errornorm(H, H_expect)/norm(H_expect)
     assert err_H < tol, f'Hessian approximation error {err_H:.4e}'
+
+
+def test_vorticity_calculation2d(interp, tol=1.0e-08):
+    r"""
+    Calculate the vorticity of the velocity field
+    :math:`\mathbf u(x, y) = \frac12(y, -x)` and check
+    that the result is unity.
+
+    :arg interp: should the velocity be interpolated as
+        as a :class:`Function`, or left as a UFL expression?
+    :kwarg tol: relative tolerance for value checking
+    """
+    mesh2d = UnitSquareMesh(4, 4)
+    x, y = SpatialCoordinate(mesh2d)
+    uv = 0.5*as_vector([y, -x])
+    if interp:
+        P1v_2d = get_functionspace(mesh2d, 'CG', 1, vector=True)
+        uv = interpolate(uv, P1v_2d)
+    P1_2d = get_functionspace(mesh2d, 'CG', 1)
+
+    # Recover vorticity
+    omega = Function(P1_2d, name='Vorticity')
+    vorticity_calculator = VorticityCalculator2D(uv, omega)
+    vorticity_calculator.solve()
+
+    # Check values
+    omega_expect = Function(P1_2d, name='Expected vorticity')
+    omega_expect.assign(1.0)
+    err = errornorm(omega, omega_expect)/norm(omega_expect)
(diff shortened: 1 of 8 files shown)


thetis/recovery.py
```diff
@@ -0,0 +1,125 @@
...
+    Hessians are recoved using double :math:`L^2`
+    projection, which is implemented using a
+    mixed finite element method.
+
+    :arg hessian_2d: :class:`Function` to hold recovered Hessian.
+    :arg solver_obj: :class:`FlowSolver2d` object.
+    :kwargs: to be passed to the :class:`LinearVariationalSolver`.
+    """
+    @PETSc.Log.EventDecorator("thetis.HessianRecoverer2D.__init__")
+    def __init__(self, hessian_2d, solver_obj, field, **kwargs):
```
tkarna, line 59: I would prefer that these routines do not depend on `solver_obj` (applies to VorticityCalculator2D as well). After all, you only use it to access `solver_obj.fields[field]` so you could just pass in the field itself. This implies a little more boilerplate code by the user but you already need some, e.g. to define the TensorFunctionSpace. If this is too much we could add a helper function somewhere.

thetis/recovery.py
```diff
@@ -0,0 +1,125 @@
...
+            raise ValueError(f'Dimension {dim} not supported')
+        if element_continuity(fs.ufl_element()).horizontal != 'cg':
+            raise NotImplementedError
+        if fs.dof_dset.dim != (2, 2):
+            raise ValueError('Expecting a square tensor function')
+        n = FacetNormal(mesh)
+
+        # Setup function spaces
+        Sigma = hessian_2d.function_space()
+        V = solver_obj.function_spaces.P1v_2d
```
tkarna, line 77: You could call `thetis.get_functionspace` instead

tkarna (commented):
Thanks, this is great! I wonder whether "recovery" is the right term to use here? These two routines calculate some diagnostic fields, so the py file could be "diagnostics", "diagnostic_fields", or "postprocessing" or such? Gradient and Hessian could go under gradient estimation umbrella, but e.g. vorticity maybe not.

jwallwork23:
> Thanks, this is great! I wonder whether "recovery" is the right term to use here? These two routines calculate some diagnostic fields, so the py file could be "diagnostics", "diagnostic_fields", or "postprocessing" or such? Gradient and Hessian could go under gradient estimation umbrella, but e.g. vorticity maybe not.

Good point. The vorticity field involves derivatives to be recovered, but I agree "diagnostics" would be a better name.

thetis/recovery.py
jwallwork23: Sure, I can do that.

jwallwork23:
I reworked both ``VorticityCalculator2D`` and ``HessianRecoverer2D`` so that they do not depend on the solver object. I also added them to tests, to check that they work and provide example usage.

Here is the magnitude of the elevation Hessian for ``swe2d/test_anisotropic.py``
hessian_elev

tkarna:
Should we also include tests against analytical solutions to check that the vorticity/hessian are computed correctly? Could go under `test/operations/`.

jwallwork23:
> Should we also include tests against analytical solutions to check that the vorticity/hessian are computed correctly? Could go under `test/operations/`.

Sure, I'll take a look at that later in the week.

tkarna (approved):
Thanks for the updates, looks good to me!

PR #281, opened 2021-12-03
Extend 2D forcing routines to work on the sphere

tkarna:
- ATMInterpolator and TidalBoundary work on the sphere
- TimeSeriesCallback works on the sphere with 3D coordinate vectors
- Adds Large and Yeager (2009) wind stress formulation

### DIFF
--- a/thetis/callback.py
+++ b/thetis/callback.py
@@ -601,9 +601,10 @@ class TimeSeriesCallback2D(DiagnosticCallback):
 
     @PETSc.Log.EventDecorator("thetis.TimeSeriesCallback2D.__init__")
     def __init__(self, solver_obj, fieldnames, x, y,
-                 location_name,
+                 location_name, z=None,
                  outputdir=None, export_to_hdf5=True,
-                 append_to_log=True):
+                 append_to_log=True,
+                 tolerance=1e-3):
         """
         :arg solver_obj: Thetis :class:`FlowSolver` object
         :arg fieldnames: List of fields to extract
@@ -623,6 +624,10 @@ class TimeSeriesCallback2D(DiagnosticCallback):
         self.location_name = location_name
         attrs = {'x': x, 'y': y}
         attrs['location_name'] = self.location_name
+        self.on_sphere = solver_obj.mesh2d.geometric_dimension() == 3
+        if self.on_sphere:
+            assert z is not None, 'z coordinate must be defined on a manifold mesh'
+            attrs['z'] = z
         field_short_names = [f.split('_')[0] for f in self.fieldnames]
         field_str = '-'.join(field_short_names)
         self.variable_names = field_short_names
@@ -637,6 +642,8 @@ class TimeSeriesCallback2D(DiagnosticCallback):
             append_to_log=append_to_log)
         self.x = x
         self.y = y
+        self.z = z
+        self.tolerance = tolerance
         self._initialized = False
 
     @PETSc.Log.EventDecorator("thetis.TimeSeriesCallback2D._initialize")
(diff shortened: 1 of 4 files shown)


thetis/callback.py
```diff
@@ -646,15 +654,22 @@ def _initialize(self):
...
 
+        # construct mesh points
+        xyz = (self.x, self.y, self.z) if self.on_sphere else (self.x, self.y)
+        self.xyz = numpy.array([xyz])
+
         # test evaluation
         try:
-            self.solver_obj.fields.bathymetry_2d.at((self.x, self.y))
+            if self.eval_func is None:
+                self.solver_obj.fields.bathymetry_2d.at(self.xyz, tolerance=self.tolerance)
```
jwallwork23, line 664: It would be good to replace the `at` calls in the callbacks with the `VertexOnlyMesh` approach, so that you keep the interpolator (AFAIK `at` doesn't do that). It's on my to do list.

jwallwork23: Although I don't think `VertexOnlyMesh` currently accepts a tolerance, so we might need that first.

thetis/forcing.py
```diff
@@ -36,18 +37,23 @@ def compute_wind_stress(wind_u, wind_v, method='LargePond1981'):
         Wind stress formulation by [1]
     - "SmithBanke1975":
         Wind stress formulation by [2]
+    - "LargeYeager2009":
+        Wind stress formulation by [2]
```
jwallwork23, line 41: I think it should be [3].

thetis/forcing.py
```diff
@@ -67,30 +80,35 @@ def compute_wind_stress(wind_u, wind_v, method='LargePond1981'):
...
+    A TimeParser class for reading atmosphere model output files.
     """
-    def __init__(self, filename, max_duration=24.*3600., verbose=False):
+    def __init__(self, filename, max_duration=None, verbose=False):
         """
         :arg filename:
-        :kwarg max_duration: Time span to read from each file (in secords,
-            default one day). Forecast files are usually daily files that
-            contain forecast for > 1 days.
+        :kwarg max_duration: Time span to read from each file (in secords).
```
jwallwork23, line 88: "secords"->"seconds"

thetis/forcing.py
```diff
@@ -644,6 +682,163 @@ def set_fields(self, time):
...
+
+        grid_lat = lat_subset.ravel()
+        grid_lon = lon_subset.ravel()
+        if numpy.ma.isMaskedArray(grid_lat):
+            grid_lat = grid_lat.filled(0.0)
+        if numpy.ma.isMaskedArray(grid_lon):
+            grid_lon = grid_lon.filled(0.0)
+        grid_latlon = numpy.vstack((grid_lat, grid_lon)).T
+
+        # building 2D interpolator, this can take a long time (minutes)
```
jwallwork23, line 760: Might be worth adding a PETSc event, given that it is a part of the program with non-negligible cost.

jwallwork23 (changes requested):
I'm not familiar with the forcing code, but the changes seem reasonable to me. A few comments and typos.

thetis/callback.py
tkarna: Yes, I was planning to do that in another PR. If we can cope without a tolerance that would be idea of course but there might be some issues with points close to the boundary.

thetis/forcing.py
tkarna: Good idea. I've now added a bunch of events in forcing/interpolation routines.

jwallwork23 (approved):
Thanks Tuomas, looks good to me!

thetis/callback.py
```diff
@@ -602,9 +602,10 @@ class TimeSeriesCallback2D(DiagnosticCallback):
 
     @PETSc.Log.EventDecorator("thetis.TimeSeriesCallback2D.__init__")
     def __init__(self, solver_obj, fieldnames, x, y,
-                 location_name,
+                 location_name, z=None,
```
stephankramer, line 605: Just a general comment unrelated to this PR. Why do we have both a `TimeSeriesCallback2D` and a `DetectorsCallback`? They seem to be very similar and we might be duplicating effort, such as the changes here...

thangel: This is a good point - `DetectorsCallback` predates the `TimeSeriesCallback2d`. I guess we could keep the latter as it is more abstract, though `DetectorsCallback` is a bit more intuitive for new users in my view ...

tkarna: Yes, I think the design difference is that `DetectorsCallback` stores all the time series in a single hdf5 file while `TimeSeriesCallback` creates separate files. In the past I've found it easier to work with separate files in post processing. But I think we'll need to revisit this issue if/when we generalize the output scheduling, metadata (e.g. storing time as calendar dates) and formats (e.g. produce standard netCDF time series files).

stephankramer (approved):
All looks good to me

jwallwork23 (approved):

PR #285, opened 2021-12-07
KineticEnergyCalculator

jwallwork23:
Introduce a diagnostic for computing dynamic pressure in 2D or 3D.

Also add some input checking for ``VorticityCalculator2D`` and ``HessianRecoverer2D``.

### DIFF
--- a/thetis/diagnostics.py
+++ b/thetis/diagnostics.py
@@ -24,17 +24,29 @@ class DiagnosticCalculator(FrozenHasTraits):
 
 class VorticityCalculator2D(DiagnosticCalculator):
     r"""
-    Linear solver for recovering fluid vorticity.
+    Linear solver for recovering fluid vorticity,
+    interpreted as a scalar field:
+
+    .. math::
+        \omega = -v_x + u_y,
+
+    for a velocity field :math:`\mathbf{u} = (u, v)`.
 
     It is recommended that the vorticity is sought
     in :math:`\mathbb P1` space.
-
-    :arg uv_2d: horizontal velocity :class:`Function`.
-    :arg vorticity_2d: :class:`Function` to hold calculated vorticity.
-    :kwargs: to be passed to the :class:`LinearVariationalSolver`.
     """
+    uv_2d = FiredrakeVectorExpression(
+        Constant(as_vector([0.0, 0.0])), help='Horizontal velocity').tag(config=True)
+
+    @unfrozen
     @PETSc.Log.EventDecorator("thetis.VorticityCalculator2D.__init__")
     def __init__(self, uv_2d, vorticity_2d, **kwargs):
+        """
+        :arg uv_2d: vector expression for the horizontal velocity.
+        :arg vorticity_2d: :class:`Function` to hold calculated vorticity.
+        :kwargs: to be passed to the :class:`LinearVariationalSolver`.
+        """
+        self.uv_2d = uv_2d
         fs = vorticity_2d.function_space()
         dim = fs.mesh().topological_dimension()
         if dim != 2:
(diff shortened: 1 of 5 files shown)


thetis/diagnostics.py
```diff
@@ -2,24 +2,30 @@
...
 
     It is recommended that the vorticity is sought
     in :math:`\mathbb P1` space.
-
-    :arg uv_2d: horizontal velocity :class:`Function`.
-    :arg vorticity_2d: :class:`Function` to hold calculated vorticity.
-    :kwargs: to be passed to the :class:`LinearVariationalSolver`.
     """
+    uv_2d = FiredrakeVectorExpression(
+        Constant(as_vector([0.0, 0.0])), help='Horizontal velocity').tag(config=True)
```
tkarna, line 19: Using traits here is nice as we can check the inputs dynamically. Having to define a default value is however a bit odd though but I guess we cannot avoid that.

thetis/diagnostics.py
```diff
@@ -140,10 +154,36 @@ def __init__(self, field_2d, hessian_2d, gradient_2d=None, **kwargs):
...
 
     @PETSc.Log.EventDecorator("thetis.HessianRecoverer2D.solve")
     def solve(self):
         self.solver.solve()
         self.hessian_2d.assign(self._hessian)
         if self.gradient_2d is not None:
             self.gradient_2d.assign(self._gradient)
+
+
+class DynamicPressureCalculator(FrozenHasTraits):
```
tkarna, line 167: I would prefer to call this `KineticEnergyCalculator` as I think that's more commonly used term.

tkarna, line 167: Note that there's already is a version implemented as a `DiagnosticCallback` https://github.com/thetisproject/thetis/blob/master/examples/freshwaterCylinder/freshwaterCylinder.py#L124
It could use the Calculator internally.

thetis/diagnostics.py
```diff
@@ -140,10 +154,36 @@ def __init__(self, field_2d, hessian_2d, gradient_2d=None, **kwargs):
...
+    r"""
+    Class for calculating dynamic pressure (i.e. kinetic energy).
+    """
+    density = FiredrakeScalarExpression(
+        Constant(1030.0), help='Fluid density').tag(config=True)
+
+    @PETSc.Log.EventDecorator("thetis.DynamicPressureCalculator.__init__")
+    def __init__(self, uv, dp, density=Constant(1030.0)):
```
tkarna, line 175: I guess by default we should use `physical_constants['rho0']` as the density value — it is the value the 2D solver uses internally. In the 3D baroclinic case the user would need to plug in the dynamical density expression `rho_0 + solver_obj.fields.density_3d` but I guess we don't need to specify it in this routine.

tkarna (changes requested):
Thanks, couple of minor comments

thetis/diagnostics.py
jwallwork23: Good idea. That would also mean we test the new calculator.

jwallwork23: Renamed as requested and added a base class.

thetis/diagnostics.py
```diff
@@ -141,9 +153,42 @@ def __init__(self, field_2d, hessian_2d, gradient_2d=None, **kwargs):
...
+        """
+        if density is not None:
+            self.density = density
+        self.ke = ke
+        u_sq = uv[0]*uv[0] + uv[1]*uv[1] if horizontal else dot(uv, uv)
+        self.ke_expr = 0.5*self.density*u_sq
+
+    @PETSc.Log.EventDecorator("thetis.KineticEnergyCalculator.calculate")
+    def calculate(self):
+        self.ke.interpolate(self.ke_expr)
```
tkarna, line 212: Should we construct a reusable `Interpolator` object? Are there cases where the used would actually like to project instead of interpolate? Projection preserves total energy I think. Maybe we could add a kwarg to choose.

tkarna (changes requested):
Thanks for the update, a couple comments.

thetis/diagnostics.py
jwallwork23: Ah yes, good idea!

thetis/diagnostics.py
```diff
@@ -2,12 +2,24 @@
...
-class VorticityCalculator2D(object):
+class DiagnosticCalculator(FrozenHasTraits):
+    """
+    Base class that defines the API for all diagnostic calculators.
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def calculate(self):
+        pass
```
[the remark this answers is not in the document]
jwallwork23: Fair enough. I changed it so that either `solve` or `__call__` works.

tkarna (approved):
Looks good, thanks for the updates!

jwallwork23:
Rebased on top of `master`.

tkarna (approved):

Issue #293, opened 2022-02-02
grid_tpxo9.nc download link error


MaihgrUmosch:
Please could I have some help with getting the netcdf tidal forcing files?

There is a link in an error thrown in forcing.py to download grid_tpxo9.nc but I can't get it to work.

ftp://ftp.oce.orst.edu/dist/tides/Global/tpxo9_netcdf.tar.gz

![image](https://user-images.githubusercontent.com/32267184/152122327-785e38a1-0410-42f2-a51a-497362a25e5c.png)

tkarna:
Thanks for reporting this, the download link is indeed obsolete. To get TPXO model data, you'll need to register on their website https://www.tpxo.net/global/tpxo9-atlas after which you'll receive a download link.

TPXO data format (e.g. variable and file names) has changed a little as well, so `TPXOTidalBoundaryForcing` will need to be adjusted a little. This update is on my list.

MaihgrUmosch:
OK

PR #294, opened 2022-02-02
TPXOTidalBoundaryForcing supports custom netcdf file names

tkarna:
This is needed to support different TPXO versions. File names were hard-coded to TPXO 9v1 file names, now set to 9v5a by
default.


tkarna:
Also removes the obsolete download link mentioned in #293.

stephankramer (approved):
Makes sense

PR #295, opened 2022-02-04
Tohoku tsunami inversion example

jwallwork23:
This PR adds a new example, building on the inversion tools added in #277. It models the 2011 Tohoku tsunami using linear shallow water equations.

Given data from 16 gauges, it can be used to invert for the tsunami source (given some choice of source model). Currently, only CG and DG sources are supported, with a mask to enforce that initial displacement is only near to the earthquake epicentre. I also intend to add some other commonly used source models from the literature and make a comparison.

The PR also extends `TimeseriesCallback2D` and `StationObservationManager` to accept start and end times. This is required in the Tohoku application because some gauges are further out into the Pacific and we don't want to overfit to whatever happens there before the tsunami arrives.


examples/tohoku_inversion/forward_run.py
```diff
@@ -0,0 +1,29 @@
...
+parser = argparse.ArgumentParser(
+    description="Tohoku tsunami propagation",
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument("--source_model", type=str, default="CG1")
+args = parser.parse_args()
+source_model = args.source_model
+
+# Solve forward
+pwd = os.path.abspath(os.path.dirname(__file__))
```
tkarna, line 18: This is needed for the tests to work? Ideally tests should not write to disk. Or if needed we should use a temp directory created with pytest setup.

tkarna (changes requested):
Looks good, only minor comments. We should make sure the CI tests run in reasonable time (a couple of minutes).

requirements.txt
```diff
@@ -3,3 +3,4 @@ netCDF4
 pyproj
 uptide
 pytz
+utm
```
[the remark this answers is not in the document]
jwallwork23: Ah okay cool, thanks

examples/tohoku_inversion/inverse_problem.py
```diff
@@ -0,0 +1,163 @@
...
+
+# compute regularization term
+op.J = reg_manager.eval_cost_function()
+
+# Solver and setup reduced functional
+solver_obj.iterate(export_func=cost_function)
+Jhat = ReducedFunctional(
+    op.J, op.control_list, derivative_cb_post=gradient_eval_callback
+)
+pause_annotation()
```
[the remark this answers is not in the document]
jwallwork23: `pause_annotation` is a function, whereas `stop_annotating` is a context manager. In fact, `pause_annotation` is the `__enter__` method for `stop_annotating`.

thetis/inversion_tools.py
```diff
@@ -182,21 +185,30 @@ def register_observation_data(self, station_names, variable, time,
...
         self.station_names = station_names
         self.variable = variable
         self.observation_time = time
         self.observation_values = values
         self.observation_x = x
         self.observation_y = y
+        if start_times is not None and end_times is None:
+            raise ValueError('If start_times is set then end_times should also be set')
+        if start_times is None and end_times is not None:
+            raise ValueError('If end_times is set then start_times should also be set')
```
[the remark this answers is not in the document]
jwallwork23: Good point.

thetis/callback.py
```diff
@@ -133,6 +133,8 @@ def export(self, variables, time=None, index=None):
         :type variables: tuple of float
         :kwarg int index: If provided, defines the time index in the file
         """
+        if variables is None:
+            return
```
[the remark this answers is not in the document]
jwallwork23: This is used in the Tohoku demo for gauges that aren't active at a particular time. An alternative would be to export `numpy.nan`.

jwallwork23:
Thanks for the review @tkarna! I think I addressed all of your comments.

thetis/callback.py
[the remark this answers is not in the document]
jwallwork23: With your recommendation to move the start/end time check to `DiagnosticCallback.evaluate`, this does indeed go away.

thetis/callback.py
```diff
@@ -675,6 +682,9 @@ def _initialize(self):
     def __call__(self):
         if not self._initialized:
             self._initialize()
+        t = self.solver_obj.simulation_time
+        if t < self.start_time or t > self.end_time:
+            return
```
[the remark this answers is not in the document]
tkarna: You haven't implemented this change, or am I missing something?

examples/tohoku_inversion/plot_convergence.py
```diff
@@ -0,0 +1,24 @@
+import glob
+import matplotlib.pyplot as plt
+import os
+
+
+fig, axes = plt.subplots()
+pwd = os.path.abspath(os.path.dirname(__file__))
+inv_dir = f"{pwd}/outputs_elev-init-optimization"
+for fpath in glob.glob(f"{inv_dir}_*"):
```
tkarna, line 9: If the directory does not exists the scripts exits silently and the used does not necessarily know what happened. So maybe raise an error if glob returns an empty list? Also, as the plot scripts are not tested, you do not need to define `pwd` here? Applies to other plot scripts as well.

thetis/callback.py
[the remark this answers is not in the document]
jwallwork23: Oh that's weird. I made the change, but must have dropped it when rebasing or something. Sorry about that.

examples/tohoku_inversion/plot_convergence.py
jwallwork23: Good points. Done!

thetis/callback.py
```diff
@@ -271,8 +271,10 @@ def evaluate(self, index=None):
         """
         Evaluates callback and pushes values to log and hdf file (if enabled)
         """
-        values = self.__call__()
         time = self.solver_obj.simulation_time
+        if time < self.start_time or time > self.end_time:
+            return
```
tkarna, line 276: `self.start_time` and `self.end_time` are not defined in this parent class?

thetis/callback.py
```diff
@@ -675,6 +682,7 @@ def _initialize(self):
     def __call__(self):
         if not self._initialized:
             self._initialize()
+        t = self.solver_obj.simulation_time
```
tkarna, line 685: t is not used?

thetis/callback.py
jwallwork23: Ah yes, I must have dropped those changes as well. Will look into it in more detail now.

thetis/callback.py
jwallwork23: Thanks, should be fixed now.

thetis/inversion_tools.py
```diff
@@ -182,21 +185,29 @@ def register_observation_data(self, station_names, variable, time,
...
+        num_stations = len(station_names)
         self.variable = variable
         self.observation_time = time
         self.observation_values = values
         self.observation_x = x
         self.observation_y = y
+        start_times = start_times or -numpy.ones(num_stations)*numpy.inf
+        self.obs_start_times = numpy.array(start_times)
+        end_times = end_times or numpy.ones(num_stations)*numpy.inf
+        self.obs_end_times = numpy.array(end_times)
```
jwallwork23, line 201: Apologies for repeatedly pushing to the PR. I had got this bit wrong (nan vs inf!) and it was breaking things in the `channel_inversion` example.

thetis/callback.py
```diff
@@ -161,7 +161,9 @@ def __init__(self, solver_obj, array_dim=1, attrs=None,
                  export_to_hdf5=True,
                  append_to_log=True,
                  include_time=True,
-                 hdf5_dtype='d'):
+                 hdf5_dtype='d',
+                 start_time=-numpy.inf,
+                 end_time=numpy.inf):
```
tkarna, line 166: I'd prefer that the default unspecified case is `None` in the API so you don't need to carry around `numpy.inf` as the default value in the derived classes. Internally setting to inf is fine.

tkarna (approved):
Thanks for the updates!

PR #297, opened 2022-02-14
Test channel inversion

jwallwork23:
@tkarna introduced a 2D channel inversion example in #277. Currently, only the forward run is being tested on CI (see [this comment](https://github.com/thetisproject/thetis/pull/277#discussion_r796430957)).

I _think_ this should do the trick: do the forward run on a subprocess when `inverse_problem.py` is run on CI.


tkarna:
Yeah, good idea. The problem here is that we are storing the outputs to disk. I think we could monkeypatch and create a temp directory and clean up afterwards. But I'm actually thinking whether it would be easier just to include the forward run time series files in the repository. Then the inverse problem could run independently.

examples/channel_inversion/model_config.py
```diff
@@ -41,12 +41,14 @@ def construct_solver(store_station_time_series=True, **model_options):
...
     options.horizontal_velocity_scale = u_mag
-    options.output_directory = 'outputs_forward'
+    options.output_directory = f'{pwd}/outputs_forward'
     options.fields_to_export = ['uv_2d', 'elev_2d']
     options.swe_timestepper_type = 'CrankNicolson'
     if not hasattr(options.swe_timestepper_options, 'use_automatic_timestep'):
         options.timestep = dt
     options.update(model_options)
+    if not options.no_exports:
+        File(f'{pwd}/manning.pvd').write(manning_2d)
```
tkarna, line 51: Oh, I think I forgot this in by mistake. Should be removed.

examples/channel_inversion/Makefile
```diff
@@ -0,0 +1,18 @@
+all: invert plot
+
+CONTROLS	= Bathymetry Manning InitialElev {Bathymetry,Manning} {Manning,InitialElev} {InitialElev,Bathymetry} {Bathymetry,Manning,InitialElev}
```
tkarna, line 3: I'm not sure if running all of the combinations makes sense, at least if the `Makefile` intended to demonstrate the capabilities to the user. I've really only tested the individual ones.

examples/channel_inversion/Makefile
```diff
@@ -0,0 +1,18 @@
...
+	done
+
+plot:
+	for controls in $(CONTROLS); do \
+		python3 plot_elevation_progress.py -c $$controls
+	done
+
+clean:
+	rm -rf __pycache__
+	rm -rf outputs_Bathymetry* outputs_Manning* outputs_InitialElev* manning*
```
tkarna, line 17: `manning*` can be removed (see below).

tkarna (approved):
looks good

examples/channel_inversion/model_config.py
jwallwork23: Done

tkarna (approved):
Thanks, looks good!

PR #299, opened 2022-02-17
Tohoku: fix bathymetry interpolation in parallel

jwallwork23:


jwallwork23:
Turns out the start/end time stuff wasn't parallel safe, either. Seems to work now!

thetis/inversion_tools.py
```diff
@@ -172,7 +172,7 @@ def __init__(self, mesh, J_scalar=None, output_directory='outputs'):
         self.initialzed = False
 
     def register_observation_data(self, station_names, variable, time,
-                                  values, x, y, start_times=None, end_times=None):
+                                  values, x, y):
```
tkarna, line 175: You've removed `start_times` and `end_times` from `register_observation_data`; this feature is now only implemented in `load_observation_data`. However, `load_observation_data` is just a (poorly named) wrapper for `register_observation_data` for the case where the observations are stored in the Thetis diagnostic time series format. So, if the observation data is in another format, start/end times cannot be set. To be more generic, you could store the start/end times in `register_observation_data` as before, and then add the logic for handling local stations in `construct_evaluator`.

tkarna (approved):

thetis/inversion_tools.py
jwallwork23: Good point, pushed a fix.

tkarna (approved):

PR #300, opened 2022-02-21
Weighted cost function

jwallwork23:
This PR adds two helper functions for inversion problems:
* default kwargs to pass to `ReducedFunctional`, namely a gradient evaluation callback;
* extract the cost function from the various inversion objects.

Inside `get_cost_function`, it also supports weighting the contributions at each station by the associated variance in the data. I turned this on for the Tohoku example and it appears to DTRT in my local runs.


tkarna (approved):
Thanks, looks good!

PR #302, opened 2022-02-28
Tohoku: add 'radial' and 'box' source models

jwallwork23:
This PR adds two additional source models to the Tohoku inversion example. Instead of inverting for a field in CG or DG space, we invert for an array of controls in R-space, which provide the coefficients for linear combinations of basis functions.

Note that this PR includes commits from #300, so that will need to be merged first.

### DIFF
--- a/examples/tohoku_inversion/inverse_problem.py
+++ b/examples/tohoku_inversion/inverse_problem.py
@@ -52,6 +52,7 @@ else:
 op = inversion_tools.OptimisationProgress(
     options.output_directory,
     real=source_model[:2] not in ("CG", "DG"),
+    no_exports=options.no_exports,
 )
 for c in controls:
     op.add_control(c)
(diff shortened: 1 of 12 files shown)


test/examples/test_examples.py
```diff
@@ -42,6 +42,7 @@
     'tohoku_inversion/plot_convergence.py',
     'tohoku_inversion/plot_elevation_initial_guess.py',
     'tohoku_inversion/plot_elevation_progress.py',
+    'tohoku_inversion/plot_optimised_source.py',
```
tkarna, line 45: Did you forget to include this file?

examples/tohoku_inversion/inverse_problem.py
```diff
@@ -18,16 +18,20 @@
 parser.add_argument("--ftol", type=float, default=1.0e-05)
 parser.add_argument("--no-consistency-test", action="store_true")
 parser.add_argument("--no-taylor-test", action="store_true")
+parser.add_argument("--suffix", type=str, default="")
```
tkarna, line 21: default can be none as you you have a if check below in any case?

tkarna (approved):

PR #304, opened 2022-03-08
Fix `get_facet_areas`

jwallwork23:
The utility function `get_facet_areas` is broken because we are no longer allowed to assemble the diagonal on interior facets. Switching from Jacobi to SOR is probably sufficient.


jwallwork23:
See https://github.com/firedrakeproject/firedrake/pull/2370.

wence-:
You need to remove the `matfree` part (I don't think we support sor for matrix-free operators).

jwallwork23:
Yes, I did drop `matfree`. I just realised I can get away with
```
sp = {
    "snes_type": "ksponly",
    "ksp_type": "preonly",
    "pc_type": "jacobi",
}
```

tkarna (approved):

PR #306, opened 2022-03-23
Tohoku: subtract initial elevation from bathymetry

jwallwork23:
It is common practice to subtract the initial elevation field from the bathymetry field in tsunami simulations.


tkarna (approved):
Sure, looks good to me

PR #307, opened 2022-03-29
Use textwrap instead of ipython_genutils

### LINKED ISSUE #305: import fails without ipython_genutils
Flagged by the Firedrake CI:

https://github.com/firedrakeproject/firedrake/runs/5524703420?check_suite_focus=true#step:8:12895

I think there is a sneaky dependency on the `ipython_genutils` package. It's possible that `ipython_genutils` is no longer installed by default with ipython.

jwallwork23:
Attempt to fix #305

### DIFF
--- a/thetis/configuration.py
+++ b/thetis/configuration.py
@@ -1,7 +1,7 @@
 """
 Utility function and extensions to traitlets used for specifying Thetis options
 """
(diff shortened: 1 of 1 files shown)


tkarna (approved):
Oh yes, thanks for looking into this!

PR #308, opened 2022-03-30
Tohoku: add Okada source model

jwallwork23:
The Tohoku tsunami source inversion example allows us to compare a variety of different source models that are commonly used in the literature. Currently we can represent the source in finite element spaces or as linear combinations of radial basis functions ("radial") or piecewise constant indicator functions ("box"). For the linear combinations, we can interpret each contribution as related to a single subfault within the greater earthquake fault.
This PR adds the Okada source model, which also involves a linear combination of contributions from subfaults, but in this case the model used on each subfault is a nonlinear function of nine "Okada parameters". Four of these parameters are fixed when we assume a particular arrangement of subfaults, but the other five can be optimised.
Below is a comparison between various source models on both the coarse mesh used in the example (suffix zero) and a finer mesh (suffix 1). The currently implemented models appear to be fairly comparable, except that the "box" basis convergence progress stagnates on the finer mesh. The Okada model seems to be a bit slower to catch up, but there is a promising downward turn in the later iterations. I will do some longer runs to establish whether or not this actually leads to anything.
optimization_progress
Aside from calibrating earthquake-tsunami models, these kind of experiments with the Okada model are valuable because they provide an approximation of physical parameters related to the earthquake.
The PR also saves station data in CSV format, like with the North Sea example.


jwallwork23:
@tkarna Doing this nicely ended up being more work than expected. Apologies that the PR is quite big now.

It adds:
* A regularization strategy that works when the controls are from R space.
* `okada.py` and `sources.py` modules within the Tohoku example. These could potentially become a `tsunami.py` or `earthquake.py` module of Thetis, if there would be interest in using them for other tsunamis.

tkarna (changes requested):
Thanks, looks good to me. The source definition indeed looks quite generic now.

I think the biggest missing bit is rebasing and migration to use CheckpointFile.

tkarna (approved):

examples/tohoku_inversion/sources.py
```diff
@@ -0,0 +1,398 @@
...
+    Base class for generating tsunami source conditions.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    fault_length = PositiveFloat(560e3, help="Length of fault plane").tag(config=True)
+    fault_width = PositiveFloat(240e3, help="Width of fault plane").tag(config=True)
+    strike_angle = PositiveFloat(198, help="Angle of fault plane from North in degrees").tag(config=True)
+    fault_centroid_lon = BoundedFloat(143.6, bounds=(-180, 180), help="Longitude of the fault centroid").tag(config=True)
+    fault_centroid_lat = BoundedFloat(37.9, bounds=(-90, 90), help="Latitude of the fault centroid").tag(config=True)
```
jwallwork23, line 25: CI seemed to be stuck, so I fixed a minor TODO (previously here), rebased and pushed to get it going again.

tkarna (approved):

PR #309, opened 2022-04-05
Option for appending to log

jwallwork23:
It would be nice to be able to continue writing to the same log file when a new solver is created, e.g. during mesh adaptive runs.


tkarna (approved):
Sure, that's a good idea!

jwallwork23:
Thanks! Can confirm it works for adaptive simulations I've been doing.

PR #311, opened 2022-04-11
InversionManager

jwallwork23:
I had a think about the way we are doing inversion and thought perhaps the API could be a bit simpler.

My idea here is to drop `OptimisationProgress` and create an `InversionManager` class that does almost everything that you would want for inversion problems. I decided to have `RegularizationManager` be created inside of it because `InversionManager` handles the controls and you need these to setup the regularisation. So the only thing that you still have to create separately is the `StationObservationManager`.

Hopefully it is agreeable. The only thing I am wondering is if the code style could be more like Thetis' solver objects by introducing a frozen `InversionOptions` class (like `ModelOptions`) that comes with `InversionManager` and which you modify afterwards, as opposed to passing multiple kwargs to the constructor.

I also realised that most of Thetis uses US English, whereas some (but not all) of the inversion tools I introduced use British English. It should be consistent now (unlike the text in this PR!).

### DIFF
--- a/examples/tohoku_inversion/inverse_problem.py
+++ b/examples/tohoku_inversion/inverse_problem.py
@@ -44,25 +44,13 @@ options = solver_obj.options
 if not options.no_exports:
     print_output(f"Exporting to {options.output_directory}")
 
-# Choose optimisation parameters
+# Setup controls
 control = "elev_init"
 nc = len(controls)
 if nc == 1:
     control_bounds = [-numpy.inf, numpy.inf]
 else:
     control_bounds = [[-numpy.inf] * nc, [numpy.inf] * nc]
-op = inversion_tools.OptimisationProgress(
-    options.output_directory,
-    real=source_model[:2] not in ("CG", "DG"),
-    no_exports=options.no_exports,
-)
-for c in controls:
-    op.add_control(c)
-
-# Define the (appropriately scaled) cost function
-dt_const = Constant(solver_obj.dt)
-total_time_const = Constant(options.simulation_end_time)
-J_scalar = dt_const / total_time_const
 
 # Set up observation and regularization managers
 observation_data_dir = f"{pwd}/observations"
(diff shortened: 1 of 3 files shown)


tkarna (dismissed):
Excellent, this is great! I think we can create the inversion options class later on if it looks useful.

My only comment: the `J_scalar` needs to be passed both to the `StationObservationManager` and `InversionManager`. I wonder if it could be passed only to the latter, as `InversionManager` evaluates the full cost functional. I guess it's just a matter of getting the annotation working (currently the scalar is a `Constant` and must be plugged inside the integral form).

jwallwork23:
Good point. I think this should fix it. (Kept the `Constant` `J_scalar` attribute of `StationObservationManager`, but set it from `InversionManager` when it is initialised.)

thetis/inversion_tools.py
```diff
@@ -12,37 +12,64 @@
...
+        """
+        assert isinstance(sta_manager, StationObservationManager)
+        self.sta_manager = sta_manager
+        self.reg_manager = None
         self.output_dir = output_dir
         self.no_exports = no_exports or real
         self.real = real
+        self.penalty_parameters = penalty_parameters
+        self.cost_function_scaling = cost_function_scaling
+        self.sta_manager.cost_function_scaling.assign(cost_function_scaling)
```
tkarna, line 47: `cost_function_scaling` is `None` by default which will raise an error in the assign. If `None` use `Constant(1)`?

examples/tohoku_inversion/inverse_problem.py
```diff
@@ -31,7 +31,7 @@
 elev_init, controls = initial_condition(mesh2d, source_model=source_model)
 
 # Setup PDE
-output_dir = f"{pwd}/outputs_elev-init-optimization_{source_model}"
+output_dir = f"{pwd}/TMPoutputs_elev-init-optimization_{source_model}"
```
tkarna, line 34: Why TMPoutputs?

tkarna (approved):

examples/tohoku_inversion/inverse_problem.py
jwallwork23: Oops!

tkarna (approved):

PR #312, opened 2022-04-18
Coordinate system object

tkarna:
Adds coordinate system object that provides transformations to/from latitude, longitude coordinate system.

API symmary:
```python
coordsys = UTMCoordinateSystem(utm_zone=34)
# convert arbitrary coordinates
lon, lat = ...  # numpy arrays
x, y = coordsys.to_xy(lon, lat)  # mesh coordinates
lon, lat = coordsys.to_lonlat(x, y)  # inverse
# rotate vectors defined in (lon, lat) to mesh coordinate system
u_east, v_north = ...  # vector components at (lon, lat) locations
vect_rotator = coordsys.get_vector_rotator(lon, lat)
u_x, v_y = vect_rotator(u_east, v_north)
# get mesh coordinate Function in lon, lat system
mesh2d = ...
ll = get_mesh_lonlat_function(mesh2d)
```
The (lon, lat) coordinate system is WGS84.

### DIFF
--- a/thetis/coordsys.py
+++ b/thetis/coordsys.py
@@ -5,15 +5,131 @@ Uses pyproj library.
 import firedrake as fd
 import pyproj
 import numpy
+from abc import ABC, abstractmethod
 
-
-UTM_ZONE10 = pyproj.Proj(
-    proj='utm',
-    zone=10,
-    datum='WGS84',
-    units='m',
-    errcheck=True)
 LL_WGS84 = pyproj.Proj(proj='latlong', datum='WGS84', errcheck=True)
+
+
+class CoordinateSystem(ABC):
+    """
+    Base class for horizontal coordinate systems
+
+    Provides methods for coordinate transformations etc.
+    """
+    @abstractmethod
+    def to_lonlat(self, x, y):
+        """Convert coordinates to latitude and longitude"""
+        pass
+
+    @abstractmethod
+    def get_vector_rotator(self, x, y):
+        """
+        Returns a vector rotator object.
+
+        The rotator converst vector-valued data to/from longitude, latitude
+        coordinates.
+        """
+        pass
+
+
+def proj_transform(x, y, trans=None, source=None, destination=None):
+    """
+    Transform coordinates from source to target system.
+
+    :arg x,y: coordinates, float or numpy.array_like
+    :kwarg trans: pyproj Transformer object (optional)
+    :kwarg source: source coordinate system, Proj object
+    :kwarg destination: destination coordinate system, Proj object
+    """
+    if trans is None:
+        assert source is not None and destination is not None, \
+            'Either trans or source and destination must be defined'
+        trans = None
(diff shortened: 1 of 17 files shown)


thetis/coordsys.py
```diff
@@ -5,17 +5,133 @@
...
-UTM_ZONE10 = pyproj.Proj(
-    proj='utm',
-    zone=10,
-    datum='WGS84',
-    units='m',
-    errcheck=True)
 LL_WGS84 = pyproj.Proj(proj='latlong', datum='WGS84', errcheck=True)
 
 
+class MeshCoordSystem(ABC):
```
tkarna, line 13: or `MeshCoordinateSystem`, `CoordinateSystem`?

tkarna:
I think this is ready now. I'll add spherical coordinate system in another PR. To simplify Thetis API, we could add a coordinate system aware mesh object (e.g. Firedrake Mesh that includes a coordinate system object).

thetis/coordsys.py
```diff
@@ -5,17 +5,133 @@
...
+        assert source is not None and destination is not None, \
+            'Either trans or source and destination must be defined'
+        trans = None
+    x_is_array = isinstance(x, numpy.ndarray)
+    y_is_array = isinstance(y, numpy.ndarray)
+    numpy_inputs = x_is_array or y_is_array
+    if numpy_inputs:
+        assert x_is_array and y_is_array, 'both x and y must be numpy arrays'
+        assert x.shape == y.shape, 'x and y must have same shape'
+        # NOTE proj may give wrong results if inputs contain NaNs
```
jwallwork23, line 54: Maybe raise this as a warning? It is unlikely that users will see this note in the source code.

thetis/coordsys.py
jwallwork23: I would prefer `CoordinateSystem` because the (sub)class doesn't depend on a mesh.

thetis/interpolation.py
```diff
@@ -326,6 +325,7 @@ def _get_subset_nodes(grid_x, grid_y, target_x, target_y):
     return nodes, ind_x, ind_y
 
 
+# TODO can I just inherit from ABC instead?
```
jwallwork23, line 328: Should be fine.

jwallwork23 (approved):
Looks great!

thetis/coordsys.py
tkarna: Yeah this is the justification why we the workaround, i.e. figure out the `good_ix` indices instead of passing `x` and `y` directly to `transform`. I'll update the comment

thetis/coordsys.py
tkarna: I agree, changed.

thetis/interpolation.py
tkarna: I've updated the `ABC` inheritance now

jwallwork23 (approved):

PR #313, opened 2022-04-19
Date-aware solver

tkarna:
Adds options `options.simulation_initital_date` and `options.simulation_end_date`, both must be timezone-aware `datetime` objects. Thus simulation duration can now be defined as:
```python
init_date = datetime.datetime(2022, 4, 1, tzinfo=pytz.utc)
end_date = datetime.datetime(2022, 4, 3, 12, tzinfo=pytz.utc)
solver_obj = ...
solver_obj.options.simulation_initital_date = init_date
solver_obj.options.simulation_end_date = end_date
```

Simulation time is then printed as dates in the log, e.g.
```
Running simulation
 from 2022-04-01T00:00:00UTC
 to   2022-04-03T12:00:00UTC
  exp  iter                   time     eta norm       u norm   Tcpu
    0     0 2022-04-01T00:00:00UTC       0.0000       0.0000   0.00
    1    10 2022-04-01T01:00:00UTC    1052.5844     184.1268   1.75
...
   60   600 2022-04-03T12:00:00UTC   11961.0809    8648.7590   0.30
```

The solver is date-aware if `simulation_initial_date` is defined. The end time can now be defined in two optional ways: either via `simulation_end_date` (if `simulation_initial_date` is set) or `simulation_end_time` (duration in seconds) as before, at least one must be defined. If both are set, `simulation_end_time` is ignored with a warning. Internally, the solver object still uses simulation time in seconds.

This PR updates the log outputs in general, i.e. column headers are now printed first on a separate line:
```
  exp  iter                   time     eta norm       u norm   Tcpu
    0     0                   0.00       0.0000       0.0000   0.00
    1    10                3600.00    1052.5844     184.1268   1.71
...
```


tkarna:
I made the choice that the `datetime` objects must include the time zone. This adds a little boilerplate code if most applications use UTC (which I think is the case). Nevertheless I thought that it's safer to force the user to think about time zones as this is a common user error (e.g. forcings are in UTC but observations are in local time zone).

thetis/solver.py
```diff
@@ -1077,20 +1078,43 @@ def load_state(self, i_export, outputdir=None, t=None, iteration=None):
...
-                                 u=norm_u, cpu=cputime))
+        # generate simulation time string
+        if self.options.simulation_initial_date is not None:
+            now = self.options.simulation_initial_date + datetime.timedelta(seconds=self.simulation_time)
+            time_str = f'{now:%Y-%m-%dT%H:%M:%S%Z}'
+        else:
+            time_str = f'{self.simulation_time:.2f}'.rjust(22)
+
+        norm_h = norm(self.fields.solution_2d.split()[1])
+        norm_u = norm(self.fields.solution_2d.split()[0])
```
jwallwork23, line 1096: Previously this was
```
norm_h = norm(self.fields.elev_2d)
norm_u = norm(self.fields.uv_3d)
```
Is this okay?

thetis/solver2d.py
```diff
@@ -834,33 +835,52 @@ def load_state(self, i_export, outputdir=None, t=None, iteration=None):
...
-                line = ('{iexp:5d} {i:5d} T={t:10.2f} '
-                        '{label:16s}: {q:10.4f} {cpu:5.2f}')
-
-                norm_label = label if len(label) < 3 or label[-3:] != '_2d' else label[:-3]
-                print_output(line.format(iexp=self.i_export, i=self.iteration,
-                                         t=self.simulation_time,
-                                         label=norm_label + ' norm',
-                                         q=norm_q, cpu=cputime))
+                e = (label, norm_q, '10.4f')
+                entries.append(e)
```
jwallwork23, line 863: I assume you realised that this isn't the same behaviour as before. I was going to protest, but then I ran the multiple tracers demo and it actually looks better than what we had before:
```
  exp  iter                   time    bell_2d    cone_2d slot_cyl_2d   Tcpu
    0     0                   0.00     1.0119     1.0289     1.0800   0.00
    1    20                   0.21     1.0119     1.0289     1.0792   0.52
    2    40                   0.42     1.0119     1.0289     1.0788   0.35
    3    60                   0.63     1.0119     1.0289     1.0785   0.36
    4    80                   0.84     1.0119     1.0289     1.0783   0.38
    5   100                   1.05     1.0119     1.0289     1.0781   0.35
```
Bunching all tracers on the same line could get messy if people are using >5 tracer fields or using long names, but I don't think anyone is using Thetis for that right now so it probably isn't a problem.

thetis/solver.py
```diff
@@ -1077,20 +1078,43 @@ def load_state(self, i_export, outputdir=None, t=None, iteration=None):
...
+
+        if print_header:
+            # generate header
+            widths = [len(('{:' + e[2] + '}').format(e[1])) for e in entries]
+            header = ' '.join([e[0].rjust(w) for e, w in zip(entries, widths)])
+            print_output(header)
+
+        # generate line
+        line_fmt = ' '.join(['{:' + e[2] + '}' for e in entries])
+        line = line_fmt.format(*[e[1] for e in entries])
```
jwallwork23, line 1116: You can get this with
`line = ' '.join([f'{e[1]:{e[2]}}' for e in entries])`

thetis/solver.py
```diff
@@ -1077,20 +1078,43 @@ def load_state(self, i_export, outputdir=None, t=None, iteration=None):
...
+            ('time', time_str, '22s'),
+            ('eta norm', norm_h, '12.4f'),
+            ('u norm', norm_u, '12.4f'),
+            ('Tcpu', cputime, '6.2f'),
+        ]
+
+        if print_header:
+            # generate header
+            widths = [len(('{:' + e[2] + '}').format(e[1])) for e in entries]
+            header = ' '.join([e[0].rjust(w) for e, w in zip(entries, widths)])
```
jwallwork23, line 1111: You can get this with
`header = ' '.join([e[0].rjust(len(f'{e[1]:{e[2]}}')) for e in entries])`

thetis/solver2d.py
```diff
@@ -834,33 +835,52 @@ def load_state(self, i_export, outputdir=None, t=None, iteration=None):
...
+
+        if print_header:
+            # generate header
+            widths = [len(('{:' + e[2] + '}').format(e[1])) for e in entries]
+            header = ' '.join([e[0].rjust(w) for e, w in zip(entries, widths)])
+            print_output(header)
+
+        # generate line
+        line_fmt = ' '.join(['{:' + e[2] + '}' for e in entries])
+        line = line_fmt.format(*[e[1] for e in entries])
```
jwallwork23, line 882: Can be cut down as above, if desired.

jwallwork23 (changes requested):
Good idea, this will definitely be a handy update.

Mostly just some minor comments and suggestions. However, I do think that the output is getting a bit dense. Perhaps replace the `T` between date and time and put more spaces around the date? Also the time zone could just be printed once in the boilerplate, so that we have something like
```
Running simulation
 from 2022-04-01  00:00:00
 to   2022-04-03  12:00:00
  exp  iter                  time     eta norm       u norm   Tcpu
    0     0  2022-04-01  00:00:00       0.0000       0.0000   0.00
    1    10  2022-04-01  01:00:00    1052.5844     184.1268   1.75
...
   60   600  2022-04-03  12:00:00   11961.0809    8648.7590   0.30
```
Maybe we could even have a `date` column?
```
Running simulation
 from 2022-04-01  00:00:00
 to   2022-04-03  12:00:00
  exp  iter        date      time     eta norm       u norm   Tcpu
    0     0  2022-04-01  00:00:00       0.0000       0.0000   0.00
    1    10  2022-04-01  01:00:00    1052.5844     184.1268   1.75
...
   60   600  2022-04-03  12:00:00   11961.0809    8648.7590   0.30
```

jwallwork23:
I hope you don't mind, but I just pushed some fixes for the demo version of the North Sea example. (We forgot to update it when we updated coordsys and I added in the new functionality you introduce here.)

thetis/solver2d.py
tkarna: Yeah I'd prefer to print everything on the same line. I've been trying to keep the log output within 80 chars or so, but you're right that it's difficult to ensure in all cases.

thetis/solver.py
tkarna: Good catch! I guess I can't just copy from the 2D solver...

tkarna:
I think I've addressed your comments. Now the date-aware log looks like:
```
Running simulation
 from 2022-04-01 00:00:00 UTC
 to   2022-04-03 12:00:00 UTC
  exp  iter        date      time       eta norm         u norm   Tcpu
    0     0  2022-04-01  00:00:00         0.0000         0.0000   0.00
    1    10  2022-04-01  01:00:00      1052.5844       184.1268   1.73
...
```
and naive version
```
  exp  iter            time       eta norm         u norm   Tcpu
    0     0            0.00         0.0000         0.0000   0.00
    1    10         3600.00      1052.5844       184.1268   1.64
...
```
Changes:
- log: added separate `date` and `time` columns
- log: time zone is included in the `Running simulation` dates
- log: adjusted column widths (norms can be long)
- callback: store init date in "units" attribute of the time array, not as global "time_units" attribute
- North Sea example: `plot_elevation.py` now reads the dates automagically from hdf5/netcdf files

jwallwork23 (approved):
Looks great, thanks!

PR #314, opened 2022-04-22
docs: fix indent from textwrap

tkarna:
Fix `textwrap.indent` call in docs generation.


jwallwork23 (approved):
Yeah my bad - should have been more careful with #307.

PR #316, opened 2022-05-05
HDF5 outputs use CheckpointFile

tkarna:
Thetis now uses Firedrake's `CheckpointFile` HDF5 output format introduced in [PR1917](firedrakeproject/firedrake/pull/1917). API is the same, only file format changes.

For backward compatibility `solver.load_state` still supports reading in old `DumbCheckpoint` files via `legacy_mode=True`  argument. This option will be deprecated by the end of 2022.

This also affects outputs of inversion runs: `InversionManager` now uses `HDF5Exporter` internally and uses `CheckpointFile` format by default. The control fields are now stored in files
```
outputs/hdf5/control_00_00000.h5  # 00=control field index, 00000=iteration index
```
instead of
```
outputs/hdf5/control_00_0000.h5
```


thetis/inversion_tools.py
```diff
@@ -49,6 +50,7 @@ def __init__(self, sta_manager, output_dir='outputs', no_exports=False, real=Fal
         self.test_gradient = test_gradient
         self.outfiles_m = []
         self.outfiles_dJdm = []
+        self.control_exporters = []
```
tkarna, line 53: We could plausibly use `ExportManager` to handle all of these outputs. The user could then decide whether to export these in vtk or hdf5 format.

examples/channel_inversion/inverse_problem.py
```diff
@@ -27,22 +27,24 @@
 do_taylor_test = not args.no_taylor_test
 no_exports = os.getenv('THETIS_REGRESSION_TEST') is not None
 
-# Create the solver object
+# Set output directory
 pwd = os.path.abspath(os.path.dirname(__file__))
+output_dir = f'{pwd}/outputs_new_' + '-'.join(controls) + '-opt'
+
+# Create the solver object
+# NOTE We disable all exports here as the they would only be
```
jwallwork23, line 35: "the they"

thetis/inversion_tools.py
jwallwork23: Makes sense.

examples/channel_inversion/inverse_problem.py
jwallwork23, line 35: I guess we should do the same for the Tohoku example as well then?

jwallwork23 (dismissed):
Thanks for this!

examples/channel_inversion/inverse_problem.py
tkarna: Yeah, good point. Looking into it there seem to be a bunch of examples that need an updatel.

tkarna:
OK, I think I've updated the examples now.

examples/north_sea/plot_setup.py
```diff
@@ -15,46 +18,43 @@
...
-    h5.load(bathymetry_2d)
 fig, axes = plt.subplots(figsize=(6.4, 4.8))
 triplot(mesh2d, axes=axes, boundary_kw={"linewidth": 0.5, "edgecolor": "k"})
-bathymetry_2d.dat.data[:] = numpy.log10(bathymetry_2d.dat.data[:])
-tc = tricontourf(bathymetry_2d, axes=axes, levels=50)
-cb = fig.colorbar(tc, ax=axes)
-ticks = cb.get_ticks()
-cb.set_ticklabels([f"{10 ** t:.0f}" for t in ticks])
+norm = mcolors.LogNorm(vmin=10, vmax=3200)
+tc = tripcolor(bathymetry_2d, axes=axes, norm=norm)
```
tkarna, line 29: I took the liberty to reformulate the plotting bits

jwallwork23: Fine by me, thanks!

thetis/solver.py
```diff
@@ -996,26 +996,51 @@ def export(self):
...
+            solver_obj.load_state(40, outputdir='outputs_spinup')
+
+        Continue simulation from another output directory and reset the
+        counters,
+
+        .. code-block:: python
+            solver_obj.load_state(40, outputdir='outputs_spinup',
+                iteration=0, t=0, i_export=0)
+
+        Model state can be loaded if all pronostic state variables have been
```
jwallwork23, line 1023: "pronostic"

jwallwork23 (dismissed):
Thanks for the additions! My comments are all just typos/wording.

jwallwork23 (approved):

PR #319, opened 2022-06-21
Add traitlets to requirements

jwallwork23:
Thetis depends on `traitlets` but this is not currently mentioned in `requirements.txt`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ netCDF4
 pyproj
 uptide
 pytz
+traitlets


jwallwork23:
Noticed this when testing out installing Thetis via `firedrake-spack`. The relevant [package.py](https://github.com/firedrakeproject/firedrake-spack/blob/main/packages/py-thetis/package.py) config was written based on `requirements.txt` so this dependency was missing.

tkarna (approved):
Nice catch!

PR #323, opened 2022-09-01
Update spun-up state for North Sea demo

jwallwork23:
When the new checkpointing came in, we updated the bathymetry file for the North Sea demo to use the new format. However, it looks like we forgot to update the spun-up state and so running the demo fails because "legacy mode" isn't turned on.


stephankramer:
This was done as part of #345

PR #324, opened 2022-09-30
Create tsunami module

joewallwork:
Minor update that's been on my to do list for a while: create a Thetis module for the tsunami utils. Would allow us to consider tsunamis other than Tohoku and will also allow me to write a Tohoku demo (also on the to do list).


stephankramer (dismissed):
@jwallwork23 : still planning to add more tsunami cases? In any case this seems to move a module from an example to thetis/ - which looks fine to me.

jwallwork23:
> @jwallwork23 : still planning to add more tsunami cases? In any case this seems to move a module from an example to thetis/ - which looks fine to me.

Thanks. No plans to do any tsunami modelling in the near future myself.

jwallwork23:
[Merged `master` into this branch to fix conflicts.]

stephankramer:
@joewallwork : same here. If you still want to proceed could you resolve conflicts, or otherwise close it? On the surface looks to be a useful change, but understand if you don't have the time for it now.

joewallwork:
> @joewallwork : same here. If you still want to proceed could you resolve conflicts, or otherwise close it? On the surface looks to be a useful change, but understand if you don't have the time for it now.

Seemed like a straightforward merge - let's see what the CI says.

thetis/tsunami.py
stephankramer, line 165:
```suggestion
            cond = ufl.And(ufl.And(x > -L / 2, x < L / 2), ufl.And(y > -W / 2, y < W / 2))
```
Just playing around with AI, and it points me at what looks like a serious bug that was already there... I'm a little confused: has it been applying the source in completely the wrong place? Previously it would have been only applied where `y > W/2`, so above the rectangle?

thetis/tsunami.py
```diff
@@ -398,3 +407,110 @@ def control_bounds(self):
...
+    x2 = -x2
+    sn = deg2sin(P.dip)
+    cs = deg2cos(P.dip)
+    p = x2 * cs + z_bot * sn
+    q = x2 * sn - z_bot * cs
+
+    def strike_slip(y1, y2):
+        d_bar = y2 * sn - q * cs
+        r = sqrt(y1**2 + y2**2 + q**2)
+        a4 = 2 * poisson * (ln(r + d_bar) - sn * ln(r + y2)) / cs
```
stephankramer, line 490: Another AI observation: we're dividing here by cs, cosine of the dip, but in the OkadaArrayTsunamiSource, we bound `0<=dip<=90` so in an optimisation it's allowed to have dip=90 which would blow things up. Might well be that this is fine, if that in fact means the optimisation will steer things away from dip=90.

jhill1:
I'll take over this.

joewallwork:
> I'll take over this.

Thanks Jon!

PR #325, opened 2022-09-30
Date aware examples

joewallwork:
A short PR to make the Tohoku example date-aware. I also noticed that the North Sea example wasn't set up properly for testing: a shorter `t_end` value was set but unused.


examples/north_sea/model_config.py
```diff
@@ -97,15 +97,15 @@ def construct_solver(spinup=False, store_station_time_series=True, **model_optio
...
+    default_start_date = datetime.datetime(2022, 1, 1, 0, 0, tzinfo=sim_tz)
+    default_end_date = datetime.datetime(2022, 1, 2, 0, 0, tzinfo=sim_tz)
     start_date = model_options.pop("start_date", default_start_date)
     end_date = model_options.pop("end_date", default_end_date)
     dt = 3600.0
     t_export = 3600.0
     t_end = (end_date - start_date).total_seconds()
     if os.getenv("THETIS_REGRESSION_TEST") is not None:
-        t_end = 5 * t_export
+        end_date = datetime.datetime(2022, 1, 1, 0, 5, tzinfo=sim_tz)
```
stephankramer, line 108: That's 5 minutes past midnight, no? but the timestep is in hours?

stephankramer (dismissed):
Other than that looks fine. Sorry this seems to have sat there a while - not sure you were still intending to merge

examples/north_sea/model_config.py
jwallwork23: Good spot! Fixed in https://github.com/thetisproject/thetis/pull/325/commits/1b073425f37bb9ba501a3a3fa8a3471257ddf1ac.

jwallwork23:
[Merged in `master` to fix conflicts.]

stephankramer:
@joewallwork : could you resolve the conflicts  - if you still want this in, I think this is good to go - otherwise could you close?

joewallwork:
> @joewallwork : could you resolve the conflicts - if you still want this in, I think this is good to go - otherwise could you close?

I think it should be sufficient to just go with `main` in the conflict - let's see what the CI says.

cpjordan:
Whilst you're looking at this, I've made some changes as part of #438 to speed up CI by reducing `num_subfaults_par` x `num_subfaults_perp` (CI only). I assume this is sensible and you could do this now to speed up testing if it's helpful: https://github.com/thetisproject/thetis/pull/438#issuecomment-4518675684

examples/tohoku_inversion/model_config.py
```diff
@@ -117,12 +118,19 @@ def construct_solver(bathymetry_2d, elev_init, store_station_time_series=True, *
     """
     mesh2d = elev_init.function_space().mesh()
 
-    t_end = 2 * 3600.0
+    # Setup temporal discretisation
     u_mag = Constant(5.0)
+    default_start_date = datetime.datetime(2011, 3, 11, 14, 46, tzinfo=sim_tz)
+    default_end_date = datetime.datetime(2011, 3, 11, 16, 46, tzinfo=sim_tz)
```
stephankramer, line 124:
```suggestion
    default_start_date = sim_tz.localize(datetime.datetime(2011, 3, 11, 14, 46))
    default_end_date = sim_tz.localize(datetime.datetime(2011, 3, 11, 16, 46))
```
If you do:
`
print(datetime.datetime(2011, 3, 11, 14, 46, tzinfo=sim_tz), sim_tz.localize(datetime.datetime(2011, 3, 11, 14, 46)))
`
you get
`
2011-03-11 14:46:00+09:19 2011-03-11 14:46:00+09:00
`
i.e. the former uses a time zone of UTC+9:19 instead of the correct UTC+9. If I understand correctly, datetime in the standard library only has limited support for timezones, and it doesn't handle the fact that timezones change historically (i.e. Japan used to be on UTC+9:19 but is now UTC+9

examples/tohoku_inversion/model_config.py
```diff
@@ -117,12 +118,19 @@ def construct_solver(bathymetry_2d, elev_init, store_station_time_series=True, *
...
     t_export = 60.0
     dt = 60.0
     if os.getenv("THETIS_REGRESSION_TEST") is not None:
-        t_end = 5 * t_export
+        model_options["simulation_initial_date"] = default_start_date
+        model_options["simulation_end_date"] = datetime.datetime(2011, 3, 11, 14, 51, tzinfo=sim_tz)
```
stephankramer, line 129:
```suggestion
        model_options["simulation_end_date"] = sim_tz.localize(datetime.datetime(2011, 3, 11, 14, 51))
```

stephankramer (changes requested):
Also added a fix for restarting with simulation_initial_date

jhill1:
I will take over this PR

joewallwork:
> I will take over this PR

Thanks Jon! Apologies that I never got back to it.

PR #327, opened 2022-10-18
Various fixes to improve running of pytest suite

stephankramer:
* test_sed_slide.py: this was previously a run-alone python script, but since pytest tries to import every python file starting with test_ it means this test was always being run even when trying to run a single actual pytest test. Now turned into an actual test
* test_examples: this test runs all python files (except those excluded) found in the examples/ directories. However the id was set to just the basename of these python files, and since there are multiple files with the same name in different subdirectories of examples/, you end up with multiple tests with the exact same name, meaning you can't easily work out which one has actually failed from the actions logs and also you can't run them individiually. Now just using the full path as the id. Note that the -k argument for pytest does not like forward slashes. If you want to run a specific test example however you can run something like:

```pytest -k"tohoku_inversion and inverse_problem.py" test/examples/```

* test_bath_smoothing.py renamed to bath_smoothing_test.py, again so it doesn't accidentily get picked up if you forget to tell pytest to only look in test/

### DIFF
--- a/examples/columbia_plume/bath_smoothing_test.py
+++ b/examples/columbia_plume/bath_smoothing_test.py
@@ -3,6 +3,7 @@ Columbia river plume simulation
 ===============================
 """
 from thetis import *
+from firedrake.output.vtk_output import VTKFile
 from bathymetry import get_bathymetry, smooth_bathymetry, smooth_bathymetry_at_bnd
 comm = COMM_WORLD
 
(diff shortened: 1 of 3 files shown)


jwallwork23 (dismissed):
These changes make sense to me, yeah! Thanks.

jwallwork23 (approved):

Issue #328, opened 2022-11-15
Issue installing qmesh


acse-ej321:
@jhill1 
I am running into install issues when trying to run 'pip install qmesh' for qmesh. I am hoping to run the example scripts found in: https://github.com/thetisproject/thetis/tree/master/examples/north_sea. The issue seems to be related to a try / except block in the setup.py around line 108, checking the gmsh version, and seems to be something that was depreciated between 2.7 and 3.x.

jhill1:
Note to self: Update docs to qmesh3 in this example

Sijbrandb:
I have problems installing qmesh as well. Seems that something is going wrong with building the wheel. Would be great if someone could help me out!

(base) C:\Users\sijbr>pip install qmesh3
Collecting qmesh3
  Using cached qmesh3-1.0.3.tar.gz (82 kB)
Requirement already satisfied: setuptools-qmesh in c:\users\sijbr\anaconda3\lib\site-packages (from qmesh3) (1.1.0)
Requirement already satisfied: GFD_basisChangeTools in c:\users\sijbr\anaconda3\lib\site-packages (from qmesh3) (1.0.0)
Requirement already satisfied: GitPython in c:\users\sijbr\anaconda3\lib\site-packages (from setuptools-qmesh->qmesh3) (3.1.29)
Requirement already satisfied: gitdb<5,>=4.0.1 in c:\users\sijbr\anaconda3\lib\site-packages (from GitPython->setuptools-qmesh->qmesh3) (4.0.9)
Requirement already satisfied: smmap<6,>=3.0.1 in c:\users\sijbr\anaconda3\lib\site-packages (from gitdb<5,>=4.0.1->GitPython->setuptools-qmesh->qmesh3) (5.0.0)
Building wheels for collected packages: qmesh3
  Building wheel for qmesh3 (setup.py) ... error
  ERROR: Command errored out with exit status 1:
[...]

PR #329, opened 2022-11-28
minor fixes to both coastline, boundary shpfiles and meshing script. 

jhill1:
Fixes the issue #328 so that the generate_mesh.py script now works as expected. Note qmesh will need updating to 1.04 (pending)


stephankramer:
@jhill1 : sorry, didn't see you already had a PR open on the North Sea example. We would need to merge with #344 for this to work (and you probably need to regenerate the bathymetry checkpoint file as I see the mesh has changed).

I think this PR looks good. The README in this example should probably be updated to point to where to get qmesh3 from?

> Note qmesh will need updating to 1.04 (pending)

Is this still ongoing? Does this refer to https://bitbucket.org/qmesh-developers/qmesh/pull-requests/6 ?

If you need more time, I would prefer to merge #344 first: the example doesn't run at all at the moment (and isn't even tested!) which really should be fixed.

jhill1:
No worries; merge #344 and I'll merge that into here and redo the PR. qmesh3 has been moved to [github](https://github.com/qmesh3) now (but still needs work).

stephankramer:
@jhill1: #344 was merged as part of #345. So if you rebase/merge on master this should be good to go

PR #330, opened 2022-12-01
assign method now only allows linear combinations of functions

stephankramer:


examples/tidal_barrage/lagoon2d.py
```diff
@@ -91,7 +92,7 @@
 
 
 def update_forcings(t_new,):
-    tidal_elev.assign(Constant(tanh((t_new)/(4*3600.)) * amplitude * sin(omega * t_new)))
+    tidal_elev.assign(math.tanh((t_new)/(4*3600.)) * sin(omega * t_new) * amplitude)
```
wence-, line 95: Uhhh, does this do compilation for every new `t_new` (or does `assign` no longer do code generation @connorjward)

stephankramer: No, that's what I thought as well at first, but these are all just floats, so the entire thing should evaluate to a float. Actually, looking at this again tidal_elev should just be a `Constant`

wence-: In the old world f.assign(float) would
compile code hardcoding the value of the float, but not sure what happens now

connorjward: Since it's a `Constant` we just assign directly to the `Global` ([source](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/constant.py#L142)). And if this were a `Function`, since we don't do codegen any more, it makes no difference.

thangel (approved):
- self.m2 += mu_comp[i_comp]*mu_comp[i_comp]
            + self.m2.interpolate(self.m2 + mu_comp[i_comp]*mu_comp[i_comp])  

Just a comment - why is the above amendment necessary - it looks fine but unsure why this needed to be rewritten like this

stephankramer:
> ```
>         - self.m2 += mu_comp[i_comp]*mu_comp[i_comp]
>         + self.m2.interpolate(self.m2 + mu_comp[i_comp]*mu_comp[i_comp])  
> ```
> 
> Just a comment - why is the above amendment necessary - it looks fine but unsure why this needed to be rewritten like this

If you do `function += expr` that's basically overloaded to do `function.assign(function + expr)` and assign is now much more restricted: you can only assign constants or linear combinations of functions (in the same function space)

PR #334, opened 2023-01-10
Pin traitlets version

tkarna:
Every new install hits the traitlets version issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ netCDF4
 pyproj
 uptide
 pytz
-traitlets
+traitlets==5.6.0


stephankramer:
I think you're referring to a different issue than in #332 which was introduced traitlets>=5.7 So 5.7.x works now

It seems pypi is now on 5.8 which has introduced a new issue
```
    from thetis import *
thetis/__init__.py:5: in <module>
    import thetis.solver as solver  # NOQA
thetis/solver.py:9: in <module>
    from . import turbulence
thetis/turbulence.py:146: in <module>
    from .options import GLSModelOptions, PacanowskiPhilanderModelOptions
thetis/options.py:7: in <module>
    from .configuration import *
thetis/configuration.py:75: in <module>
    class PositiveInteger(Integer):
E   NameError: name 'Integer' is not defined
```

I don't think pinning on an exact version is the solution here. Traitlets is quite a low level package (used in ipython for instance) and having different packages demand different versions would cause issues. I'll take a look if there's a quick fix

tkarna:
Yes, that is the same error I get. Thanks for taking a look.

stephankramer:
See #335

PR #335, opened 2023-01-10
Fix thetis with traitlets 5.8

stephankramer:
Apparently "from traitlets import *" doesn't actually import anything we wan't. Import symbols explicitly instead.


tkarna (approved):
OK, I see. Thanks for tracking this down!

PR #338, opened 2023-03-08
Incorporate discrete turbines updates into master

s1626662:
Incorporate the most up-to-date discrete turbines implementation into master from the discrete-turbines-cleanup branch which has been rebased.


examples/discrete_turbines/channel-optimisation.py
```diff
@@ -0,0 +1,227 @@
...
+outflow_bc = {'elev': 0.}
+freeslip_bc = {'un': 0.}
+
+solver_obj.bnd_functions['shallow_water'] = {
+    left_tag: inflow_bc,
+    right_tag: outflow_bc,
+    coasts_tag: freeslip_bc
+}
+
+# initialise discrete turbine farm characteristics
```
stephankramer, line 66: Could you add a bit more explanation in the comments on what these two arrays represent - mention "thrust curve"?

examples/discrete_turbines/channel-optimisation.py
```diff
@@ -0,0 +1,227 @@
...
+# Regardless of this, the area where the turbines are places need to have sufficiently high
+# resolution to resolve the Gaussian bump functions that represent the turbines
+options.discrete_tidal_turbine_farms[2] = farm_options
+
+# set initial condition (initial velocity should not be exactly 0 to avoid failures in the Newton solve)
+solver_obj.assign_initial_conditions(uv=(as_vector((1e-3, 0.0))))
+
+# Operation of tidal turbine farm through a callback
+# This can be used for diagnostic output (to the log and hdf5 file)
+# but here is also used to define the functional that we optimised for
```
stephankramer, line 106: *optimise

examples/discrete_turbines/tidal_array.py
```diff
@@ -0,0 +1,108 @@
+from thetis import *
```
stephankramer, line 1: Could you add some comments at the start that describe what setup is defined here?

examples/discrete_turbines/tidal_array.py
```diff
@@ -0,0 +1,108 @@
...
+
+options.swe_timestepper_options.solver_parameters = {
+    'snes_type': 'newtonls',
+    'snes_rtol': 1e-3,
+    'snes_linesearch_type': 'bt',
+    'snes_max_it': 20,
+    'ksp_type': 'preonly',
+    'pc_type': 'lu',
+    'pc_factor_mat_solver_type': 'mumps',
+}
```
stephankramer, line 80: Are these needed? In examples I prefer to use default solver settings, or if needed explain why they are needed...

stephankramer (commented):
I've just removed the rated turbine implementation (we shouldn't merge anything that doesn't actually work and/or is not tested). With a few more comments in the example I think this is good to go!

examples/discrete_turbines/channel-optimisation.py
s1626662: Done

examples/discrete_turbines/tidal_array.py
s1626662: Done

stephankramer (approved):
Nice. Thanks for cleaning this up @s1626662
Could you merge this with a merge commit? There's quite a stream of little commits, don't care about squashing it, but it's probably easier to see what happened if there is a merge commit

thangel (approved):
Really nice contributions and glad that we are finally making this merge;
A few steps to do next:

- Alter turbine subdomain structure to allow different TidalTurbineFarm objects for the same subdomain
- Make changes on thetis main branch to avoid the errors associated with split()

PR #339, opened 2023-03-23
Fix parallel tests and tohoku example

stephankramer:
Use firedrake's pytest-mpi to mark parallel tests. Previously we have copied and pasted the pytest configuration that handles tests that have been marked paralel from firedrake's test suite, but this has now been separated
(https://github.com/firedrakeproject/firedrake/pull/2763). This reduces the code on our side and has the advantage we no longer need to copy updates and fixes as they'll get applied in pytest-mpi.

Fix test associated with the tohoku inversion example: cannot mix function spaces derived from different mesh objects even if they correspond to the exact same mesh loaded from the same checkpoint file.


jwallwork23 (approved):
Makes sense, thanks @stephankramer!

PR #340, opened 2023-05-15
Replace firedrake par_loops with op2 par_loops

stephankramer:
Also directly unit test get_cell_widths.

### DIFF
--- a/test/operations/test_operations_2d-3d.py
+++ b/test/operations/test_operations_2d-3d.py
@@ -189,6 +189,11 @@ def test_minimum_angle(mesh2d):
     assert numpy.allclose(min_angle, pi/4)
 
 
+def test_cell_widths(mesh2d):
+    cell_width = utility.get_cell_widths_2d(mesh2d).dat.data
+    assert numpy.allclose(cell_width, 1./5)
+
+
 if __name__ == '__main__':
     """Run all tests"""
     import os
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Nice, I hadn't seen the `arity` attribute before so will take note of that. Looks good, assuming the tests pass.

stephankramer:
Should have mentioned: removing "firedrake par_loops" is because they have been removed from firedrake, so the tests were already failing.

The next issue is that firedrake no longer builds petsc with Eigen support which is used in anisotropic_cell_size. @jwallwork23: since it's just a simple 2x2 matrix maybe we can just replace it with dlaev2 from lapack?

stephankramer:
Well I've done something different now. Not sure I entirely understand the cell size measure tbh as it seems to produce different answers depending on the orientation (i.e. same physical triangle can give different sizes), but in any case it does what it did before. Could you have a look again @jwallwork23 ?

jwallwork23 (approved):
Thanks for this, @stephankramer. I hadn't noticed the dependence on orientation so will do some digging to see if there is a better approach. Probably not a priority right now, though, so maybe we go ahead and merge?

stephankramer:
Yeah I don't think it matters much indeed - you need a "minimal grid distance" of some sort, and this seems to work so let's just stick with that.

PR #343, opened 2023-06-27
Fixing issues around domain-less Constants and adjoint

stephankramer:
Changes required to fix adjoint:
* Constants used as control (and any Constant that changes value during the run) requires a domain for the adjoint to work
* callbacks pre and post derivative calculation of a ReducedFunctional now need to pass through resp. the control and the derivatives
* avoid assigning to Constants internally as this causes issues with the adjoint - notably this was done in updating fields_old in some timesteppers (we maybe should just get rid of fields_old altogether)

### DIFF
--- a/examples/discrete_turbines/channel-optimisation.py
+++ b/examples/discrete_turbines/channel-optimisation.py
@@ -91,7 +91,7 @@ site_y_start = 80.
 r = farm_options.turbine_options.diameter/2.
 
 # a list contains the coordinates of all turbines: regular, non-staggered 4 x 2 layout
-farm_options.turbine_coordinates = [[Constant(x+cos(y)), Constant(y+numpy.sin(x))]
+farm_options.turbine_coordinates = [[Constant(x+cos(y), domain=mesh2d), Constant(y+numpy.sin(x), domain=mesh2d)]
                                     for x in np.linspace(site_x_start + 4*r, site_x_start + site_x - 4*r, 4)
                                     for y in np.linspace(site_y_start + 0.5*site_y-2*r,
                                                          site_y_start + 0.5*site_y+2*r, 2)]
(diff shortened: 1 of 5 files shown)


thetis/timeintegrator.py
```diff
@@ -85,6 +85,32 @@ def advance_picard(self, t, update_forcings=None, update_lagged=True, update_fie
...
+                    # are now just Functions on the Real space, so are dealt with under isinstance(..., Function)
+                    self.fields_old[k] = self.fields[k]
+
+    def update_fields_old(self):
+        """
+        Update the values of fields_old with those of fields
+        """
+        for k in sorted(self.fields):
+            if isinstance(self.fields[k], Function):
+                self.fields_old[k].assign(self.fields[k])
```
jwallwork23, line 112: Looks good. It's also a nice refactoring to move this to the base class anyway :)

examples/discrete_turbines/channel-optimisation.py
```diff
@@ -166,7 +166,8 @@ def eval_cb_pre(controls):
 
 def derivative_cb_pre(controls):
     print_output("ADJOINT RUN:")
-    print_output("positions: {}".format([float(c_) for c_ in controls]))
+    print_output("positions: {}".format([float(c_.data()) for c_ in controls]))
```
jwallwork23, line 169: Out of interest, why was this change needed? Does taking the `float` of a `Constant` no longer do what you expect?

jwallwork23 (approved):
Thanks Stephan, this all looks great. Good typo catches, too.

examples/discrete_turbines/channel-optimisation.py
stephankramer: Ah, because if you create a `Constant(..., domain=...)` you no longer actually get an object of type `Constant` it actually returns a `Function` on the R FunctionSpace. And taking the float of a Function is not defined of course.

jwallwork23: Ah okay makes sense, thanks.

PR #344, opened 2023-07-31
Fix northsea bathymetry

stephankramer:
Run north sea example in CI.  Only runs a few timesteps of north sea example spinup, to ensure we pick up stuff like the bathymetry checkpoint breaking, etc.
    
Also ensure we use the mesh from the bathymetry checkpoint rather reading separately from .msh file as this is no longer allowed.


examples/north_sea/model_config.py
```diff
@@ -152,23 +152,33 @@ def construct_solver(spinup=False, store_station_time_series=True, **model_optio
...
+        tbnd.set_tidal_field(0.0)
+    else:
+        # for CI testing setup dummy tbnd
+        elev_tide_2d.assign(1.)
+
+        def set_tidal_field(t):
+            pass
+
+        tbnd = types.SimpleNamespace()
+        tbnd.set_tidal_field = set_tidal_field
```
jwallwork23, line 181: Nice, I haven't seen this approach before. Looks similar to using `unittest.mock`.

examples/north_sea/model_config.py
```diff
@@ -74,20 +76,18 @@ def construct_solver(spinup=False, store_station_time_series=True, **model_optio
         simulation and a function for updating forcings
     """
 
+    pwd = os.path.abspath(os.path.dirname(__file__))
+    h5_file_name = os.path.join(pwd, "north_sea_bathymetry.h5")
+    with CheckpointFile(h5_file_name, "r") as f:
+        mesh2d = f.load_mesh("firedrake_default")
+        bathymetry_2d = f.load_function(mesh2d, "Bathymetry")
```
jwallwork23, line 83: Please could you also make this and other changes related to setting up the bathymetry in `demos/demo_2d_north_sea.py`?

jwallwork23 (changes requested):
Thanks for this @stephankramer. Just one small request but otherwise looks good to go.

examples/north_sea/model_config.py
stephankramer: Good point. Done

jwallwork23 (approved):

stephankramer:
Sorry this made me realize that the example is still broken: the spinup run works but picking up from a checkpoint is more generally broken. This needs fixing first (see #345) as we need agreement how to fix. Will probably merge the relevant bits from this branch into the other PR.

PR #345, opened 2023-08-02
Fix Thetis checkpoint loading

### LINKED ISSUE #336: Error with new checkpointing
If I run a script for a short time; with hdf5 output. Then continue from one of the checkpoints, I receive:

> (firedrake) jh1889@envpc319 ~/work/projects/Tsunamis/thetis_runs/rectangle_test $ python run_wd_cont.py 
> Using constant wetting and drying parameter (value 1.00)
> /home/jh1889/firedrake/src/firedrake/firedrake/function.py:317: FutureWarning: The .split() method is deprecated, please use the .subfunctions property instead
>   warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
> /home/jh1889/firedrake/src/firedrake/firedrake/functionspaceimpl.py:125: FutureWarning: The .split() method is deprecated, please use the .subfunctions property instead
>   warnings.warn("The .split() method is deprecated, please use the .subfunctions property instead", category=FutureWarning)
> Element family: dg-dg, degree: 1
> 2D cell type: triangle
> 2D mesh: 6704 vertices, 13106 elements
> Horizontal element size: 337.93 ... 829.43 m
> Number of 2D elevation DOFs: 39318
> Number of 2D velocity DOFs: 78636
> Number of cores: 1
> Elevation DOFs per core: ~39318.0
> dt = 10
> Using time integrator: CrankNicolson
> Traceback (most recent call last):
>   File "/home/jh1889/work/projects/Tsunamis/thetis_runs/rectangle_test/run_wd_cont.py", line 78, in <module>
>     solverObj.load_state(checkpoint, outputdir=output_dir)
>   File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
>   File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
>   File "/home/jh1889/firedrake/src/thetis/thetis/solver2d.py", line 840, in load_state
>     e.exporters['uv_2d'].load(i_stored, self.fields.uv_2d)
>   File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
>   File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
>   File "/home/jh1889/firedrake/src/thetis/thetis/exporter.py", line 224, in load
>     function.assign(g)
>   File "PETSc/Log.pyx", line 115, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
>   File "PETSc/Log.pyx", line 116, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
>   File "/home/jh1889/firedrake/src/firedrake/firedrake/adjoint/function.py", line 135, in wrapper
>     ret = assign(self, other, *args, **kwargs)
>   File "/home/jh1889/firedrake/src/firedrake/firedrake/function.py", line 404, in assign
>     Assigner(self, expr, subset).assign()
>   File "/home/jh1889/firedrake/src/firedrake/firedrake/assign.py", line 141, in __init__
>     raise ValueError("All functions in the expression must use the same "
> ValueError: All functions in the expression must use the same mesh as the assignee

The mesh coming from the checkpoint is not identical according to UFL:

From assign.py in firedrake:
```python
                if coeff.ufl_element() != assignee.ufl_element():
                    raise ValueError("All functions in the expression must have the same "
                                     "element as the assignee")
                if coeff.ufl_domain() != assignee.ufl_domain():
                    raise ValueError("All functions in the expression must use the same "
                                     "mesh as the assignee")
```
Version is as of 1st Feb:
jh1889@envpc319 ~/firedrake/src/firedrake $ git log
commit c0b45ce2123fdeadf358df1d5655ce42f3b3d74b (HEAD -> master, origin/master, origin/HEAD)
Author: Uygar <83709209+uygarkov@users.noreply.github.com>
Date:   Wed Feb 1 17:16:40 2023 +0100

The[ code to reproduce can be found here](https://drive.google.com/drive/folders/15RCLrmHE1TLJJE-LmlI5ZJt3_VXpzr58?usp=sharing).

stephankramer:
Previously we loaded the mesh from checkpoint, then loaded the field into a temporary function based on that mesh, and then assigned it to the field on the solver object, which is based on whatever mesh the user provided us. This is no longer allowed (and is also unsafe).

Now we would ask the user to provide us with a mesh read from a checkpoint, and directly read in the field into a function based on that same mesh. To help the user achieve that, added a read_mesh_from_checkpoint() utility function. To discuss: one of the downsides of this approach is that it requires the user to point to a specific .h5 file, whereas load_state only requires the output directory (and adds `'hdf5/Velocity2d_000...'` etc.)

See also discussion https://github.com/firedrakeproject/firedrake/discussions/3053

Should also fix #336. Note that both the case where the user had already loaded the mesh from a checkpoint - because they needed to load some other preprocessed data like in #336 - but also the case where the user re-creates the mesh by reading from a .msh file and then calling load_state() are currently broken. This should fix both cases - but requires changing the setup to read the mesh from checkpoint for the latter case.


jhill1:
Sounds good to me. Also need to edit the docs page to match this: https://github.com/thetisproject/thetis/blob/master/docs/source/outputs_and_visu.rst
Something like:
```
diff --git a/docs/source/outputs_and_visu.rst b/docs/source/outputs_and_visu.rst
index 45aea1c1..3bd6023d 100644
--- a/docs/source/outputs_and_visu.rst
+++ b/docs/source/outputs_and_visu.rst
@@ -98,15 +98,17 @@ Restarting a simulation
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 If you have stored the required HDF5 files, you can continue a simulation
-using :py:meth:`~.FlowSolver.load_state` method, provided that you use the same
-mesh and the same number of MPI processes. This call replaces the
+using :py:meth:`~.FlowSolver.load_state` method, provided that you use the
+mesh from that checkpoint file. This call replaces the
 :py:meth:`~.FlowSolver.assign_initial_conditions` call.
 If initial conditions are not set, add ``load_state`` call above
 the :py:meth:`~.FlowSolver.iterate` call.
 
-In the simplest form, one only defines the export index that is used as initial
-condition::
+In the simplest form, the mesh needs to be loaded from a HDF5 checkpointfile
+and then the index given to the :py:meth:`~.FlowSolver.load_state` method::
 
+    mesh2d = read_mesh_from_checkpoint(outputdir + '/hdf5/Velocity2d_155.h5')
+    ...set-up the thetis solver object with mesh2d...
     solver_obj.load_state(155)
 
 This also loads simulation time from the stored state.
```
However, when I was writing the docs, it didn't seem the most elegant of solutions. I wonder if we could instantiate the solver object with either a mesh object or a checkpoint file? Perhaps that's a job for later!

stephankramer:
Thanks Jon.

> However, when I was writing the docs, it didn't seem the most elegant of solutions. I wonder if we could instantiate the solver object with either a mesh object or a checkpoint file?

Yeah thought about that. The issue is that typically users need to do some set up that already requires the mesh, like creating a bathymetry function - also you wouldn't be able to provide the bathymetry as a function on creation. So then you would get something like:
```
solver_obj = FlowSolver2d('outputs/hdf5/Velocity2d_155.h5')
mesh2d = solver_obj.mesh2d
V = get_function_space(mesh2d, "CG", 1)
bathymetry = Function(V).interpolate(...)
solver_obj.set_bathymetry(bathymetry)
# or alternatively, something like
solver_obj.options.bathymetry = bathymetry
```

jhill1:
Agreed; that is messier for more users, then the checkpointing issue. Let's stick with your solution. I can't approve the review, but I do approve

thetis/utility.py
```diff
@@ -248,6 +248,32 @@ def create_directory(path, comm=COMM_WORLD):
...
+    separate runs/scripts, make sure that only one of those scripts created the
+    original mesh by reading a .msh file or using a mesh creation utility like
+    RectangleMesh, and all other script read their mesh from the checkpoint
+    created by that first script.  For example, a preprocessing script might
+    read in a .msh file and interpolate and smoothen the bathymetry on that mesh
+    and write out the result in a checkpoint. A first Thetis run should read
+    its mesh from that checkpoint, and may then save a series of hdf5 files.
+    A second Thetis run can then read in the mesh and bathymetry from
+    the checkpoint of the preprocessing script, and call load_state()
+    to pick up from the previous run.
```
jwallwork23, line 267: This text is all really useful but could you break it into separate paragraphs?

jwallwork23 (changes requested):
Looks good. Just one minor request. Thanks!

docs/Makefile
```diff
@@ -26,7 +26,7 @@ copy_demos:
 	cp ../demos/*.py source/demos
 	cp ../demos/figures/*.png source/demos
 	cp ../demos/demo_references.bib source/demos
-	for file in source/demos/*.py; do ${VIRTUAL_ENV}/src/firedrake/pylit/pylit.py -c $$file; mv $$file.txt $$file.rst; done
+	for file in source/demos/*.py; do pylit -c $$file; mv $$file.txt $$file.rst; done
```
jwallwork23, line 29: Would you mind adding `pylit` to `requirements.txt`?

jwallwork23 (approved):
Thanks! Just one very minor request but otherwise looks good to go.

jwallwork23 (approved):

PR #346, opened 2023-08-11
BUGFIX: Rossby wave test was being run on dg-dg only

stephankramer:
The call to options.update(model_options) which a.o. sets the element_family option only happened after the call to create_function_spaces() so that the function spaces were always based on the default dg-dg rather than going through dg-dg, dg-cg, rt-dg and bdm-dg as intended.

Also the setting of the timestepper type should probably not be repeated as it may overwrite the timestepper sub-options that were set previously.


stephankramer:
Thanks for reporting @ACSE-yl4222

PR #348, opened 2023-08-30
Adding three new thetis papers

jhill1:
Lee et al, 2022, Hill et al 2023 and Woodroffe et al 2023.


stephankramer (approved):
Nice! TIL the word ooid

PR #349, opened 2023-11-09
Incorporate changes to form2indicator from goalie.

stephankramer:
These were changes to form2indicator
related to the Cofunction introduction that have been fixed in goalie (formerly part of pyroteus).


jwallwork23 (approved):
Thanks for updating this, @stephankramer. Changes look good. Also a step towards getting Thetis hooked back up with goal-oriented error estimation/mesh adaptation.

It does make me wonder, though, whether we should avoid the duplication so this doesn't happen again. We obviously don't want Thetis to depend on Goalie, but maybe we could put something in here so that this part of the code only works if you have Goalie installed and if so imports `form2indicator`?

stephankramer:
Yes, I agree, we could make it a soft/optional dependency indeed

PR #353, opened 2023-12-16
Add two 2023 papers

jwallwork23:
We should also rebuild the website after merging this (I notice Jon's additions in #348 don't appear).

### DIFF
--- a/docs/source/publications.rst
+++ b/docs/source/publications.rst
@@ -30,7 +30,12 @@ Hill, J., Rush, G., Peakall, J., Johnson, M., Hodson, L., Barlow, N.L.M.,
 Bowman, E.T., Gehrels, W.R., Hodgson, D.M., Kesserwani, G:
 Resolving tsunami wave dynamics: integrating sedimentology and numerical modelling.
 *The Depositional Record*
-doi: `10.1002/dep2.247 <https://doi.org/10.1002/dep2.247`>_, 2023.
+doi: `10.1002/dep2.247 <https://doi.org/10.1002/dep2.247>`_, 2023.
+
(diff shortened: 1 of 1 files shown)


stephankramer (approved):
Thanks, yes we should rebuild indeed

Added a few more recent ones and fixed broken links in Jon's recent ones.

jwallwork23:
All links look good. Have updated the webpage.

docs/source/publications.rst
```diff
@@ -30,15 +30,20 @@ Hill, J., Rush, G., Peakall, J., Johnson, M., Hodson, L., Barlow, N.L.M.,
 Bowman, E.T., Gehrels, W.R., Hodgson, D.M., Kesserwani, G:
 Resolving tsunami wave dynamics: integrating sedimentology and numerical modelling.
 *The Depositional Record*
-doi: `10.1002/dep2.247 <https://doi.org/10.1002/dep2.247`>_, 2023.
+doi: `10.1002/dep2.247 <https://doi.org/10.1002/dep2.247>`_, 2023.
+
+Zhang, C., Zhang, J., Angeloudis, A., Zhou, Y., Kramer, S.C., Piggott, M.D.:
```
thangel, line 35: I am not sure this paper used any thetis simulations... are we sure we want to list this here?

stephankramer: Ah yes, you're quite right - that one shouldn't have been included

thangel: I plan with @cpjordan to add some changes to some examples  shortly, maybe I can try to add a few more then and correct this

PR #357, opened 2024-01-17
Fix jupyter notebook demos

cpjordan:
- Update plot imports from Firedrake
- Update from DumbCheckpoint to CheckpointFile
- Update corresponding text
- Update bathymetry file loading hint on tsunami example


cpjordan:
I think the failure is because I have used a differently named Jupyter kernel locally which alters the notebook metadata. I will re-run the notebooks with the default kernel name and hopefully that fixes this issue.

jwallwork23 (changes requested):
Thanks very much for this!

I notice you added a new cell at the end of the tsunami demo that describes using `CheckpointFile` instead of `DumbCheckpoint`. Would you mind updating the bathymetry file so that it is in the new format? i.e. read with `DumbCheckpoint` and then overwrite with `CheckpointFile`? It would be good if we could move away from referring to `DumbCheckpoint` at all as it was replaced a while ago now.

Would you mind making a couple of other minor changes?
* Drop the "this mesh was actually created using [qmesh](http://qmesh.org/) mentioned in yesterday's lecture" bracket in the tsunami notebook.
* Capitalise Firedrake and Python in both notebooks.
Much appreciated.

jwallwork23 (approved):
Nice - I see you fixed a few things I'd missed, too :)

thangel (approved):

PR #358, opened 2024-02-27
Allow multiple tidal energy farms per subdomain

cpjordan:
- Farms must now be defined in a list for each subdomain ID
- Warning provided when a list is not used
- Corresponding updates to examples

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -881,12 +881,12 @@ class ModelOptions2d(CommonModelOptions):
 
         Note this is only relevant if `use_automatic_wetting_and_drying_alpha` is set to ``True``.
         """).tag(config=True)
-    tidal_turbine_farms = Dict(trait=TidalTurbineFarmOptions(),
-                               default_value={}, help='Dictionary mapping subdomain ids to the options of the corresponding farm')
+    tidal_turbine_farms = Dict(trait=List(TidalTurbineFarmOptions()),
+                               default_value={}, help='Dictionary mapping subdomain ids to the options of the corresponding farm(s)')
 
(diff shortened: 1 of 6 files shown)


thetis/options.py
```diff
@@ -881,12 +881,12 @@ class ModelOptions2d(CommonModelOptions):
 
         Note this is only relevant if `use_automatic_wetting_and_drying_alpha` is set to ``True``.
         """).tag(config=True)
-    tidal_turbine_farms = Dict(trait=TidalTurbineFarmOptions(),
-                               default_value={}, help='Dictionary mapping subdomain ids to the options of the corresponding farm')
+    tidal_turbine_farms = Dict(trait=List(TidalTurbineFarmOptions()),
+                               default_value={}, help='Dictionary mapping subdomain ids to the options of the corresponding farm(s)')
```
stephankramer, line 885:
```suggestion
                               default_value={}, help='Dictionary mapping subdomain ids to a list of TidalTurbineFarmOptions instances corresponding
                               to one or more farms.')
```

stephankramer (approved):
Nice. Looks good. Just a small suggestion and one question: what is the use case for having multiple farms per domain id for the continuous case? Presumably that means having multiple turbine densities that act in the same area? I'm fine with also changing that to a list but was just wondering

thetis/options.py
thangel: I suppose you could have different densities for different turbine characteristics?

cpjordan: I've just done that out of consistency more than anything, happy to go with whatever is preferred!

thangel (approved):

PR #362, opened 2024-04-09
Discrete turbines callback example

cpjordan:
Adds a callback to the existing discrete turbines example that allows monitoring of individual discrete turbines. We compare the results to the farm callback to make sure that the results are the same. This callback can be used with or without turbines added to the simulation for comparison in the 'ambient' case.

Callback is slow for meshes with a large number of elements in the subdomain of interest, but fine for smaller scales.


thangel (approved):
This looks good to me - and I think the callback functionality within the example is helpful to distinguish between core thetis utilities and applying application-specific extensions for use cases. It looks good and instructive

PR #365, opened 2024-06-07
Reduce Firedrake deprectation warnings

stephankramer:
1) Don't use Constant(value, domain=mesh). New utility function
   domain_constant(value, mesh) that creates Function in R space
2) Don't use interpolate(expr, V) (without assemble) instead use
   Function(V).interpolate(expr)


cpjordan (approved):
Looks good - thanks Stephan!

PR #366, opened 2024-06-07
Update channel inversion example

cpjordan:
- Correct file paths in example for plotting
- Correct VTKFile paths for import
- Load times for the optimisation fields (one timestep less than existing files)
- Import from collections.abc (deprecation update)
- Update makefile so it works again and cycles through all the stations for plotting

### DIFF
--- a/examples/channel_inversion/Makefile
+++ b/examples/channel_inversion/Makefile
@@ -1,6 +1,7 @@
 all: invert plot
 
 CONTROLS	= Bathymetry Manning InitialElev
+STATIONS    = stationA stationB stationC stationD stationE
 
 invert:
 	for controls in $(CONTROLS); do \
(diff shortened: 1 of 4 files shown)


stephankramer (changes requested):
VTKFile should now be available again from the top level firedrake module, and so again directly from either `from firedrake import *` or `from thetis import *` (see https://github.com/firedrakeproject/firedrake/pull/3535). I prefer doing that as internal module hierarchies are likely to change...

Other than that, all looks good!

cpjordan:
Ah - was working on an older version of Firedrake as it's a pain to update on my system because of the Linux version. Updated now & tests running, thanks Stephan!

stephankramer (approved):

PR #370, opened 2024-07-19
Fix time lag issue for inversion

cpjordan:
- t passed to cost_fn(t) was out of sync by one timestep so the optimal result was not correct for channel inversion
- Now need to pass the cost function through the export_func rather than through update_forcings
- Typo fixed in Tohoku Makefile as well

### DIFF
--- a/thetis/inversion_tools.py
+++ b/thetis/inversion_tools.py
@@ -3,7 +3,7 @@ from firedrake.adjoint import *
 import ufl
 from .configuration import FrozenHasTraits
 from .solver2d import FlowSolver2d
-from .utility import create_directory, print_function_value_range, get_functionspace, unfrozen
+from .utility import create_directory, print_function_value_range, get_functionspace, unfrozen, domain_constant
 from .log import print_output
 from .diagnostics import HessianRecoverer2D
 from .exporter import HDF5Exporter
@@ -477,7 +477,7 @@ class StationObservationManager:
         self.mod_values_0d = fd.Function(self.fs_points_0d, name='model values')
         self.indicator_0d = fd.Function(self.fs_points_0d, name='station use indicator')
         self.indicator_0d.assign(1.0)
-        self.cost_function_scaling_0d = fd.Constant(0.0, domain=mesh0d)
+        self.cost_function_scaling_0d = domain_constant(0.0, mesh0d)
         self.cost_function_scaling_0d.assign(self.cost_function_scaling)
         self.station_weight_0d = fd.Function(self.fs_points_0d, name='station-wise weighting')
         self.station_weight_0d.assign(1.0)
@@ -615,7 +615,7 @@ class RegularizationCalculator(abc.ABC):
         self.regularization_expr = 0
         self.mesh = function.function_space().mesh()
         # calculate mesh area (to scale the cost function)
-        self.mesh_area = fd.assemble(fd.Constant(1.0, domain=self.mesh) * fd.dx)
+        self.mesh_area = fd.assemble(fd.Constant(1.0) * fd.dx(domain=self.mesh))
         self.name = function.name()
 
     def eval_cost_function(self):
--- a/thetis/inversion_tools.py
+++ b/thetis/inversion_tools.py
@@ -488,26 +488,29 @@ class StationObservationManager:
         # Construct timeseries interpolator
         self.station_interpolators = []
         self.local_station_index = []
-        for i in range(self.fs_points_0d.dof_dset.size):
-            # loop over local DOFs and match coordinates to observations
-            # NOTE this must be done manually as VertexOnlyMesh reorders points
-            x_mesh, y_mesh = mesh0d.coordinates.dat.data[i, :]
-            xy_diff = xy - numpy.array([x_mesh, y_mesh])
-            xy_dist = numpy.sqrt(xy_diff[:, 0]**2 + xy_diff[:, 1]**2)
-            j = numpy.argmin(xy_dist)
-            self.local_station_index.append(j)
+        if len(mesh0d.coordinates.dat.data[:]) > 0:
+            for i in range(self.fs_points_0d.dof_dset.size):
+                # loop over local DOFs and match coordinates to observations
+                # NOTE this must be done manually as VertexOnlyMesh reorders points
+                x_mesh, y_mesh = mesh0d.coordinates.dat.data[i, :]
+                xy_diff = xy - numpy.array([x_mesh, y_mesh])
+                xy_dist = numpy.sqrt(xy_diff[:, 0]**2 + xy_diff[:, 1]**2)
+                j = numpy.argmin(xy_dist)
+                self.local_station_index.append(j)
 
-            x, y = xy[j, :]
-            t = self.observation_time[j]
-            v = self.observation_values[j]
-            x_mesh, y_mesh = mesh0d.coordinates.dat.data[i, :]
+                x, y = xy[j, :]
+                t = self.observation_time[j]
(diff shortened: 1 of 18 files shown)


stephankramer:
Looks good. Just to clarify: if you use `update_forcings` the time-level at which it is evaluated depends on your choice of timestepper (and can be multiple times per timestep for multi-stage timesteppers). For CrankNicolson it appears to be coded to always be at `t+dt` (but should probably be at `t+theta*dt` instead!). I'm not that familiar with the inversion_tools, but I presume that you are now using `export_func` is because you want it to line up with the output of the `TimeSeriesCallback` of the synthetic truth? I guess in the general case, you would have date at an independent arbitrary time interval, but still want to call eval_cost_function every timestep - note that this only works with `export_func` if indeed `t_export=dt`

cpjordan:
@stephankramer: that makes sense as to why there was a dt time lag on the channel inversion example before. I had changed from using update_forcings to export_func as the model timestep that was exported to calculate the loss was out by dt. As the evaluation procedure changes depending on the timestepper it looks like we might need to be more careful. It sounds like I need to look at this again w.r.t. export_func as I thought you might be able to set it so that it occured at dt rather than the export interval. 

Another reason for doing this is so that we have an example where we do calibration where we want to define the loss function separately from the forcing function (though I'm sure you could merge into one) - basically following on from this example: https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022MS003169

cpjordan:
@stephankramer - I've looked over this, this morning and tested for the majority of timesteppers. In the current configuration the time lag exists for all of them because in the inversion tools we have:

```
def cost_fn(t):
    misfit = self.sta_manager.eval_cost_function(t)
    self.J_misfit += misfit
    self.J += misfit
```

Which passes the wrong timestep when we advance (timeintegrator - line 238 will pass `t + dt`/ `t + theta*dt` , instead of `t`). This means when we evaluate the loss we have a timing mismatch (inversion tools, Station Manager):

```
def eval_cost_function(self, t):
        ...
        self.simulation_time.append(t)
        # evaluate observations at simulation time and stash the result
        obs_func = fd.Function(self.fs_points_0d)
        obs_func.dat.data[:] = self.eval_observation_at_time(t)
        ...
```

My fix enforces that the time is correct and lines up with the field being passed:

```
def cost_fn():
    t = solver_obj.simulation_time
    misfit = self.sta_manager.eval_cost_function(t)
    self.J_misfit += misfit
    self.J += misfit
```

This works independently of the timestepper being used. I use `export_func` because if we use `update_forcings` it passes the time to the cost function leading to:

`TypeError: cost_fn() takes 0 positional arguments but 1 was given`

We could instead just accept the time and not use it e.g. `def cost_fn(_)` to make this work with `update_forcings`. However, I think using `export_func` is fine because if we can use a time dependent condition for any other exporting we want to do in either `update_forcings` or `export_func`? But perhaps something worth noting on the example?

stephankramer:
Just to be clear: I'm completely fine with what you propose. I just tried to explain why `update_forcings` is _not_ designed to do these sort of time-integrated diagnostics, it's there for user code to evaluate _inputs_ to the model at times prescribed by Thetis (timestepper). `export_func` is not entirely optimal either because it's really there to be called at the frequence of h5 checkpoints/vtu outputs - and for instance it will depend on how you've set `export_initial_state` and forces you to set `t_export=dt` which might not be no ideal for large models, writing entire solution fields every timestep. The functionality that Thetis does provide to evaluate diagnostics every timestep is really through a custom DiagnosticCallback which you'd add via `solver.add_callback(callback, eval_interval='timestep')` - they are just a bit more cumbersome to write.

cpjordan:
Ah, I slightly misunderstood - thanks for elaborating. I'm happy to add a DiagnosticCallback to the model config that can be called instead of using `export_func`, or otherwise I think it can be merged as it is if you're happy with it.

thetis/inversion_tools.py
```diff
@@ -487,26 +488,29 @@ def construct_evaluator(self):
...
-            v = self.observation_values[j]
-            x_mesh, y_mesh = mesh0d.coordinates.dat.data[i, :]
-
-            msg = 'bad station location ' \
-                f'{j} {i} {x} {x_mesh} {y} {y_mesh} {x-x_mesh} {y-y_mesh}'
-            assert numpy.allclose([x, y], [x_mesh, y_mesh]), msg
-            # create temporal interpolator
-            ip = interp1d(t, v, **interp_kw)
-            self.station_interpolators.append(ip)
+        if len(mesh0d.coordinates.dat.data[:]) > 0:
```
cpjordan, line 491: In parallel, you can have mesh partitions which don't contain any of the Station locations `xy`, so the corresponding processor will hang because `mesh0d` (L474) doesn't have any coordinates (and therefore size). Adding a check that the length is greater than 0 allows any processors without stations in their mesh partition to pass.

stephankramer: That's a little odd - do you mean it's hanging in evaluating `self.fs_points_0d.dof_dset.size`? I would have thought that should just evaluate to 0 in that case, and then it should automatically skip the loop. Weird, but anyway....

thetis/inversion_tools.py
```diff
@@ -487,26 +488,29 @@ def construct_evaluator(self):
...
+                x_mesh, y_mesh = mesh0d.coordinates.dat.data[i, :]
+
+                msg = 'bad station location ' \
+                    f'{j} {i} {x} {x_mesh} {y} {y_mesh} {x-x_mesh} {y-y_mesh}'
+                assert numpy.allclose([x, y], [x_mesh, y_mesh]), msg
+                # create temporal interpolator
+                ip = interp1d(t, v, **interp_kw)
+                self.station_interpolators.append(ip)
+        else:
+            pass
```
stephankramer, line 513: Not a fan of adding a spurious else case

stephankramer (dismissed):
Great, yes I think using `export_func` is pretty reasonable for now - and in any case this is an improvement over using `update_forcings` if there was mismatch indeed

thetis/inversion_tools.py
cpjordan: I've found that if I add this print statement anywhere after the creation of `mesh0d`:

`print(len(mesh0d.coordinates.dat.data[:]))`

It will also advance (without needing to add the `if` statement I've added here). Otherwise it hangs once you get to the loop. I can do a `print_output('Here...')` statement up to the `for` loop but not afterwards. I can't do `print('Here...')` statements at any point once it's reached `construct_evaluator` without either 'fix'. Pretty strange.

I've checked this on both my HP and Dell machine and a colleagues - so not unique to a Firedrake install or machine as far as I can tell.

thetis/inversion_tools.py
cpjordan: I'll remove the unnecessary `else` and then merge - thank you! I'll update to adding a `DiagnosticCallback` instead of `export_func` for my other PR..!

thetis/inversion_tools.py
stephankramer: Ah ok, so I suspect what's happening is that whenever the `.dat` of that `mesh.coordinates` is first accessed, there is some initialisation code that needs to happens _collectively_ (i.e. on all processes). Without the if statement, on a process where `self.fs_points_0d.dof_dset.size` is zero, the .dat never gets accessed, and so the other processes will be waiting forever for that process to participate. But having the .dat being accessed in the `if` statement apparently is already sufficient to ensure it is being accessed on all processes. So that seems a perfectly good solution.

stephankramer (approved):

PR #373, opened 2024-07-30
Replace .at() interpolation in 2D callbacks

cpjordan:
.at() raises an error when the number of processors used is large, use a VertexOnlyMesh instead for interpolation


cpjordan:
Prevents the Firedrake issue [#2790](https://github.com/firedrakeproject/firedrake/issues/2790) whilst Firedrake issue [#3080](https://github.com/firedrakeproject/firedrake/issues/3080) is waiting to be resolved. This method is slower than `.at()` so an alternative is to add a try except with `.at()` before moving onto using a VertexOnlyMesh.

There are some remaining direct applications of `.at()` but they are for 3D examples where either the callbacks are only being used in serial or where the problem is so large that `.at()` is functioning fine.

thetis/callback.py
```diff
@@ -673,6 +694,7 @@ def __init__(self, solver_obj, fieldnames, x, y,
         self.tolerance = tolerance
         self.eval_func = eval_func
         self._initialized = False
+        self.bathymetry_val = None
```
tkarna, line 697: There is no need to store the bathymetry value

thetis/callback.py
```diff
@@ -684,10 +706,31 @@ def _initialize(self):
...
             if self.eval_func is None:
-                self.solver_obj.fields.bathymetry_2d.at(self.xyz, tolerance=self.tolerance)
+                bathymetry_field = self.solver_obj.fields.bathymetry_2d
+                P0DG = FunctionSpace(self.vom, "DG", 0)
+                P0DG_input_ordering = FunctionSpace(self.vom.input_ordering, "DG", 0)
+                f_at_points = Function(P0DG)
+                f_at_input_points = Function(P0DG_input_ordering)
+                f_at_points.interpolate(bathymetry_field)
+                f_at_input_points.interpolate(f_at_points)
+                self.bathymetry_val = f_at_input_points.dat.data_ro[:]
```
tkarna, line 733: remove

thetis/callback.py
```diff
@@ -684,10 +706,31 @@ def _initialize(self):
...
         try:
             if self.eval_func is None:
-                self.solver_obj.fields.bathymetry_2d.at(self.xyz, tolerance=self.tolerance)
+                bathymetry_field = self.solver_obj.fields.bathymetry_2d
+                P0DG = FunctionSpace(self.vom, "DG", 0)
+                P0DG_input_ordering = FunctionSpace(self.vom.input_ordering, "DG", 0)
+                f_at_points = Function(P0DG)
+                f_at_input_points = Function(P0DG_input_ordering)
+                f_at_points.interpolate(bathymetry_field)
+                f_at_input_points.interpolate(f_at_points)
```
tkarna, line 732: fyi iirc this is just to check that the interpolation works (points aren't outside the domain) and is included in initialization to fail early rather than waiting for the first usage (which typically only happens after some iterations). Any field could be interpolated, and the result can be discarded

thetis/callback.py
```diff
@@ -529,6 +529,20 @@ def __init__(self, solver_obj,
         self.field_names = field_names
         self._name = name
 
+        # create VertexOnlyMesh and function spaces for interpolation
+        self.mesh = solver_obj.fields[field_names[0]].function_space().mesh()
```
tkarna, line 533: or just `solver_obj.mesh2d` ?

thetis/callback.py
```diff
@@ -529,6 +529,20 @@ def __init__(self, solver_obj,
...
+        self.function_spaces = {}
+        for field_name in field_names:
+            field = solver_obj.fields[field_name]
+            if isinstance(field.function_space().ufl_element(), VectorElement):
+                P0DG = VectorFunctionSpace(self.vom, "DG", 0)
+                P0DG_input_ordering = VectorFunctionSpace(self.vom.input_ordering, "DG", 0)
+            else:
+                P0DG = FunctionSpace(self.vom, "DG", 0)
+                P0DG_input_ordering = FunctionSpace(self.vom.input_ordering, "DG", 0)
+            self.function_spaces[field_name] = (P0DG, P0DG_input_ordering)
```
tkarna, line 544: minor: there's quite a lot of duplicated code in the vom creation and evaluation. Could plausibly be refactored in to helper functions for example

thetis/callback.py
```diff
@@ -554,7 +569,13 @@ def message_str(self, *args):
             for name, values in zip(self.detector_names, args))
 
     def _evaluate_field(self, field_name):
-        return self.solver_obj.fields[field_name](self.detector_locations)
+        field = self.solver_obj.fields[field_name]
+        P0DG, P0DG_input_ordering = self.function_spaces[field_name]
+        f_at_points = Function(P0DG)
+        f_at_input_points = Function(P0DG_input_ordering)
```
tkarna, line 575: These function are created on every evaluation. Instead of storing the function spaces in `self.function_spaces` you could just store the functions themselves.

thetis/callback.py
```diff
@@ -707,7 +750,12 @@ def __call__(self):
             try:
                 field = self.solver_obj.fields[fieldname]
                 if self.eval_func is None:
-                    val = field.at(self.xyz, tolerance=self.tolerance)
+                    P0DG, P0DG_input_ordering = self.function_spaces[fieldname]
+                    f_at_points = Function(P0DG)
+                    f_at_input_points = Function(P0DG_input_ordering)
+                    f_at_points.interpolate(field)
+                    f_at_input_points.interpolate(f_at_points)
```
tkarna, line 757: Do you know if this vom evaluation works for 2D mesh on the sphere (in which case the point coordinates are 3d x,y,z)? The whole `eval_func` option was added to enable `at()`-like evaluation in this case (`eval_func` was a modified version of Firedrake's `at`)

tkarna (changes requested):
Thanks, looks good. Only minor comments

thetis/callback.py
cpjordan: Ah, that makes sense - thanks

thetis/callback.py
cpjordan: I've created a new function in utility.py [L1159](https://github.com/thetisproject/thetis/pull/373/files#diff-a2f80a5edab6cf135404e407d32a7cce2cbbd4e911b63dc1b13dce5e65e50ee1R1159): `vom_interpolator_function`. It generates and returns the Functions only as suggested so we remove all the unnecessary Function duplication I had. We could probably go a step further and create a vom_interpolator class which also performs the interpolation as well rather than repeating these lines:

```
f_at_points, f_at_input_points = self.interp_functions[field_name]
f_at_points.interpolate(field)
f_at_input_points.interpolate(f_at_points)
return f_at_input_points.dat.data_ro[:]
```

Though maybe it's fine how it is for now. When [Firedrake issue 3080](https://github.com/firedrakeproject/firedrake/issues/3080) is dealt with we can probably revert to `.at()` anyway.

thetis/callback.py
cpjordan: This is much more sensible

thetis/callback.py
cpjordan: Unsure as I only usually do boring plain 2D meshes! I mainly fixed because of the parallelisation issues for big simulations. Happy to look into it in the future or perhaps someone in the Firedrake group might have answer for this already.

cpjordan:
The current version of Firedrake is failing, which seems to be causing all the Thetis tests to fail. Waiting for this to be updated.

thetis/callback.py
tkarna: Thanks, this is good for now

thetis/callback.py
tkarna: OK, we can address it in another PR.

tkarna (approved):
Thanks!

PR #374, opened 2024-08-08
Fix NonHydrostaticTimeIntegrator2D for projection of solutions with adaptivity

stephankramer:
The advance method of a timeintegrator should always advance from the solved-for-fields, which are the same fields that will contain the solution at the end of the timestep _after_ calling advance(). This is because this is the only field(s) that the "user" of a time integrator has control over, and the user may want to adjust/overwrite/project (for adaptivity) at the end of a timestep and expect the timestepper to advance from the adjusted values in the next timestep.

Thus, if a time-integrator stores some copy of a "solution_old" it should be set at the _beginning_ of advance() to the current solution, rather than at the end.


cpjordan (approved):
This update ensures that `elev_old` should reflect any updates that the user makes to `elev_2d` between timesteps. Makes sense for adaptivity purposes and shouldn't impact otherwise as verified by running one of the non-hydrostatic examples. Looks good to me!

PR #375, opened 2024-10-17
Remove pytest cache clearing

cpjordan:
- This was updated in Firedrake PR #3730


cpjordan:
[Firedrake PR #3370](https://github.com/firedrakeproject/firedrake/pull/3730) and [PyOP2 PR #724](https://github.com/OP2/PyOP2/pull/724) updated the caching mechanisms which caused the hooks in test/conftest.py and test_adjoint/conftest.py to fail. These hooks were originally introduced in PR #2 to reduce memory usage and also prevent errors caused by inter-test dependencies.

With the recent caching updates in Firedrake and PyOP2, these issues should be resolved, so the cache-clearing hooks are no longer necessary, or at least until further issues arise.

stephankramer:
This looks good wrt tsfc/pyop2 cache but I would leave the clearing of the adjoint tape in place as it is. If you don't clear the tape between tests that use the adjoint then the model annotated in each tests just gets appended. This is a massive memory hole as you end up having the annotated forward model of each test in memory simultaneously. It potentially also slows  down things, as a replay/backward pass may (depending on how clever pyadjoint is being)  end up going through the entire tape including all operations of previous tests.

tkarna (approved):
Thanks, looks good to me

PR #376, opened 2024-11-28
Fix CI errors

cpjordan:
Fix various issues with testing:

- ufl_element() -> ufl_function_space()
- Fatal Python error: Aborted - TO BE FIXED
- AttributeError: 'Cofunction' object has no attribute '_ufl_class_' - TO BE FIXED

### DIFF
--- a/examples/discrete_turbines/turbine_callback.py
+++ b/examples/discrete_turbines/turbine_callback.py
@@ -32,7 +32,7 @@ class TidalPowerCallback(DiagnosticCallback):
         self.field_dims = []
         for field_name in field_names:
             if field_name != 'pow':
-                self.field_dims.append(solver_obj.fields[field_name].function_space().block_size)
+                self.field_dims.append(solver_obj.fields[field_name].function_space().value_size)
         attrs = {
             # use null-padded ascii strings, dtype='U' not supported in hdf5,
             # see http://docs.h5py.org/en/latest/strings.html
--- a/examples/discrete_turbines/turbine_callback.py
+++ b/examples/discrete_turbines/turbine_callback.py
@@ -32,7 +32,7 @@ class TidalPowerCallback(DiagnosticCallback):
         self.field_dims = []
         for field_name in field_names:
             if field_name != 'pow':
-                self.field_dims.append(solver_obj.fields[field_name].function_space().value_size)
+                self.field_dims.append(solver_obj.fields[field_name].function_space().block_size)
         attrs = {
             # use null-padded ascii strings, dtype='U' not supported in hdf5,
             # see http://docs.h5py.org/en/latest/strings.html
--- a/thetis/callback.py
+++ b/thetis/callback.py
@@ -508,7 +508,7 @@ class DetectorsCallback(DiagnosticCallback):
         """
         # printing all detector output to log is probably not a useful default:
         kwargs.setdefault('append_to_log', False)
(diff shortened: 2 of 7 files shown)


thetis/callback.py
```diff
@@ -996,7 +996,7 @@ def __init__(self, solver_obj, fieldnames, x, y,
         self.field_dims = {}
         for f in self.fieldnames:
             func = solver_obj.fields[f]
-            self.field_dims[f] = func.function_space().value_size
+            self.field_dims[f] = func.function_space().block_size
```
stephankramer, line 999: Same here

thetis/exporter.py
```diff
@@ -22,11 +22,11 @@ def get_visu_space(fs):
     """
     mesh = fs.mesh()
     family = 'Lagrange' if is_cg(fs) else 'Discontinuous Lagrange'
-    if len(fs.ufl_element().value_shape) == 1:
-        dim = fs.ufl_element().value_shape[0]
+    if len(fs.ufl_function_space().value_shape) == 1:
```
stephankramer, line 25: Does `fs.value_shape` not work?

thetis/callback.py
```diff
@@ -508,7 +508,7 @@ def __init__(self, solver_obj,
         """
         # printing all detector output to log is probably not a useful default:
         kwargs.setdefault('append_to_log', False)
-        self.field_dims = [solver_obj.fields[field_name].function_space().value_size
+        self.field_dims = [solver_obj.fields[field_name].function_space().block_size
```
stephankramer, line 511: I'm a little confused tbh. I would have expected this should be `value_size` as that would correspond to the size of what you get when interpolating in a point (i.e. the size of the value) - which is not necessarily the same as the n/o DOFs per node which is what `block_size` gives IIUC. I know that your change should give back the old behaviour. Might it be that `DetectorsCallback` just has not been tested with RT or BDFM velocity fields?

examples/discrete_turbines/turbine_callback.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, solver_obj,
         self.field_dims = []
         for field_name in field_names:
             if field_name != 'pow':
-                self.field_dims.append(solver_obj.fields[field_name].function_space().value_size)
+                self.field_dims.append(solver_obj.fields[field_name].function_space().block_size)
```
stephankramer, line 35: See comment for DetectorsCallback

stephankramer (commented):
Thanks @cpjordan 
So it might be I'm just misunderstanding - but I would have expected that for interpolated value size we should in fact be using `value_size` and not `block_size`. The latter being the n/o DOFs per node, which was what previously returned by `value_size` in which case it would in fact have been incorrect in Thetis previously. The difference would only appear for RT and BDM velocity fields which we don't use very frequently. I do agree/understand the changes to the OP2 kernel headers in utility.py and utility3d.py - these should indeed be `block_size` now.

stephankramer:
Related to https://github.com/firedrakeproject/firedrake/pull/3862 and https://github.com/firedrakeproject/firedrake/pull/3362

cpjordan:
@stephankramer - thanks for reviewing and explanation. I had discussed the problems with @connorjward who found that `value_size` in the utility kernels was the cause of the failing tests, but I just replaced all instances assuming that the previous implementation in Thetis was correct - I should have looked into it more!

Updated based on the comments above.

stephankramer (approved):
Perfect, many thanks!

PR #386, opened 2025-01-15
Print duration of all pytests in CI

stephankramer:


cpjordan:
It seems like none of the test times now show up on the log after adding the `--durations-min=60.0` flag. Previously there were various tests not within [test/bottomFriction](https://github.com/thetisproject/thetis/tree/master/test/bottomFriction) that took longer than 60s, but these aren't showing up e.g.: https://github.com/thetisproject/thetis/actions/runs/13389397497/job/37393428304#step:6:1537.

I've checked that each worker had a test that is longer than 60s, so I don't think it is a parallelism issue. I can look at testing some other stuff later (provided I'm not missing something obvious!).

stephankramer:
Ah thanks. Looks like I need `--durations=0` _in addition_ to `--durations-min=60.0`

cpjordan (approved):
Nice! This should be really useful in the future.

PR #387, opened 2025-01-15
#368 modify iterate method to allow timestep control

### LINKED ISSUE #368: Allow single timestep control over Thetis run
This issue has come up coupling Thetis to a goal-based mesh-adaptivity framework [goalie](https://github.com/mesh-adaptation/goalie/) but would equally be important coupling Thetis to other models via a coupling framework: Thetis only provides per timestep control via callbacks, which keeps Thetis in charge of runtime, and does not cater for an external framework to request a single timestep to be performed.

Two potential implementations that I can think of:
1. Simply separate the timestep code into a separate (user callable) method
2. Use a generator approach.
So change the current code
```
   def iterate(self, ...):
       <prep-stuff>
       while t<t_end:
           <timestep-code>
```
to either 1:
```
   def advance(self):
        <timestep-code>

   def prep-stuff(self):
       <prep_stuff>

   def iterate(self, ...):
      self.prep_stuff()
       while t<t_end:
          self.advance()
```
or  2:
```
   def iterate(self, ...):
        for t in self.generator(...):
             pass

   def generator(self, ...):
       <prep-stuff>
       while t<t_end:
          <timestep-code>
          yield t
```
where in both cases the user can use `solver_obj.iterate()` in exactly the same way as before.

To get per time-step control, in case 1, the user (coupling framework) would have to do three things
```
    solver_obj.prep_stuff()
    while t<t_end:  # i.e. repeat the timestep logic
        solver_obj.advance()
        # user code
```
whereas in case 2, that would simply be:
```
   for t in solver_obj.generator():
        # user code
```
or if there's some further logic, or other looping requirements:
```
   thetis_timestepper = solver_obj.generator()
   while t_Thetis<t_end and .... :
      t_Thetis = thetis_timestepper.next()
```

So at the moment I'm inclined to go for option 2.

acse-ej321:
closes #368 

Implements the 2nd option proposed by @stephankramer in #368 by splitting `iterate` into two parts:
 - `create_iterator` - which gives a facelift refactor to the existing `iterate` functionality as a generator
 - `iterate` - which wraps `create_iterator` and preserves the existing accessibility/structure

### DIFF
--- a/thetis/solver2d.py
+++ b/thetis/solver2d.py
@@ -947,7 +947,8 @@ class FlowSolver2d(FrozenClass):
         """
         Runs the simulation
 
-        Iterates over the time loop until time ``options.simulation_end_time`` is reached.
+        Wrapper for function for `create_iterator` generator and automatically
+        iterates over the time loop until time ``options.simulation_end_time`` is reached.
         Exports fields to disk on ``options.simulation_export_time`` intervals.
 
         :kwarg update_forcings: User-defined function that takes simulation
@@ -956,6 +957,42 @@ class FlowSolver2d(FrozenClass):
         :kwarg export_func: User-defined function (with no arguments) that will
             be called on every export.
         """
+        for _ in self.create_iterator(update_forcings=update_forcings,
+                export_func=export_func):
+             pass
+            
+    @PETSc.Log.EventDecorator("thetis.FlowSolver2d.create_iterator")
+    def create_iterator(self, update_forcings=None,
+                export_func=None):
+        """
+        Creates a generator to iterate through the simulation and return access 
+        to time advancing function when time control is handled eternally.
+
+        Iterates over the time loop until time ``options.simulation_end_time`` is reached.
+        Exports fields to disk on ``options.simulation_export_time`` intervals.
+
+        For example:
+
+        .. code-block:: python
+
+            for t in solver_obj.generator():
+                # user code
+
+        or, to get per time-step control:
+                
+        .. code-block:: python
+
+            thetis_timestepper = solver_obj.generator()
+            while t_Thetis<t_end and .... :
+                t_Thetis = next(thetis_timestepper)
+
+        :kwarg update_forcings: User-defined function that takes simulation
+            time as an argument and updates time-dependent boundary conditions
+            (if any).
+        :kwarg export_func: User-defined function (with no arguments) that will
+            be called on every export.
+        """
+
         if not self._initialized:
             self.initialize()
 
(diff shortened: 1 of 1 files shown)


stephankramer (changes requested):
Perfect! The only other thing I was going to ask is to use it in one of the examples or tests - just so that the `create_iterator()` is explicitly called somewhere in CI

demos/demo_2d_tracer.py
```diff
@@ -87,6 +89,7 @@
 
 # The velocity field is set up using a simple analytic expression. ::
 
+x, y = SpatialCoordinate(mesh2d)
```
stephankramer, line 92: Hm, that's a good demo to showcase the use of the iterator but it brings up another issue. Apparently the demo's aren't tested - demonstrated by the fact that you had to add this line - which is bad! See https://github.com/thetisproject/thetis/issues/389. The reason I asked for the `create_iterator()` method to be used in a example or test, is precisely because we intend it to be public API so we want to know when it breaks.

acse-ej321: I modified `test_consistency_2d.py` to use the `create_iterator()` function for inclusion in test coverage in https://github.com/thetisproject/thetis/pull/387/commits/cab045bf937274ed7381a01e4595f0fe064b5e09. Additionally, I moved the checks initially preceding the solve to be included in the time stepping as per the `TODO` in the file.

test/tracerEq/test_consistency_2d.py
```diff
@@ -80,7 +81,16 @@ def run_tracer_consistency(constant_c=True, **model_options):
...
+
+    # time stepping
+    thetis_timestepper = solver_obj.create_iterator()
+    t_Thetis = solver_obj.simulation_time  # initial time
+    t_epsilon = 1.0e-5
+    while t_Thetis <= t_end - t_epsilon:
+        try:
+            t_Thetis = next(thetis_timestepper)
+        except StopIteration:
+            break
```
stephankramer, line 93:
```suggestion
    t_epsilon = 1.0e-5
    while True:
        try:
            t_Thetis = next(thetis_timestepper)
        except StopIteration:
            break
    assert t_Thetis>=t_end - t_epsilon
```
The loop you had either hits the StopIteration, or the while loop criterion - so one of them is redundant. I presume it's meant to hit the StopIteration. The assert that I suggest is then a way to check that it stops when we expect it.

stephankramer (changes requested):
Great, nicely tested now. Just a final few minor nitpicks

stephankramer (approved):
Great, thanks Eleda

PR #390, opened 2025-02-04
Add projected area option for turbines

cpjordan:
- Concentrates the thrust into a smaller (or larger) area, allowing a narrower or wider wake for the same applied thrust
- Swept diameter defaults to normal diameter
- No changes to the examples

Based on [previous feature of OpenTidalFarm](https://github.com/OpenTidalFarm/OpenTidalFarm/blob/master/opentidalfarm/turbines/thrust_turbine.py).

### DIFF
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -462,8 +462,8 @@ class TidalTurbineOptions(FrozenHasTraits):
     name = 'Tidal turbine options'
     diameter = PositiveFloat(
         18., help='Turbine diameter').tag(config=True)
-    swept_diameter = PositiveFloat(
-        None, allow_none=True, help='Projected diameter of the turbine '
+    projected_diameter = PositiveFloat(
+        None, allow_none=True, help='Extent over which turbine drag is applied'
                                     '(defaults to diameter if not provided)').tag(config=True)
     C_support = NonNegativeFloat(
         0., help='Thrust coefficient for support structure').tag(config=True)
--- a/thetis/options.py
+++ b/thetis/options.py
@@ -462,6 +462,9 @@ class TidalTurbineOptions(FrozenHasTraits):
     name = 'Tidal turbine options'
     diameter = PositiveFloat(
         18., help='Turbine diameter').tag(config=True)
+    swept_diameter = PositiveFloat(
+        None, allow_none=True, help='Projected diameter of the turbine '
+                                    '(defaults to diameter if not provided)').tag(config=True)
     C_support = NonNegativeFloat(
         0., help='Thrust coefficient for support structure').tag(config=True)
     A_support = NonNegativeFloat(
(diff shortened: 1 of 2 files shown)


thetis/options.py
```diff
@@ -462,6 +462,9 @@ class TidalTurbineOptions(FrozenHasTraits):
     name = 'Tidal turbine options'
     diameter = PositiveFloat(
         18., help='Turbine diameter').tag(config=True)
+    swept_diameter = PositiveFloat(
+        None, allow_none=True, help='Projected diameter of the turbine '
```
stephankramer, line 466: I'm not sure "projected diameter" is immediately clear. Maybe something like "extent over which turbine drag is applied" ?

thetis/turbines.py
```diff
@@ -14,22 +14,31 @@
...
         self.A_support = options.A_support
         self.upwind_correction = upwind_correction
 
+        # Check that the parameter choices make some sense - these won't break
+        # the simulation but may give unexpected results if the choice isn't
+        # understood.
+        if self.swept_diameter != self.diameter:
+            log(INFO, 'Warning - swept_diameter and plan_diameter are not equal. '
+                      'Upwind correction may not be accurate.')
+
```
stephankramer, line 28: This might be a bit confusing: isn't the whole point of the swept_diameter option to be able to choose it differently from the diameter? Also there's no `plan_diameter` option

thetis/turbines.py
```diff
@@ -14,22 +14,31 @@
...
     def _thrust_area(self, uv):
         C_T = self.thrust_coefficient(uv)
-        A_T = pi * self.diameter**2 / 4
-        fric = C_T * A_T
+        A_T = pi * self.swept_diameter**2 / 4
+        swept_area_factor = self.diameter**2/self.swept_diameter**2
```
stephankramer, line 32: what's the point of this `swept_area_factor`? You're multiplying and then dividing out the swep_diameter. So I think this code can remain just like it was before (i.e. using diameter only) which I think is much clearer - the intent is to apply an integrated force F=0.5 C_T A_T with A_T=pi diameter**2/4 regardless over which area this force has been smeared out over.

stephankramer (changes requested):
Some minor comments. Looks useful!

thetis/options.py
cpjordan: Yep, agree - and personally I don't like the term "swept" diameter - I think projected would be better with the updated description. Happy to go with whatever you think is best though!

thetis/turbines.py
cpjordan: Sorry, you're absolutely right of course - that was just because of my thought process when changing it (projecting to a smaller/larger area -> then amplify by a factor to keep correct thrust). Fixed!

stephankramer (approved):
Looks good, thanks!

PR #392, opened 2025-02-20
pip install firedrake support

connorjward:
Hi, I have had a go at getting `pip install Firedrake` (https://github.com/firedrakeproject/firedrake/pull/4011) support added to your CI (and also made you a `pyproject.toml`).

The website will also need updating.

Note that I haven't yet figured out how to get versioneer working properly. If you feel strongly about keeping it then some extra hacking around is needed. I have simply removed references to it in the build process and set the version to 0.0.1 for now.

PR #395, opened 2025-03-17
Test with updated containers

connorjward:
Note that this PR sits on top of https://github.com/thetisproject/thetis/pull/392

We are changing to having `root` as the default user instead of `firedrake` and I want to ensure that this works for users.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,17 +18,13 @@ jobs:
     runs-on: [self-hosted, Linux]
     # The docker container to use.
     container:
-      image: firedrakeproject/firedrake-vanilla-default:latest
+      # UNDO ME
+      # image: firedrakeproject/firedrake-vanilla-default:latest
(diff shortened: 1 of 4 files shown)


connorjward:
> @connorjward just for my understanding - `/home/firedrake` no longer exists in the Firedrake container so this just moves where cache files go? At the moment `/home/firedrake` will be generated by calling `echo "HOME=/home/firedrake" >> $GITHUB_ENV"` and it's only used for caching or config files, but better to just allow these to go directly to `root`. Anything else to note?

Yeah it's just tidying things up a bit. `$HOME` for `root` is `/root` not `/home/firedrake`.

cpjordan (approved):
Thanks!

cpjordan:
Test failure unrelated to PR - linked to change in [ufl #371](https://github.com/FEniCS/ufl/pull/371)

PR #399, opened 2025-04-04
Update installation instructions

cpjordan:
Updates the website installation instructions (change due to Firedrake [#4011](https://github.com/firedrakeproject/firedrake/pull/4011)). 

Also added:
- the symbols in for the webpage to be consistent with Firedrake
- optional documentation build dependencies


docs/source/download.rst
```diff
@@ -9,45 +9,68 @@ Thetis requires installation of `Firedrake
...
+Check out the `Thetis repository <http://github.com/thetisproject/thetis>`_ from GitHub.
+You then need to add the Thetis repository to your ``PYTHONPATH`` in the Firedrake virtualenv. You can do this with ``pip``:
+
+.. code-block:: none
 
-Check out the `Thetis <http://github.com/thetisproject/thetis>`_
-repository from Github.
-You then need to add the Thetis repository to your ``PYTHONPATH`` in the
-Firedrake virtualenv. You can do this with ``pip``::
+   pip install -e <path-to-thetis-repository>
```
stephankramer, line 75: I would just remove this entire last section. It was basically suggesting an (editable) pip install in contrast with an install using `firedrake-install --install thetis` which was recommended as the normal way. Now however, we always `pip install` separately _after_ firedrake install - so we don't need separate instructions for shared systems where firedrake is installed already. Also, the line about ``PYTHONPATH`` is confusing as it then recommends `pip`.

TBH I don't really know how to setup a shared firedrake install in such a way that people can still `pip install` on top of it (without affecting other users) but in any case it doesn't matter whether it's editable or not it still puts something in site-packages (just a reference to the editable install).

stephankramer (changes requested):
Great, some minor comments....

docs/source/download.rst
cpjordan: Yep, I've replaced this stuff with a reference to Docker installation and contact details for if a user gets stuck with installation. Or I can drop this new stuff altogether if you think it's not worth adding!

docs/source/download.rst
```diff
@@ -9,45 +9,67 @@ Thetis requires installation of `Firedrake
...
 within the Firedrake virtual environment.
 
-Install Firedrake and Thetis
------------------------------
+Installing Firedrake
+---------------------
 
-You can install both Firedrake and Thetis by running::
+You can install Firedrake by following the download documentation on the
+`Firedrake website <http://firedrakeproject.org/download.html>`_.
```
connorjward, line 16:
```suggestion
`Firedrake website <http://firedrakeproject.org/install.html>`_.
```

docs/source/download.rst
```diff
@@ -9,45 +9,67 @@ Thetis requires installation of `Firedrake
...
 
-    curl -O https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install
-    python3 firedrake-install --install thetis
+.. note::
 
-See  `Firedrake website <http://firedrakeproject.org/download.html>`_ for more
-information on the installation procedure. Note that the install proceduce may
-take up to one hour depending on your system.
+   **Installing Firedrake can take up to an hour depending on the system.**
```
connorjward, line 20: But this isn't true! It's much much faster now. Maybe 20 minutes on a slow non-Ubuntu computer.

docs/source/download.rst
```diff
@@ -9,45 +9,67 @@ Thetis requires installation of `Firedrake
...
+An IDE such as PyCharm will not recognize Thetis when installed in this fashion for any project outside the cloned repository,
+as the source is not in site-packages. It will still run, and if you would like to enable full code navigation, you
+can add the Thetis cloned repository as a content root, then add the ``thetis`` sub-directory as a sources root.
+
+.. _alternative-installation-methods:
+
+Alternative installation methods
+=================================
+
+As well as being installable through ``pip``, Firedrake also provides Docker containers which typically contain the
```
connorjward, line 69: This isn't currently true: https://github.com/firedrakeproject/firedrake/blob/master/docker/Dockerfile.firedrake#L12

I will get that fixed.

connorjward: https://github.com/firedrakeproject/firedrake/pull/4225

docs/source/download.rst
connorjward: Just you wait for https://github.com/firedrakeproject/firedrake/pull/4122! We are building PETSc in under 2 minutes

stephankramer (approved):
Looks good to me now. Thanks @connorjward for all the updates, happy with the install instructions now?

docs/source/download.rst
```diff
@@ -9,45 +9,67 @@ Thetis requires installation of `Firedrake
...
 
-Using this install method you should
-**not** add Thetis to your ``PYTHONPATH``. Instead, Thetis will
-automatically be available to import whenever your Firedrake
-virtualenv is active.
+.. warning::
 
-If you have already installed Firedrake
----------------------------------------
+   **Please check that Firedrake is working before trying to install Thetis.**
```
connorjward, line 29: Instead of this I would probably say something above like: "You should check that the Firedrake install has been successful by running the command `firedrake-check`."

docs/source/download.rst
```diff
@@ -9,45 +9,67 @@ Thetis requires installation of `Firedrake
...
-You then need to add the Thetis repository to your ``PYTHONPATH`` in the
-Firedrake virtualenv. You can do this with ``pip``::
+You can install Thetis in your Firedrake installation by activating the Firedrake virtualenv and running:
+
+.. code-block:: none
+
+   pip install git+https://github.com/thetisproject/thetis.git
+
+Using this install method, you should **not** add Thetis to your
+``PYTHONPATH``. Instead, Thetis will automatically be available to import whenever your Firedrake virtualenv is active.
```
connorjward, line 46: This more of a personal preference. But I would avoid mentioning this and also the warning about `pip install thetis` above. It makes it more confusing to follow what the actual instructions are.

docs/source/download.rst
```diff
@@ -9,45 +9,67 @@ Thetis requires installation of `Firedrake
...
+.. _editable-install:
+
+Editable install
+=============================================================================
+
+If you want to install Thetis from a local checkout of the repository that you can directly edit, update (pull) from github, switch branches, etc., it is recommended to use an editable install using:
+
+.. code-block:: none
+
+   git clone https://github.com/thetisproject/thetis
```
connorjward, line 57: Some people will want `git@github.com:thetisproject/thetis`.

connorjward (approved):

PR #401, opened 2025-05-29
Add gradient-based regularisation for inversion

cpjordan:
Create a Gradient based equivalent of the Hessian regularisation for the adjoint. 

- ControlRegularizationManager also needs to be updated so you can specify gradient or Hessian regularisation
- Needs to be added to an example
- You could possibly just merge the recoverers and calculators and specify whether the Hessian should also be calculated
- You might also want to penalise with both the gradient and Hessian


cpjordan:
> Nice! Note that I was planning to upstream gradient and Hessian recovery into Firedrake itself, since it's implemented both in Thetis and in [Animate](https://github.com/mesh-adaptation/animate/blob/main/animate/recovery.py), and would likely be of use to others in the broader Firedrake community. Might you be interested in working together on this?

Sounds good! Sorry for the delay in getting back to you - I'll first check whether others have looked at this in the Firedrake slack channel?

Once done, it would probably be nice for me to revise #371 to show the implementation of all these various regularisations!

cpjordan:
> A few minor suggestions but otherwise looks good.

Thanks - updated

PR #402, opened 2025-06-13
Avoid subspace warning

jwallwork23:
Avoids deprecation warning about using the `subfunctions` property of a `FunctionSpace`.


cpjordan (approved):
Thanks, looks good.

Sorry for delay - I've been on leave for a while!

PR #403, opened 2025-07-04
Update readme and add code coverage test

cpjordan:
This pull request updates the main README and adds a README to the example repo to provide short descriptions for each example in the examples/ directory.

Additionally, this PR adds code coverage.

These changes aim to improve the introduction of the project for new users and make it feel a bit more welcoming.

PR #406, opened 2025-07-11
Add code coverage

cpjordan:
Adds code coverage to Thetis testing. Will rebase the branch once #403 is merged. 

Currently have the code coverage configured with `dry_run` which simulates the coverage upload, without actually uploading it. This needs to be amended if this is a welcome addition so that the code coverage is uploaded. This will need the repo to be connected on https://about.codecov.io/.

Also needs a couple more checks. The latest [check](https://github.com/thetisproject/thetis/actions/runs/16117475269/job/45474636168#step:8:669) suggests that an inversion example runs yet we [report no coverage](https://github.com/thetisproject/thetis/actions/runs/16117475269/job/45474636168#step:8:1601).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
           : # Pass '--system-site-packages' so Firedrake can be found
           python3 -m venv --system-site-packages venv-thetis
           . venv-thetis/bin/activate
-          pip install ./thetis-repo
+          pip install ./thetis-repo[dev]
           pip list
 
       - name: Lint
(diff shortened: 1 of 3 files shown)


cpjordan:
Cannot accurately add code coverage due to facing the issue described in https://github.com/pytest-dev/pytest-cov/issues/277. I find that everything works locally but as soon as I move into a container with the Firedrake image, install Thetis and relevant dependencies, code coverage fails. Most obvious when running the adjoint tests as CI suggests inversion_tools.py have no coverage. 

I have tried various ways to fix this as per the linked Issue but to no avail. Will therefore leave it out altogether.

PR #407, opened 2025-07-17
Add testing for demos, fixes #389

### LINKED ISSUE #389: Tutorials aren't tested in CI
It looks like other than the notebooks, none of the tutorials (demos) are currently tested and thus likely to break. Currently at least the `demo_2d_tracer.py` is broken (missing `x, y = SpatialCoordinate(mesh)` line. I know some of them are (more or less) duplicated in the examples/ but that's not a reliable way to ensure they are still runnable. As I don't wont to add significantly to the CI time - and adding `if os.getenv('THETIS_REGRESSION_TEST'):` logic would be confusing/ugly - maybe we can preprocess the scripts before running in CI, or reversely strip out the logic when building the documentation.

stephankramer:
Tests are run in a similar way as for the examples. We only run them to ensure they don't crash - in particular we want to spot errors due to API changes - no checks for correctness.

This required some minor changes in north sea demo:
* move all input files to demos/runfiles
* add if os.environ['THETIS_REGRESSION_TEST'] logic to limit n/o timesteps, which should be hidden on web page
* add two dummy TPXO9 files so we can actually test the TPXO reader

The latter flagged an issue with uptide and the latest version of numpy in which .tostring() (which is now .tobytes()) has been removed. Fixed uptide version now on pypi.


demos/demo_2d_north_sea.py
```diff
@@ -110,6 +108,17 @@ def read_station_data():
 dt = 3600.0
 t_export = 3600.0
 
+# .. raw:: html
+#
+#    <!-- in the CI we only run a few timesteps; commented out in html for the web version
+
+if os.environ['THETIS_REGRESSION_TEST']:
```
cpjordan, line 115:
```suggestion
if os.environ.get('THETIS_REGRESSION_TEST'):
```

Otherwise this will throw an error locally as the flag isn't set outside of testing. In the [3D channel demo](https://github.com/thetisproject/thetis/blob/master/demos/demo_3d_channel.py#L19) we should also add the flag to alter the end time as per the [example](https://github.com/thetisproject/thetis/blob/master/examples/channel3d/channel3d.py#L26) (100s test without it). Will also need to add a comment/comment out test in html version.

I'm running other stuff locally but indicatively: single tracer and 2d channel 8-10s, 2d channel with boundaries 30s, multiple tracers demo takes 20s. They're probably OK without flags.

demos/demo_2d_north_sea.py
```diff
@@ -110,6 +108,17 @@ def read_station_data():
...
+    end_date = datetime.datetime(2022, 1, 15, 9, tzinfo=sim_tz)
```
cpjordan, line 116: Why 9 hours out of interest? If it's just to test that everything is working we could just use a couple of timesteps instead to go from ~90s -> 30s in test time. We have enough other long test times that it doesn't really make too much difference though.

demos/demo_2d_north_sea.py
cpjordan, line 1: [Current L176](https://github.com/thetisproject/thetis/blob/master/demos/demo_2d_north_sea.py#L167) also needs changes to run locally without having to set environment variables and run in terminal (i.e. a student using PyCharm) e.g.:

```
default_data = os.path.join(os.path.dirname(__file__), "runfiles")
data_root = os.environ.get("DATA", default_data)
data_dir = os.path.join(data_root, "tpxo")
```

demos/runfiles/tpxo/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains *DUMMY* TPXO 9 files to test the TPXO reader during CI
+The actual TPXO 9 tidal forcing files can be downloaded from <https://www.tpxo.net/>`__
```
cpjordan, line 2:
```suggestion
The actual TPXO 9 tidal forcing files can be downloaded from <https://www.tpxo.net/>
```

I don't think this is rendering the way you intended

cpjordan (commented):
Added some recommended changes so that the demos can be run in Python locally

demos/demo_2d_north_sea.py
stephankramer: Yeah, fair enough. No particular reason tbh, two timesteps should be enough, changed

demos/demo_2d_north_sea.py
stephankramer: Do you mean it should default to "runfiles/"? The issue with that is that it will then use the dummy TPXO files which really give a very random forcing at the boundary, which I think might be a little confusing for users. This is how the demo is currently setup: the users have to download their own tpxo9 atlas for the demo to work. Alternatively, we could commit some interpolated boundary elevation timeseries but that also gets a bit messy as you have to then explain where these come from, how they would go about creating their own etc. The downloading and registering for tpxo is a bit of a pain, but once you've done it, it is the most flexible approach. Maybe we should discuss in a separate issue?

cpjordan (approved):
Thanks!

PR #410, opened 2025-08-12
Add references to website

cpjordan:
Website has not been updated since #399. When it is updated this will also include Colab links for demo Notebook examples as well as updates from: #387, #402, #401, #404, #407, #403.


cpjordan:
> Two more suggestions

Perfect, thanks for spotting those. I've also fixed the README reference. 

I'll have a few more to add myself by the end of the year so if there are any others missing I'll include them in another update.

Will do a PR on the website page once merged.

stephankramer (approved):
Thanks for that!

> I'll have a few more to add myself by the end of the year

Great, looking forward!

PR #428, opened 2025-10-29
Update ufl methods to properties

cpjordan:
Updates some UFL changes where methods become properties. See https://github.com/FEniCS/ufl/pull/385.


thangel (approved):
Looks good!

PR #431, opened 2025-11-17
Fix weekly release test badge (main)

cpjordan:
At the moment the weekly release test correctly tests on release, but the GitHub actions badge says main. I think this will fix it. I'll just revert if not but I can't test this locally obviously! I think I need to update both branches for this to work but would become consistent anyway.


stephankramer (approved):
See comments at #432

Issue #435, opened 2025-12-04
Direct solver for the tidalfarm headland example


jonathclee:
Hi,

I have been trying one of the example, `tidalfarm/tidalfarm.py`. The optimisation of `turbine_density` uses a direct solver for the shallow water equation (swe):

```
options.swe_timestepper_type = 'CrankNicolson'
options.swe_timestepper_options.implicitness_theta = 0.6
# using direct solver as PressurePicard doesn't work with dolfin-adjoint (due to .split() not being annotated correctly)
options.swe_timestepper_options.solver_parameters = {'snes_monitor': None,
                                                     'snes_rtol': 1e-9,
                                                     'ksp_type': 'preonly',
                                                     'pc_type': 'lu',
                                                     'pc_factor_mat_solver_type': 'mumps',
                                                     'mat_type': 'aij'
                                                     }
```
 
In the `headland_inversion` example, the same direct solver setup is used. When I change to iterative method, say the DIRK22 default solver, the adjoint optimisation likely breaks and returns errors. 

I am not quite sure what is happening, as it gets quite deep into the code. I am wondering if direct solver must be used for several of the time-dependent optimisation/inversion with Thetis. It said `.split()` is not annotated correctly, so iterative methods using splitting like fieldsplit is going to fail (?) Is it possible to use iterative methods for these optimisation/inversion, say maybe for a larger problem that a direct solver is not recommended?

Perhaps a small issue, but just want to know more details. Thank you in advance.

jonathclee:
HI @cpjordan, thank you for your reply. Sorry for taking so long to reply. I was trying to update (or reinstall) newer version of Firedrake, and testing various things before replying.

* Currently, the Firedrake version = 2025.10.2 . 

* Not sure about Thetis version. I follow the installation instruction on webpage : `pip install git+https://github.com/thetisproject/thetis.git@release`, which installed a stable release version but not the latest version on github (?)

* I checked the `channel_inversion/inverse_problem.py` example (`make invert CONTROLS=Manning`) and it finished without error.  I don't think it is due to control defined on real space. There seems to be no solver_parameters issues (?)

* I cannot test headland inversion example as I am not on latest version of Thetis (Some changes / objects are missing).

==================================================================

[...]

PR #452, opened 2026-05-01
Update DeferredExportManager assertion [main]

cpjordan:
See #442 & #435. Already committed in release.


stephankramer (approved):
Looks good

PR #453, opened 2026-05-01
Update release branch [2026.4.1]

cpjordan:
Update release to mirror Firedrake, merging all the main (development) branch commits since [2025.10.1](https://github.com/thetisproject/thetis/releases/tag/v2025.10.1) into the stable release branch.


cpjordan:
Note that I can't do a hard reset of `release` because of branch protection rule we want to keep.

stephankramer (approved):
Yes, I think that's fine (merging main into release instead of moving the release branch pointer to main and merging old release into it) - IIRC from git's perspective merges are symmetric. It (the ordering of the parents) may affect how things are displayed on github I think but I'm not sure which way around is better..

stephankramer:
Oh hang on the spatialindex/rtree change is not in release. So this one https://github.com/thetisproject/thetis/pull/453/commits/808b4d5549f92fcb7f3580f225243ceb83b72995 shouldn't go into Thetis release? I'm confused why it's passing the tests...

stephankramer:
Ah they are failing now. Looks like (push) is testing against firedrake main and (pull_request) is testing against firedrake release.

cpjordan:
> Oh hang on the spatialindex/rtree change is not in release. So this one [808b4d5](https://github.com/thetisproject/thetis/commit/808b4d5549f92fcb7f3580f225243ceb83b72995) shouldn't go into Thetis release? I'm confused why it's passing the tests...

Looks to be the case - I need to fix the Thetis branch tags. This will teach me for going on holiday.

cpjordan:
> Yes, I think that's fine (merging main into release instead of moving the release branch pointer to main and merging old release into it) - IIRC from git's perspective merges are symmetric. It (the ordering of the parents) may affect how things are displayed on github I think but I'm not sure which way around is better..

I couldn't decide either. If things are always cherry-picked from `main` into `release` between major updates moving forward (my fault), then I think this won't matter so much as git should recognise that the commits are identical on both branches.

> Ah they are failing now. Looks like (push) is testing against firedrake main and (pull_request) is testing against firedrake release.

This is weird - it seems like `release` has been merged into `main` as part of this PR and the push into `main` (from `release`) is showing up here (?!). I'm not sure how this has happened but that is concerning.

EDIT: this is because from GitHub's perspective on this PR, `main` is the working branch and `release` is the target and to do a PR the working branch needs to be update with the target branch.

Seems like I've somehow made a mess of this - I'll try and fix it all now.

cpjordan:
See #454.

PR #454, opened 2026-05-01
Update release branch [2026.4.0]

cpjordan:
Update release to mirror Firedrake, merging all the `main` (development) branch commits since [Thetis 2025.10.1](https://github.com/thetisproject/thetis/releases/tag/v2025.10.1) into the stable release branch. A `sync/release-update` branch is being used rather than `main` directly because:

- There are some recent updates we have made that are not compatible with [Firedrake 2026.4.0](https://github.com/firedrakeproject/firedrake/releases/tag/2026.4.0)
- It becomes very easy to inadvertently update `main` if `release` has updates that were not cherry picked from `main`


cpjordan:
@stephankramer I've removed the incorrect 2026.4.1 tag and erroneous merge commit to `main` from `release`. I'd suggest now doing a merge commit and bypassing rules (rather than updating the `sync/release-update` branch first, which will create a second merge commit). `release` will end up two commits behind `main` and we should be clean again as long as any future updates to `release` are done by cherry picking from `main` first.

stephankramer (approved):
I think this looks good. So this is merging everything from main minus the last two commits right?
* the rtree fix which we don't want in release
* and #452
Just for my curiosity: are you still planning to get that last one in release? If so how? Tbh I'm not sure what the cleanest way to do that. Cherry-picking may cause merge conflicts later on but I don't think it's a major issue really, tiny commit anyway.

cpjordan:
> I think this looks good. So this is merging everything from main minus the last two commits right?

Yes

> Just for my curiosity: are you still planning to get that last one in release? If so how? Tbh I'm not sure what the cleanest way to do that. Cherry-picking may cause merge conflicts later on but I don't think it's a major issue really, tiny commit anyway.

No, I will leave #452 for next time. #452 (main) was just #442 (release), so the fix is now on both branches. In the next `release` update, we can simply force update `release` and re-write the `release` history cleanly, which will remove this duplication. From now on I'll make sure anything in `release` is cherry picked from `main` so I don't get us into this mess again!

cpjordan:
I would just move `release` over next time. I can't think of any major consequences to that. We can discuss next time or at the next meeting. Thanks for your help & sorry for the mess - lesson learned!

jhill1:
Just to note, is the process of producing a release documented?

cpjordan:
> Just to note, is the process of producing a release documented?

Not currently, I'm not sure we know what the best approach exactly is yet - last time I caught up with Stephan I mentioned that we should have a set of notes for website updates, Zenodo, producing a release etc. to help with continuity. I can do a PR for this soon - nearly out of a busy period. It would be nice to have Release notes (per major Firedrake version) moving forward as it seems there is a reasonable number of people using `release` rather than `main`

PR #455, opened 2026-05-06
Add scheduler to tidalfarm.py

### LINKED ISSUE #451: Adding adjoint scheduler to tidalfarm example
Hi,

I was trying to add the update #430 (also used in [headland_inversion](https://github.com/thetisproject/thetis/tree/808b4d5549f92fcb7f3580f225243ceb83b72995/examples/headland_inversion) example) to the [tidalfarm.py](https://github.com/thetisproject/thetis/blob/808b4d5549f92fcb7f3580f225243ceb83b72995/examples/tidalfarm/tidalfarm.py) example. Following the steps by adding `tape.enable_checkpointing` and `adj_timestepper` appropriately:
```Python
...

tape.enable_checkpointing(
    SingleMemoryStorageSchedule(),
    gc_timestep_frequency=28,
    gc_generation=2,
)

# ...(code from tidalfarm.py)...

# adj_timestepper
num_steps = int(np.ceil(solver_obj.options.simulation_end_time / solver_obj.dt))
adj_timestepper = tape.timestepper(iter(range(num_steps)))
solver_obj.iterate(update_forcings=update_forcings, adj_timestepper=adj_timestepper)

# ...(rest of the code)...
```
But running the script gave error and failed taylor test (with `THETIS_REGRESSION_TEST=1`). Here is part of the log:
```
...

Functional value: None
Costs, average power and profit for each farm: [0.0], [0.0], [0.0]

...

Order of convergence with taylor test (should be 2) = 0.9985956654071169
Running Taylor test
Computed residuals: [0.0025555889249275765, 0.0012790388866612545, 0.0006398324532082, 0.0003199947183842866]
Computed convergence rates: [0.9985956654071169, 0.9992940522332351, 0.9996460766815117]
Traceback (most recent call last):
  File "/home/ubuntu/work/testing/tidalfarm_adjtap.py", line 229, in <module>
    assert minconv > 1.95
           ^^^^^^^^^^^^^^
AssertionError
```
But removing only the lines with `tape.enable_checkpointing(...)`, it finishes without error (and passing taylor test).

Currently on Firedrake `2026.4.0` and on the `main` branch of Thetis.

It works in other inversion examples like headland inversion, but seems not the proper way to add adjoint scheduler for the tidalfarm example (?) I would guess the major difference is `InversionManager` and `TurbineFunctionalCallback`, different handling of cost functional (and therefore tape).

Is there a suggestive/correct way to add adjoint scheduler to the tidalfarm example? Do we need to construct a class similar to `InversionManager` to make this work? 

Would appreciate if there is any advice. Thank you!

cpjordan:
This PR adds checkpointing for adjoint optimisation in [examples/tidalfarm/tidalfarm.py](https://github.com/thetisproject/thetis/blob/main/examples/tidalfarm/tidalfarm.py).

The Taylor test is made to pass by syncing the live control `Function` before each `ReducedFunctional` replay, but the optimisation diagnostics need to be updated to fix the inversion.

All optimisation callbacks were being run as `derivative_cb_post`. With checkpointing, this is too late for diagnostics that depend on replayed forward values. By the time the adjoint derivative callback runs, checkpointing may already have discarded some saved forward values, so the callback can see `None` or stale zero values.
The fix separates the callbacks by when their data is valid:

```
rf = ReducedFunctional(
    scaled_functional,
    c,
    eval_cb_pre=update_control,
    eval_cb_post=callback_list,
    derivative_cb_post=derivative_callback_list,
)
```

where `update_control` syncs the live control before each replay:

```
def update_control(value):
    turbine_density.assign(value)
    projector.project()
```

`eval_cb_post` now handles functional/turbine/control diagnostics immediately after the forward replay. `derivative_cb_post` now only exports the adjoint derivative.

**For future reference**
Without checkpointing, pyadjoint keeps enough forward state in memory that the adjoint can run directly backwards through the recorded tape.
With checkpointing, pyadjoint deliberately does not keep all forward states. Instead, it stores selected restart states, discards other intermediate values, and later recomputes parts of the forward model when the adjoint needs them. This saves memory, but it means the timing of callbacks matters: some replayed forward values are only reliably available immediately after the replay, not after the later adjoint cleanup.
There is also a control-state subtlety. `ReducedFunctional` updates pyadjoint’s checkpointed value of the control, but the live Firedrake `Function` can still contain the old value. In checkpointed replay, some operations can fall back to that live `Function`. So we explicitly assign the trial control into the live `turbine_density` before each replay.

### DIFF
--- a/examples/tidalfarm/tidalfarm.py
+++ b/examples/tidalfarm/tidalfarm.py
@@ -22,27 +22,17 @@ the forward model and automatically derives the adjoint.
 from thetis import *
 from firedrake.adjoint import *
 from checkpoint_schedules import SingleMemoryStorageSchedule
-import logging
 op2.init(log_level=INFO)
 
 # start annotating all Firedrake opterations in the forward model
 continue_annotation()
 
 tape = get_working_tape()
-output_logger.handlers = []
-output_logger.propagate = False
 tape.enable_checkpointing(
     SingleMemoryStorageSchedule(),
     gc_timestep_frequency=56,  # every 56 time steps we will do garbage collection
     gc_generation=2
 )
-# TODO: this is a logging workaround - update when Firedrake #4753 is addressed
-if COMM_WORLD.rank == 0:
-    handler = logging.StreamHandler()
-    handler.setFormatter(logging.Formatter('%(message)s'))
-    output_logger.addHandler(handler)
-else:
-    output_logger.addHandler(logging.NullHandler())
 
 # setup the Thetis solver obj as usual:
 mesh2d = Mesh('headland.msh')
--- a/examples/tidalfarm/tidalfarm.py
+++ b/examples/tidalfarm/tidalfarm.py
@@ -21,10 +21,28 @@ the forward model and automatically derives the adjoint.
 
 from thetis import *
 from firedrake.adjoint import *
+from checkpoint_schedules import SingleMemoryStorageSchedule
+import logging
 op2.init(log_level=INFO)
 
 # start annotating all Firedrake opterations in the forward model
 continue_annotation()
+
+tape = get_working_tape()
+output_logger.handlers = []
+output_logger.propagate = False
+tape.enable_checkpointing(
+    SingleMemoryStorageSchedule(),
+    gc_timestep_frequency=56,  # every 56 time steps we will do garbage collection
+    gc_generation=2
+)
+# TODO: this is a logging workaround - update when Firedrake #4753 is addressed
+if COMM_WORLD.rank == 0:
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter('%(message)s'))
+    output_logger.addHandler(handler)
+else:
+    output_logger.addHandler(logging.NullHandler())
 
 # setup the Thetis solver obj as usual:
 mesh2d = Mesh('headland.msh')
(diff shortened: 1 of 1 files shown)


cpjordan:
Closes #451.

examples/tidalfarm/tidalfarm.py
```diff
@@ -21,11 +21,29 @@
...
+    gc_timestep_frequency=56,  # every 56 time steps we will do garbage collection
+    gc_generation=2
+)
+# TODO: this is a logging workaround - update when Firedrake #4753 is addressed
+if COMM_WORLD.rank == 0:
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter('%(message)s'))
+    output_logger.addHandler(handler)
+else:
+    output_logger.addHandler(logging.NullHandler())
```
cpjordan: Can be removed when #457 is merged.

examples/tidalfarm/tidalfarm.py
```diff
@@ -164,19 +184,33 @@ def update_forcings(t):
...
     turbines.TurbineOptimisationCallback(solver_obj, cb),
 ])
+derivative_callback_list = optimisation.OptimisationCallbackList([
+    optimisation.DerivativesExportOptimisationCallback(solver_obj),
+])
+
+
+def update_control(value):
+    turbine_density.assign(value)
+    projector.project()
```
cpjordan, line 188: Should be able to avoid this when https://github.com/dolfin-adjoint/pyadjoint/pull/257 is merged

PR #457, opened 2026-05-13
Remove logging duplication with adjoint scheduler

cpjordan:
Fixes duplicated Thetis log output under MPI by disabling propagation on loggers when installing handlers. See https://github.com/firedrakeproject/firedrake/pull/5086.

Change: set `logger.propagate = False` inside `set_thetis_loggers()` in `thetis/log.py`.

Result: `print_output(...)` and `info(...)` no longer double-print (or print in mixed formats) when running with `mpirun -np N`.


stephankramer (approved):
Perfect, thanks Connor

PR #458, opened 2026-05-13
Add developer notes to website

cpjordan:
Tidies up repo a little and updates website documentation (including `main` vs `release` strategy).

Key changes:
- Added a developer notes page (docs/source/developer_notes.rst) to explain various aspects of maintaining Thetis.
- Clarified installation modes (editable/dev, CI-like --system-site-packages, docs builds), CI behavior, website publishing workflow (built and published from release into thetisproject.github.io), Zenodo failure recovery, and a concrete, no-cherry-picks main/release maintenance policy with worked examples and GitHub merge guidance (including branch-protection “Update branch” pitfalls).
- Simplified docs/source/download.rst to be explicitly release-branch instructions with a pointer to developer notes for main.
- Added optional dependency groups to pyproject.toml for linting and testing.
- Improved ignores for common local artifacts (.idea/, .pytest_cache/, .vscode/, .venv/, .mypy_cache/, .ruff_cache/, .codex/, .agents/).
- Standardized README naming/links (readme.md -> README.md, fixed references to examples/README.md, updated pyproject.toml readme path).
- Updated docs homepage sidebar to show both release and main weekly CI badges.


docs/source/developer_notes.rst
```diff
@@ -0,0 +1,339 @@
...
+================
+
+This page is aimed at people developing Thetis, maintaining CI, and doing
+releases.
+
+
+Installing Firedrake
+---------------------
+
+Installation of Firedrake for the ``main`` branch follows a different set
```
stephankramer: If you're targetting people who haven't used firedrake main before, you need to first explain why this is relevant for Thetis development. I would start with a small introduction section saying that development of new features should happen on Thetis main which is only compatible with Firedrake main. Maybe move your overview of branches below up?

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,339 @@
...
+
+Alternatively, the GitHub Actions workflow installs Thetis into a fresh venv *that can
+still see the Firedrake site-packages* by using ``--system-site-packages``::
+
+  python3 -m venv --system-site-packages venv-thetis
+  . venv-thetis/bin/activate
+  pip install .
+
+This pattern is useful when you want to reproduce CI locally without polluting
+your base Firedrake environment.
```
stephankramer: I think this is a bit confusing. I think this only works in the CI container because there firedrake is installed in the *system* /usr/local/python3.X/site-packages locations (it's using an editable install so the source is still elsewhere, but the way it's discovered is through the system python paths). If you install firedrake in an(other) venv itself as per the standard instructions, this won't work. Installing firedrake in system locations on an actual desktop/laptop is absolutely not recommended. So I would just drop this. Or explain this is only when you are developing in a firedrake-vanilla container.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,339 @@
...
+Release vs main, and updating release
+--------------------------------------
+
+Branch intent
+~~~~~~~~~~~~~
+
+* ``main`` is the development branch and is tested against the Firedrake
+  ``dev-main`` Docker image in CI.
+* ``release`` is the stable branch and is tested against the Firedrake
+  ``latest`` Docker image in CI.
```
stephankramer: Yeah I would have this section right on top. As it saves you quite a few forward references above. Also say explicitly "main" is the development branch that track the firedrake main branch, and release is the branch that is compatible with the latest firedrake release - rather than only saying which Firedrake docker image it uses in CI (cause knowing which branch that corresponds to is a bit more obscure knowledge)

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,339 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Thetis updates to the stable branch are done with PRs and must preserve history
+(use merge commits, not squash/rebase).
+
+To update ``release``:
+
+1. Open a PR with base branch ``release``.
+2. Wait for CI to pass and the PR to be reviewed.
+3. Merge using "Create a merge commit" (do not squash merge or rebase merge).
```
stephankramer: Could you clarify why this rule? I mean I do prefer a history that maintains nicely commented commits if they are not WIP with a merge commit that summarizes them. I would guess most PRs on release (hot-fixes) would be single commits though. But the main thing is that the branch is that's merged in is based of release and not main. If people have based it off main, none of the merge strategies prevent commits on main getting into release if that's what you were aiming at.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,339 @@
...
+2. When a user-facing bugfix is needed on the stable branch.
+
+Policy: keep ``release`` changes minimal and compatibility-driven. Prefer
+targeted fixes over large feature merges.
+
+Updating ``release`` On GitHub
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Thetis updates to the stable branch are done with PRs and must preserve history
+(use merge commits, not squash/rebase).
```
stephankramer: Oh, I see this might answer my comment below: note that there's no reason not to squash or rebase on *branches* other than release or main. What people do on branches is first of all their own business - and if people have developed something and there's lot of work-in-progress commits, that don't really work on their own, commits that try something out and are then reverted again - I would want people to rewrite the history *on that branch* and get rid of unnecessary commits. All that matters is what the branch looks like when its presented in the PR. And even then, if after a discussion on the PR, people will need to undo some stuff and replace it with something and then it's actually *preferable* if people clean up their branch and force-push it to the branch that's up for a PR. Removing changes in the history of the branches that are then undone or in a different way, helps preventing merge conflicts in the future.

Then as long as you do everything via PRs, it doesn't matter how you get it into release or main, if you choose "rebase and squash" it just means you get a single commit on top of release or main, there's no history rewritten on main or release.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,339 @@
...
+
+1. Open a PR with base branch ``release``.
+2. Wait for CI to pass and the PR to be reviewed.
+3. Merge using "Create a merge commit" (do not squash merge or rebase merge).
+4. Sync ``release`` into ``main`` when needed:
+
+   * Required: if the PR introduced commits that do not already exist on
+     ``main`` (for example a hotfix implemented on ``release``).
+   * Optional: if the PR only advanced ``release`` to a vetted commit that was
+     already on ``main`` (for example during a Firedrake stable release update).
```
stephankramer: To make this a little less confusing, I would create a separate section for your workflow to create a new release. Most people that want to just contribute a fix to release, do not need to know all the ad-hoc nitty-gritty of that.

docs/source/developer_notes.rst
stephankramer: Just to add: this of course doesn't apply if you're merging release into main (or vice versa) via a PR as you point out-below: then of course you have to be careful not to rewrite history on the to-be-merged-in branch

stephankramer (changes requested):
Some initial comments to get the conversation going. I mostly agree with everything I've read so far - just wanted to clarify a bit on history-rewriting restrictions.

cpjordan:
> Some initial comments to get the conversation going. I mostly agree with everything I've read so far - just wanted to clarify a bit on history-rewriting restrictions.

Thanks - I imagine this will take a bit of iteration, particularly on the `main` vs `release` stuff. I think I'll do a demo repo to check what I am saying & strategy makes sense (taking into account your comments). I should have time to work on this again next week - I'm particularly keen on prioritising this so it's clear & easy moving forward.

docs/source/developer_notes.rst
cpjordan: Agree - removed but kept a note of it in the CI section

docs/source/developer_notes.rst
cpjordan: This is hopefully clarified now. Indeed, we would generally prefer squash commits, I just initially wanted to make it really clear that we don't want to lose individual commits when updating release with main.

cpjordan:
Sent some slides over to @stephankramer after doing a demo repo and testing how I can minimise the number of wasteful commits whilst maintaining release-main ancestry and prohibiting force pushes to main or release. It's the best I can currently think of. 

My opinion on the "easiest" alternative to maintain is to hard reset release at each Firedrake major update, with a protected branch before doing so that the release history is kept somewhere. No git pull on release at that point, but a user is basically going to have to re-install PETSc and Firedrake anyway if they're updating. Keeps main as one continuous chain.

Happy whichever way as long as it is documented!

(I assume the runners are currently down due to the heatwave, not that this PR needs any testing)

cpjordan:
I would prefer to deal with and merge #438 before this, so I can update the CI notes. Will try to do that this afternoon if I don't get distracted.

cpjordan:
Closing this PR because the release/versioning approach has changed substantially. I’ll open a new PR with the updated approach and revised release workflow docs.

PR #461, opened 2026-06-13
Add developer notes to website and move to dynamic versioning

cpjordan:
This PR updates the Thetis release workflow and switches package versions to dynamic git-based versioning.

**What Changed**
- Added developer notes for managing main and release.
- Documented the hard-reset workflow for major Firedrake releases:
  - tag the old release state;
  - hard-reset release to a vetted main commit;
  - tag the new release state.
- Clarified that most fixes go to main first.
- Clarified when release should be updated and tagged.
- Updated install docs - they are release focussed, otherwise re-directed to developer notes.
- Added a way to download only examples/demos without cloning all of Thetis.

**Versioning**
Propose that Thetis now uses dynamic versioning with setuptools-scm.
- main: YYYY.M.devN
  - Example: 2026.6.dev4
- release: YYYY.M.postN
  - Example: 2026.4.post0, 2026.4.post1

**Why Packaging Changed**
Previously, the version was hard-coded in pyproject.toml, so it had to be edited manually.
This PR makes git tags the source of truth instead. It removes old Versioneer files and uses setuptools-scm so package versions follow the current branch/tag state.
This also avoids confusion with Firedrake patch versions: Thetis 2026.4.post1 does not imply it matches Firedrake 2026.4.1. Firedrake might make lots of release patches and get to e.g. 2026.4.5, with no Thetis changes at all - it would confuse me that there was only a Thetis 2026.4.0 - hence adding post to differentiate.


docs/source/download.rst
```diff
@@ -74,26 +77,45 @@ If you have SSH keys set up with GitHub, you can use the SSH-based clone instead
    git checkout <branch_name>
    pip install -e .
 
-.. note::
+In an editable install, ``git pull`` updates the Thetis source code, but it does
```
stephankramer:
```suggestion
In an editable install, ``git pull`` updates the Thetis source code and these changes take immediate effect in your (run-time) python environment, but it does
```

stephankramer: Just to clarify that of course "git pull" always updates the source code, it's just that in an editable install these changes are also the ones that get imported.

docs/source/download.rst
```diff
@@ -74,26 +77,45 @@ If you have SSH keys set up with GitHub, you can use the SSH-based clone instead
...
+of the repository, you can export just those directories with Subversion:
 
-An IDE such as PyCharm will not recognize Thetis when installed in this fashion for any project outside the cloned repository,
-as the source is not in site-packages. It will still run, and if you would like to enable full code navigation, you
-can add the Thetis cloned repository as a content root, then add the ``thetis`` sub-directory as a sources root.
+.. code-block:: none
+
+   mkdir thetis-examples
+   svn export https://github.com/thetisproject/thetis/branches/release/examples thetis-examples/examples
+   svn export https://github.com/thetisproject/thetis/branches/release/demos thetis-examples/demos
```
stephankramer: This does not actually work! I presume you asked copilot - which gave me the same suggestion. So did claude. When I complained, it said "You're right, GitHub deprecated SVN support in 2023. Here are the working alternatives".
For people that want to access the demo and examples but want to avoid git, I would just point to https://api.github.com/repos/thetisproject/thetis/tarball/release which gives you a .tar.gz of release. If you want command line instructions use `curl -JLO` to get a properly named .tar.gz file, or use `curl -L https://api.github.com/repos/thetisproject/thetis/tarball/release | tar xz` to download and extract in one line.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+* Target branch: PRs should usually target ``main``. PRs to ``release`` should
+  only be used for release-management work, release-only fixes, or exceptional
+  backports that must ship before the next Firedrake release cycle.
+* CI: PRs are tested by ``.github/workflows/pr.yml`` against the appropriate
+  Firedrake Docker image (based on the PR base branch).
+* Before opening/merging a PR, it is expected that lint is clean
+  (``make lint``),
+  tests are run where practical (see the CI commands above), and docs changes
+  render as expected (build the Sphinx site locally if you edited
+  ``docs/source/*``).
```
stephankramer:
```suggestion
  render as expected (build the Sphinx site locally, using ``make html`` in the ``docs/`` directory,  if you edited
  ``docs/source/*``).
```

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+* ``main`` is the normal development branch. New features and ordinary bug fixes
+  should land on ``main``.
+* ``release`` is the current stable branch. It tracks the latest stable
+  Firedrake release and is the branch used for the published website.
+* When Firedrake makes a major release, Thetis ``release`` is hard-reset
+  to a vetted ``main`` commit that is compatible with that Firedrake release.
+* Between these reset points, avoid routine backports. Only patch ``release`` if
+  the current stable line has a release-only issue, or if an urgent shared bug
+  must be shipped before the next reset.
```
stephankramer: I feel this could be a bit clearer. I personally wouldn't want to discourage "less urgent" bug-fixes to be cherry-picked to release. What I would explicitly want to proscribe: new functionality, rewrites of a substantial part of a routine - anything with the potential to introduce new bugs - and in particular any API changes: a script that runs against some release version, should still run against any further patch levels.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+      git checkout -b release-backport/<short-description> origin/release
+      git cherry-pick -x <main-fix-sha>
+      git push -u origin release-backport/<short-description>
+
+2. Open a PR with base branch ``release``.
+3. Merge with a method that leaves a single release-side fix commit where
+   possible (squash merge, or rebase merge if enabled). If using squash merge,
+   keep the original ``cherry picked from`` line or mention the source ``main``
+   PR/commit in the squash commit message.
+4. If this produces a new published release state, tag the resulting
```
stephankramer: When would it not "produce a new published release state" ?

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+   git reset --hard <vetted-main-sha-or-origin/main>
+
+Tag and push the new release state immediately after the hard reset. The package
+version is derived from git tags, so no version-only commit is needed at the
+hard reset point.
+
+.. code-block:: none
+
+   git tag -a <new-release-version> HEAD -m "Thetis <new-release-version>"
+   git push --force-with-lease origin release
```
stephankramer, line 344: TIL about `--force-with-lease`. I like it!

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+   .. code-block:: none
+
+      git fetch origin
+      git checkout -b release-fix/<short-description> origin/release
+
+2. Commit the fix and open a PR with base branch ``release``.
+3. Merge with a method that leaves a single release-side fix commit where
+   possible.
+4. If this produces a new published release state, tag the resulting
```
stephankramer: Same question

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+   main:    A -- B -- C -- D
+   release: A -- B -- C -- D
+   tag:     A -- B -- R1
+
+
+Cutting A Tagged Release
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+When cutting a tagged release, push a tag, create a GitHub Release, and verify
```
stephankramer: Maybe it's because I'm not a native speaker (but so are other devs) but I find the phrase "cutting a release" a bit confusing, someone who's not familiar with it could think you're talking about *deleting* a tagged release.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+Tags are normally created only:
+
+* immediately before a hard reset, to preserve the final installable state of
+  the old stable line;
+* immediately after a hard reset, to mark the new Thetis release for the new
+  major Firedrake release; or
+* when the current ``release`` branch must be updated because a Firedrake patch
+  or other breaking change in the stable Firedrake stack requires a new Thetis
+  release state.
```
stephankramer: Hm, the point of patch releases in firedrake is that this should *never* happen. So I think this is a bit confusing. Above it seems you're suggesting that with *any* new commits on release (either shared fix or release-only fix workflows) people should create a new tag - but then again I'm probably misunderstanding what you meant by "new state" above...

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+Use separate development and release forms:
+
+* ``main`` uses ``YYYY.M.devN``, where ``YYYY.M`` is the month of the current
+  ``HEAD`` commit and ``N`` is the number of commits in that month.
```
stephankramer: I think I'm missing some context here: thus far you were talking about tagging releases, but now you're talking about labels for the main branch? When do they come up? Are these instructions for how people should tag inbetween main commits if they wish to do so? It should say somewhere...

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+* ``release`` uses ``YYYY.M.postN``, where ``YYYY.M`` is the Firedrake major
+  release series used by that Thetis release branch and ``N`` is incremented for
+  each Thetis release update on that branch.
+
+The ``postN`` spelling is the PEP 440-compatible package-version equivalent of a
+``relN`` label. Do not use tags such as ``2026.4.rel1`` for Python package
+versions: they are not valid PEP 440 versions.
```
stephankramer: This second sentence seems out-of-place: we're not here to instruct people on how they should tag python packages in general.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+For example, development commits made in June 2026 should produce versions like
+``2026.6.devN``. Using the commit month rather than the build month keeps a
+given commit's version stable. The first Thetis release against the Firedrake
```
stephankramer: Hm, this opens up a can of worms. I'm not sure what you mean by "build month", but we need to distinguish between "committer date" and "author date". The first is when the commit was (re)written (changes on a rebase). The second is when the author signed off on the commit (maintained when rebased and also (by default) on git commit --amend). So the phrase "development commits made in June 2026" is a bit ambiguous. The issue with author date is that it can be a year(s) before it was actually merged. So I agree that it should be "commit date" as that's most likely when it was merged (and tested!) in main. That also means that people shouldn't randomly tag commits if they are part of a PR that was merged through a merge commit.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+numbering.
+
+Thetis uses dynamic versioning from git tags and branch state. A tagged
+``release`` commit has the exact release version from the tag. Untagged
+development commits on ``main`` and ordinary feature branches get the
+current-month development version. Untagged commits on ``release`` and
+``release-*``/``release/*`` maintenance branches get the next ``postN``
+development version for that release line. If ``main`` and ``release`` point at
+the same tagged commit, they report the same exact version; once ``main``
+receives new commits, it reports a development version again.
```
stephankramer: Oh I see where you're going with this. This whole section is a bit confusing tbh: it's mixing up tags and python package versions all over the place, and it's never clear which one you're talking about. Further above you are instructing people to tag releases whenever it's patched with a "see below for tag conventions" and then they arrive at this section where you're talking about 2026.6.devN.

I would suggest the following: have a separate section *first* to explain *tag* conventions - which is something the developer needs to specify (following these rules). Then another section on how the python package version is *automatically* derived, but don't go into too many details about the motivation because it's not something that the developer needs to worry about.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+---------------
+
+Content
+~~~~~~~
+
+Thetis documentation content lives in this repository under ``docs/source`` and
+is built with Sphinx (see ``docs/Makefile`` and ``docs/source/conf.py``).
+
+Thetis has a single published documentation website. The published site is
+updated whenever the ``release`` branch is updated, and it must be built from
```
stephankramer:
```suggestion
updated whenever the documentation content in the ``release`` branch is updated, and it must be built from
```

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+   ``docs/source/download.rst``).
+
+3. Rebuild locally (step 1) to check the rendered output.
+
+4. Commit the changes in the Thetis repo and open a PR.
+
+   If the change is intended for the published website, the PR must target
+   ``release`` (since the published website is built from ``release``). It is
+   fine to iterate on documentation changes in a PR targeting ``main``, but do
+   not publish a website build from ``main``.
```
stephankramer: This now goes against our policy to in-principle merge into main first. If a website update goes into release only, it will disappear on the next reset.

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,470 @@
...
+
+5. When the PR is approved, merge it. At the same time, copy the contents of
+   ``docs/build/html/`` into the ``thetisproject.github.io`` repository and
+   merge there as well.
```
stephankramer: Clarify: although it adds another round of waiting for approval, I think it can be worth doing this via a PR as well - I wouldn't read the whole thing again (some of it is auto-generated) but it's quick to  check people have done the correct thing (also for yourself). It's easy to introduce some unintended changes. So maybe: "It is encouraged to do this via a PR and to check on github that all changes are as intended. If unsure, ask for a review" - I think it's currently unprotected. Would also be happy with a stricter approach if we add standard branch protection.

docs/source/publications.rst
```diff
@@ -16,22 +16,37 @@ Toward a digital twin of the Great Barrier Reef: Impact of extreme model resolut
...
+Guichard J., Rawlinson-Smith R., Coles D. and Greaves D.:
+Energy storage from a wind, tidal, and wave farm via hydrogen.
+*Front. Energy Res.* 14, p. 1732813,
+doi: `10.3389/fenrg.2026.1732813 <https://doi.org/10.3389/fenrg.2026.1732813>`_, 2026.
+
+Bolhuis, S., Alday, M. and Lavidas, G.:
+Tidal stream energy in the Netherlands—Resource assessment and future effects due to mean sea level rise,
+*Renewable Energy*, 256, Part F, p. 124214,
+doi: `10.1016/j.renene.2025.124214 <https://doi.org/10.1016/j.renene.2025.124214>`_, 2026.
+
```
stephankramer, line 33: This is another case, where it'd be happy for a publications update to be cherry-picked into release so that the website is updated immediately...

setup.py
```diff
@@ -2,5 +2,6 @@
 
 from distutils.core import setup
 from glob import glob
+from os.path import isfile
 
-setup(scripts=glob('scripts/*'))
+setup(scripts=[path for path in glob('scripts/*') if isfile(path)])
```
stephankramer: Why would we have directories under scripts/ ? And if so, would they contain scripts that need discovering?

stephankramer (changes requested):
Looking really good. Agree with almost all of it. Just some clarifications.

Main sticky point for me is that we need to think a bit more about how we ensure the release after a reset is actually tested *before* we publish it. The whole point about stable release tags is that we can go back to a known stable and tested version, that is known to work with a certain release version of Firedrake, or in the associated container. For the patch release `.postN`, N>1 this is guaranteed through github's CI but the workflow for the release reset doesn't. It should be that release at that point is the same as main, which should have been recently tested against firedrake main, which should be close to the just released firedrake release. However that's still a bit approximate, and in particular in the run-up to release firedrake main can change quite rapidly. What we want to guarantee is that the release version runs correctly in the latest firedrake release container. You could do this locally (in a docker container) but that's a bit messy and doesn't guarantee you've done exactly the same steps as we've set up in CI. We can't easily do this via a "draft" PR, because if the base is set to main, it will run the wrong container, and if the base is set to "release" it may have conflicts with the extra commits from current release. We could just do the force push to release, github will then trigger the push.yml workflow, and cross our fingers that it passes and then tag .post0 - but what to do if it doesn't pass?

stephankramer:
How about we introduce another special branch name "release-candidate" that whenever you push to it triggers a CI run in the firedrake: latest container (i.e. add it to the rules in push.yml). So you:
1. hard-reset release-candidate to main
2. force push release-candidate
3. check the actions tab to see that it passes
4. if it doesn't implement necessary fixes directly on release-candidate, and push them until it works - you can rewrite the fix commits as much as you like as long as it keeps current main as its base unchanged.
So at this point release-candidate should pass CI, then:
5. If new commits were introduced in 4. fast-forward main to release-candidate, push to main even it breaks CI (by passing branch protection that requires a PR but only using `git push` **without** --force to ensure we're not rewriting history)
6. reset release to release-candidate/main
7. force push release
8. tag release
9. if you have broken main under 5. open up a PR as normal to fix it

To clarify point 5+9: if we're a little late (people are busy!) generating the new release after firedrake's, firedrake main may have moved on already. This may for instance mean that at the point that we try to generate a release, Thetis main is already broken against current firedrake main (even though it passed the last time the CI ran). At that point there's another rule we should introduce - as it went wrong the last time: as soon as new Firedrake release has been released, we should not accept any changes to main before we have finished the corresponding Thetis release. Otherwise we run the risk of only working with Firedrake main and (no longer) against the latest release, and these commits would have to be undone on the release branch. In any case, if fixes were needed under 4. chances are that firedrake main was already broken, so by-passing CI checks on main in step 5 is not a big deal.

A slightly simpler version of this would be to immediately push to release (as it currently is described) and immediately try to fix it directly on release if it doesn't pass CI (and then do 5. to 9.). Downside is that there's a window of time with an unstable release....

jhill1:
I like this process. It means if there are major issues with a release candidate we have time to implement any fixes without pressure.

cpjordan:
Looking unlikely that I have time to fix this before I go away - back 29th June but happy for anyone else to move this forward in the meantime.

docs/source/developer_notes.rst
cpjordan: See revised wording - hopefully clearer now

docs/source/developer_notes.rst
cpjordan: I think I was hesitant on whether we want to tag every single update of release. "Published release state" was referring to a 'significant' update. In Firedrake they have e.g. 2026.4.0, 2026.4.1, ... but each new version (.1, .2, ...) contains multiple commits. We might have wanted to do something similar (e.g. we wouldn't bother tagging release when one of us has updated the publications on the website), but I guess it doesn't align because Firedrake have proper pip installation for release.

I've now updated the instructions so that we always tag release when it is updated. That said, we might not want to do a Zenodo release each time.

For me this is probably the main remaining bit of discussion. Happy to have a meeting if easier!

docs/source/developer_notes.rst
cpjordan: I've changed this to "Publishing a Tagged Release".

cpjordan: This is where we might want to state when we would want to do a Zenodo/GitHub release. This infers you do it every time we update release, which is probably excessive. My thoughts would be:

- Before a Firedrake release update. You then have the final version working with an old Firedrake release version.
- After a Firedrake release update. You then have the initial working version with the new Firedrake version. This also notifies subscribed users by email that there is a "new" version of Thetis (release).
- If we are asked to because someone wants a specific Zenodo release for their paper into an open code journal.

docs/source/developer_notes.rst
cpjordan: Clarified as per your later comment - I had started mixing them up because the tag should also ideally relate to the version shown if you pip list.

docs/source/developer_notes.rst
cpjordan: I think I have done this now.

setup.py
cpjordan: I think I was future-proofing - I've undone it

pyproject.toml
```diff
@@ -49,3 +56,21 @@ docs = [
...
+  "ipympl",  # needed for notebook testing
+  "matplotlib",
+  "mpi-pytest>=2026.0",
+  "nbval",
+  "pylit",
+  "pytest",
+  "pytest-timeout",
+  "pytest-xdist",
+  "vtk",
+  "decorator<=4.4.2",  # Firedrake
```
stephankramer: what does this comment mean?

pyproject.toml
```diff
@@ -1,10 +1,11 @@
 [build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
+requires = ["setuptools>=64", "setuptools-scm>=8"]
+build-backend = "_thetis_build_backend"
+backend-path = ["."]
```
stephankramer, line 8: `_thetis_build_backend.py` seems to just import everything from `setuptools.build_meta` - what's the difference?

docs/source/developer_notes.rst
```diff
@@ -0,0 +1,544 @@
...
+
+Do not use tags such as ``2026.4.rel1`` for Thetis releases: the release tag is
+used as the Python package version, and ``rel1`` is not valid in that position
+under PEP 440.
+
+
+Python package versions
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Thetis uses dynamic versioning from git tags and branch state. Developers
```
stephankramer, line 431: Much clearer, thanks!

stephankramer (dismissed):
This all looks good now. Just some minor queries from my side.
Many thanks for sorting this all out!

docs/source/developer_notes.rst
stephankramer: Yeah, I don't have strong opinions either way. I like it like you've written it now, which suggests to tag after every change to release. Whether that also needs a github release + automatically triggered zenodo release at every tag again I don't have strong opinions. I'd say it doesn't hurt, but if someone forgets it can also be done at a later stage upon request.

pyproject.toml
cpjordan: Constraint is due to Firedrake, but actually we can just remove this altogether as it should be installed with Firedrake by default

pyproject.toml
cpjordan: It is a shim. It builds exactly like `setuptools.build_meta`, but because it is loaded from this repository, the build can import _thetis_version.py before Thetis is installed. This makes the dynamic versioning work.

cpjordan: I'll add a comment

cpjordan:
@stephankramer just made some tiny changes on the last commit for your minor review comments. If you're happy then I'll rebase again into 4 commits as it's too many changes for squash and merge (imo).

pyproject.toml
stephankramer: Unless I'm missing something, it looks like we're not using the decorator *package* in Thetis at all? In which case it shouldn't be a dependency indeed

pyproject.toml
stephankramer: I'm still not sure I understand. Why can it not just import setuptools.build_meta before Thetis is installed - i.e. `build-backend = "setuptools.build_meta"` ?

pyproject.toml
stephankramer: Oh I see you did already

stephankramer (dismissed):

pyproject.toml
cpjordan: Using it directly does not put the Thetis checkout on the build import path, so setuptools-scm cannot import the local _thetis_version.py hook when computing the dynamic version. I have a small demo repo to demonstrate if needed.

stephankramer: I see, thanks!

stephankramer:
Good to go! I'll let you do the squash and merge however you see fit

thangel (approved):
Looks good to me! Thanks

PR #463, opened 2026-07-10
Fix generate mesh in North Sea example

jhill1:
Switch over to qmesh3. Had to also adjust the boundaries as gmsh's BSplines use a slightly different algorithm so coastlines and boundaries intersect. This now runs ok.


stephankramer (dismissed):
Brilliant, thanks Jon! Tried out locally and works perfectly for me - provided I'm using gmsh 2 which I'll open another issue for.

Oh could you add a line in the README.md in this directory to point to qmesh3? Maybe keep a reference to gmsh.org (http://www.gmsh.org does not seem to work but http://qmesh.org/ does ?!) for the documentation?

jhill1:
The qmesh.org domain is effectively dead (I suspect Alex has it one some kind of direct debit!). I had plans to make a new one for qmesh3, but haven't found time yet (and SSI didn't fund the project for Than and I to update qmesh3). I've stuck in the link to the repo for now.

stephankramer (approved):

Issue #459, opened 2026-06-02
Parallel MPI testing hangs


cpjordan:
## Summary

`firedrake-run-split-tests` can hang indefinitely even after pytest has printed a successful result for every selected test. The hang is not in the Thetis test body. It happens after pytest success, during MPI/PETSc shutdown, so the wrapper never reaches the code that records the split job's exit status. Similar behaviour also happens if a test hits the timeout limit. This can be tested using:

```
firedrake-run-split-tests 2 4 \
  --durations=0 --durations-min=60.0 \
  --timeout=600 --timeout-method=thread \
  test
```

(To re-produce failure where the timeout occurs but the processes don't end, the timeout can be reduced to e.g. 180s.)

In the reproduced case, group 2 printed:

```text
test/solver3d/test_baroclinic_mms.py::test_baroclinic_mms PASSED
========== 1 passed, 689 deselected ... ==========
```

but the job still had live processes:

```text
firedrake-run-split-tests
  parallel
    bash -c "mpiexec ... pytest ... | tee ...; echo ${PIPESTATUS[0]} > job2.errcode"
      mpiexec
        python3 -m pytest   # rank 0
        python3 -m pytest   # rank 1
      tee pytest_nprocs2_job2.log
```

`job1.errcode`, `job3.errcode`, and `job4.errcode` existed, but `job2.errcode` did not. That means the shell for group 2 never reached:

```bash
echo ${PIPESTATUS[0]} > job2.errcode
```

so GNU `parallel` kept waiting and the whole CI step remained stuck.

## Debugging

The logs showed that the selected test had passed. The process tree showed that `mpiexec`, both pytest ranks, and `tee` were still alive after pytest's success summary.

Attaching `gdb` to the two live Python ranks showed that the ranks had diverged during shutdown:

Rank 0:

```text
ompi_mpi_finalize()
```

Rank 1:

```text
PetscFinalize()
  PetscGarbageCleanup()
    GarbageKeyAllReduceIntersect_Private()
      PMPI_Allreduce()
```

This explains the deadlock. `PetscGarbageCleanup()` is collective: it performs an MPI allreduce across the communicator. One rank was still in PETSc's collective cleanup while the other rank had already entered OpenMPI finalization. The PETSc allreduce can no longer complete, so `mpiexec` never exits. Since `mpiexec` never exits, `tee` keeps waiting on the pipe and the wrapper never writes `job2.errcode`.

## Why pytest-timeout does not solve this

[...]

PR #464, opened 2026-07-29
Adding ensemble support in forward runs and inversion

Brazovskis:
This PR adds ensemble support to Thetis for both forward simulations and adjoint-based optimisation. At the library level, `thetis/exporter.py` now supports HDF5 and VTK exporting in ensembles, and `thetis/inversion_tools.py` now supports ensemble-aware reduced functionals. The `headland_inversion` example has been updated to demonstrate ensemble usage in both forward runs and inversion.

#### Changes
- Added `comm=` arguments to I/O-related functions creating output directories or writing files
- Made `thetis/inversion_tools.py` ensemble-aware and added support for `EnsembleReducedFunctional` to handle cost function evaluation across multiple ensemble members
- Added ensemble functionality to `forward_run.py` and `inverse_problem.py` in the `headland_inversion` example
- Updated related plotting scripts and `README` in `headland_inversion` to accommodate ensemble workflow and document usage

#### Why work with ensembles?
- Ensemble forward runs allow for running Thetis models with multiple permutations of input parameters in parallel, improving efficiency when problem no longer benefits from spatial parallelism
- Ensemble-based adjoint optimisation allows for optimising an input field across a number of Thetis model instances that may have different input parameters otherwise. E.g., fitting a Manning field to minimise model discrepancies across multiple observation windows

### DIFF
--- a/examples/headland_inversion/forward_run.py
+++ b/examples/headland_inversion/forward_run.py
@@ -6,6 +6,7 @@ from model_config import construct_solver
 from shapely.geometry import Point
 from mpi4py import MPI
 import argparse
+import numpy as np
 
 # ---------------------------------------- Step 1: set up mesh and ground truth ----------------------------------------
 parser = argparse.ArgumentParser(
@@ -57,8 +58,8 @@ elev_init_2d = solver_obj.fields.elev_2d
 
 coordinates = mesh2d.coordinates.dat.data[:]
 x, y = coordinates[:, 0], coordinates[:, 1]
-lx = mesh2d.comm.allreduce(numpy.max(x), MPI.MAX)
-ly = mesh2d.comm.allreduce(numpy.max(y), MPI.MAX)
+lx = mesh2d.comm.allreduce(np.max(x), MPI.MAX)
+ly = mesh2d.comm.allreduce(np.max(y), MPI.MAX)
 
 # Create a FunctionSpace on the mesh (corresponds to Manning)
 V = get_functionspace(mesh2d, 'CG', 1)
--- a/examples/headland_inversion/forward_run.py
+++ b/examples/headland_inversion/forward_run.py
@@ -5,16 +5,49 @@ import geopandas as gpd
 from model_config import construct_solver
 from shapely.geometry import Point
 from mpi4py import MPI
+import argparse
 
 # ---------------------------------------- Step 1: set up mesh and ground truth ----------------------------------------
+parser = argparse.ArgumentParser(
+    description='Run the headland forward model in standard or ensemble mode.',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument('--ensemble', action='store_true',
+                    help='Enable ensemble mode with one station/time offset per ensemble member')
+parser.add_argument('--ranks-per-member', type=int, default=2,
+                    help='Number of MPI ranks per ensemble member when --ensemble is used')
+args = parser.parse_args()
 
+ensemble = None
+comm = COMM_WORLD
+ensemble_rank = 0
+ensemble_size = 1
+distribution_parameters = None
 pwd = os.path.abspath(os.path.dirname(__file__))
 output_dir_forward = os.path.join(pwd, 'outputs', 'outputs_forward')
+time_offset = 0.
+station_index = None
+
+if args.ensemble:
+    ensemble = Ensemble(MPI.COMM_WORLD, args.ranks_per_member)
+    comm = ensemble.comm
+    ensemble_rank = ensemble.ensemble_rank
+    ensemble_size = ensemble.ensemble_size
+    distribution_parameters = {'partitioner_type': 'simple'}
+    output_dir_forward = os.path.join(pwd, 'outputs', 'outputs_forward', f'member_{ensemble_rank}')
+    # Each member set up to start its run at a slightly different point along the sinusoidal forcing.
+    time_offset = ensemble_rank * 800.0
+    # Each member then stores observations at its assigned station.
+    station_index = ensemble_rank
 
 solver_obj, update_forcings = construct_solver(
     output_directory=output_dir_forward,
     store_station_time_series=True,
     no_exports=False,
+    comm=comm,
+    distribution_parameters=distribution_parameters,
+    time_offset=time_offset,
+    station_index=station_index,
 )
 
 mesh2d = solver_obj.mesh2d
@@ -24,8 +57,8 @@ elev_init_2d = solver_obj.fields.elev_2d
 
 coordinates = mesh2d.coordinates.dat.data[:]
 x, y = coordinates[:, 0], coordinates[:, 1]
-lx = mesh2d.comm.allreduce(np.max(x), MPI.MAX)
-ly = mesh2d.comm.allreduce(np.max(y), MPI.MAX)
+lx = mesh2d.comm.allreduce(numpy.max(x), MPI.MAX)
+ly = mesh2d.comm.allreduce(numpy.max(y), MPI.MAX)
 
 # Create a FunctionSpace on the mesh (corresponds to Manning)
 V = get_functionspace(mesh2d, 'CG', 1)
@@ -66,10 +99,10 @@ for i, (region_id, group) in enumerate(polygons_by_id):
     mask_values.append(values)
     m_true.append(domain_constant(manning_value, mesh2d))
 
-overlap_counts = np.zeros(len(x))
+overlap_counts = numpy.zeros(len(x))
 
 for values in mask_values:
-    overlap_counts += np.array(values)
+    overlap_counts += numpy.array(values)
 
 for values in mask_values:
     for i in range(len(values)):
@@ -84,8 +117,12 @@ for m_, mask_ in zip(m_true, masks):
     manning_2d += m_ * mask_
 
 # Overwrite the default initial manning value
-VTKFile(os.path.join(output_dir_forward, 'manning_init.pvd')).write(manning_2d)
+VTKFile(os.path.join(output_dir_forward, 'manning_init.pvd'), comm=comm).write(manning_2d)
 
+if comm.rank == 0 and args.ensemble:
+    print(f'Ensemble member {ensemble_rank + 1}/{ensemble_size}: '
+          f'time offset = {time_offset}; '
+          f'detector station index = {station_index}')
 print_output('Exporting to ' + solver_obj.options.output_directory)
 
 print_output('Solving the forward problem...')
--- a/thetis/exporter.py
+++ b/thetis/exporter.py
@@ -39,15 +39,16 @@ class ExporterBase(object):
     """
     Base class for exporter objects.
     """
-    def __init__(self, filename, outputdir, next_export_ix=0, verbose=False):
+    def __init__(self, filename, outputdir, next_export_ix=0, verbose=False, comm=COMM_WORLD):
         """
         :arg string filename: output file name (without directory)
         :arg string outputdir: directory where file is stored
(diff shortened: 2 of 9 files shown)


examples/headland_inversion/forward_run.py
```diff
@@ -24,8 +57,8 @@
 
 coordinates = mesh2d.coordinates.dat.data[:]
 x, y = coordinates[:, 0], coordinates[:, 1]
-lx = mesh2d.comm.allreduce(np.max(x), MPI.MAX)
-ly = mesh2d.comm.allreduce(np.max(y), MPI.MAX)
+lx = mesh2d.comm.allreduce(numpy.max(x), MPI.MAX)
```
stephankramer: Ah so this indicates forward_run.py is currently broken? So it seems previously it relied on thetis or firedrake leaking `np` into its namespace, and now you're relying on thetis leaking `numpy` instead. Neither is good. Should explicitly "import numpy" on top, but I prefer if you "import numpy as np" and revert these changes (avoiding unnecessary changes)

thetis/exporter.py
```diff
@@ -174,7 +152,7 @@ def export_as_index(self, iexport, function, time=None):
             with DumbCheckpoint(filename, mode=FILE_CREATE, comm=function.comm) as f:
                 f.store(function)
         else:
-            with CheckpointFile(filename, 'w') as f:
+            with CheckpointFile(filename, 'w', comm=function.function_space().mesh().comm) as f:
```
stephankramer: the line above it uses function.comm - does that not work? Otherwise this line is a bit long, so I would `comm = ....` on a separate line before and then use `comm=comm` here

thetis/exporter.py
```diff
@@ -174,7 +152,7 @@ def export_as_index(self, iexport, function, time=None):
...
                 mesh = function.function_space().mesh()
```
stephankramer, line 179: Oh alternatively move this before the `with CheckpointFile` block and use `mesh.comm`

thetis/exporter.py
```diff
@@ -68,19 +69,8 @@ class VTKExporter(ExporterBase):
...
-        :arg filename: name of the pvd file
-        :kwarg int next_export_ix: index for next export (default 0)
-        :kwarg bool project_output: project function to output space instead of
-            interpolating
-        :kwarg bool verbose: print debug info to stdout
-        """
-        super(VTKExporter, self).__init__(filename, outputdir, next_export_ix,
-                                          verbose)
+        self.comm = fs_visu.mesh().comm
+        ExporterBase.__init__(self, filename, outputdir, next_export_ix, verbose, comm=self.comm)
```
stephankramer: Why are you not using `super().__init__(...)` - and why did you nuke the doc-string?

thetis/exporter.py
```diff
@@ -128,20 +118,8 @@ class HDF5Exporter(ExporterBase):
...
-        :arg string filename_prefix: prefix of output filename. Filename is
-            prefix_nnnnn.h5 where nnnnn is the export number.
-        :kwarg int next_export_ix: index for next export (default 0)
-        :kwarg bool legacy_mode: use legacy DumbCheckpoint format
-        :kwarg bool verbose: print debug info to stdout
-        """
-        super(HDF5Exporter, self).__init__(filename_prefix, outputdir,
-                                           next_export_ix, verbose)
+        mesh_comm = function_space.mesh().comm
+        ExporterBase.__init__(self, filename_prefix, outputdir, next_export_ix, verbose, comm=mesh_comm)
```
stephankramer: Same here

thetis/inversion_tools.py
```diff
@@ -170,6 +171,7 @@ def __init__(self, sta_manager, output_dir='outputs', no_exports=False, real=Fal
             which the :class:`ReducedFunctional` can recompute values
         :kwarg test_gradient: toggle testing the correctness with
             which the :class:`ReducedFunctional` can recompute gradients
+        :kwarg ensemble: an ensemble object
```
stephankramer:
```suggestion
        :kwarg ensemble: an :class:`Ensemble` object to enable ensemble parallelism
```

thetis/inversion_tools.py
```diff
@@ -507,9 +528,29 @@ def reduced_functional(self):
...
+        if self.ensemble is None:
+            return True
+        return self.ensemble.ensemble_rank == 0
+
+    @property
+    def is_station_export_root(self):
+        """Return True on ranks that should write station time-series outputs."""
+        if self.no_exports:
+            return False
+        return True
```
stephankramer: The logic here seems different than the name suggests which suggest only the root exports. I'm guessing you later realized station export should be collective? But then it would be much clearer if you just revert to `if not self.no_exports:`

thetis/inversion_tools.py
```diff
@@ -548,14 +589,19 @@ def minimize(self, opt_method="BFGS", bounds=None, **opt_options):
         print_output(f'Running {opt_method} optimization')
         self.reset_counters()
         self.start_clock()
-        J = float(self.reduced_functional(self.control_coeff_list))
-        self.set_initial_state(J, self.reduced_functional.derivative(apply_riesz=True), self.control_coeff_list)
-        if not self.no_exports:
+        J = self.reduced_functional(self.control_coeff_list)
+        self._update_objective_from_evaluation(J)
+        self.set_initial_state(self.J if self.ensemble is not None else J,
```
stephankramer: Can this not be `self.J` in both cases? Note that *before* J was explicitly converted to float, but now it no longer is (if ensemble is None) - if you use self.J in both cases you do get the conversion. There might be something more subtle going on, but in that case the code needs to be cleared up as it is very hard to understand why we're using J instead of self.J in the non-ensemble case.

thetis/inversion_tools.py
```diff
@@ -507,9 +528,29 @@ def reduced_functional(self):
         Create a Pyadjoint :class:`ReducedFunctional` for the optimization.
         """
         if self.Jhat is None:
-            self.Jhat = ReducedFunctional(self.J, self.control_list, **self.rf_kwargs)
+            rf_cls = EnsembleReducedFunctional if self.ensemble is not None else ReducedFunctional
+            rf_args = [self.J, self.control_list]
+            if self.ensemble is not None:
+                rf_args.append(self.ensemble)
+            self.Jhat = rf_cls(*rf_args, **self.rf_kwargs)
```
stephankramer:
```suggestion
            if self.ensemble is None:
                self.Jhat = ReducedFunctional(self.J, self.control_list, **self.rf_kwargs)
            else:
                self.Jhat = EnsembleReducedFunctional(self.J, self.control_list, self.ensemble, **self.rf_kwargs)
```
Much simpler, no?

thetis/inversion_tools.py
```diff
@@ -548,14 +589,19 @@ def minimize(self, opt_method="BFGS", bounds=None, **opt_options):
...
+        J = self.reduced_functional(self.control_coeff_list)
+        self._update_objective_from_evaluation(J)
+        self.set_initial_state(self.J if self.ensemble is not None else J,
+                               self.reduced_functional.derivative(apply_riesz=True), self.control_coeff_list)
+        if self.is_export_root:
             self.sta_manager.collect_time_series(self.i)
             self.sta_manager.dump_time_series()
+        objective = self.reduced_functional
+        if self.ensemble is not None:
+            objective = ReducedFunctionalNumPy(self.reduced_functional)
```
stephankramer, line 595: Why is this needed for ensemble parallel? Presumably the conversion happens automatically for non-ensemble - is this something that needs fixing in pyadjoint?

thetis/inversion_tools.py
```diff
@@ -587,7 +634,10 @@ def taylor_test(self):
         for f in self.control_coeff_list:
             dc = f.copy(deepcopy=True)
             func_list.append(dc)
-        minconv = taylor_test(self.reduced_functional, self.control_coeff_list, func_list)
+        if self.ensemble:
+            minconv = taylor_test(self.reduced_functional.local_reduced_functional, self.control_coeff_list, func_list)
```
stephankramer: Does this mean it runs an independent taylor test for each ensemble? Is that what we want? Do we not want to test correctness of the combined functional and gradient?

stephankramer (commented):
Oh I see this is still in "draft" mode, so maybe my review is a bit premature. In any case hope the comments are helpful.
This is great and a nice addition to functionality. I would say the main thing still missing is hooking it up in the test suite. This sort of stuff is easily broken by other devs that don't think about the ensemble case.

cpjordan:
We are maybe overloading this example a bit as well now. It would require a bit of extra work but perhaps we could have two inversion scripts? _Inversion script 1_ keeps the focus on spatial regularisation and we restrict it to Stations A-E which are for velocity only, then _inversion script 2_ looks at ensemble to inverting for both velocity (as per inversion script 1) and elevation, using both the friction and boundary forcing. That way the elevation gauges actually contribute as they are obviously more sensitive to boundary forcing in such a small domain. 

Happy to help if we split this, given I want to look at boundary forcing inversion for North Sea cases.

Brazovskis:
Thank you for taking the time to look at it Stephan, much appreciated! Will work to address the comments.

@cpjordan `headland_inversion` is probably getting a bit bloated for an example, I agree. Not sure about  _inversion script 2_ introducing both ensemble operations _and_ the use of boundary forcing as a control in that case – it could end up being as overloaded as headland_inversion is now.

thetis/inversion_tools.py
Brazovskis: I have not checked if it would work without the ReducedFunctionalNumPy() wrapping, but can give it a go. Was following the procedure given in [Steps 4-6 of this Firedrake example](https://www.firedrakeproject.org/demos/full_waveform_inversion.py.html):

> The default minimize function uses scipy.minimize, and wraps the ReducedFunctional in a ReducedFunctionalNumPy that handles transferring data between Firedrake and numpy data structures. However, because we have a custom ReducedFunctional, we need to do this ourselves

Brazovskis: Seems like `ReducedFunctionalNumPy` is needed. I'm assuming it's because EnsembleReducedFunctional is a class from firedrake/adjoint rather than being a pyadjoint class like regular ReducedFunctional.

Brazovskis:
I believe the failing test is to do with issue https://github.com/thetisproject/thetis/issues/459.

### Changes
* Added ensemble-specific tests. Should be checking the behaviour of both regular ensemble runs (`test_ensemble_exports.py`) and `EnsembleReducedFunctional` (`test_ensemble_examples.py`)
* Importing numpy as np in `forward_run.py`
* Shortened paths for accessing `comm` wherever possible
* Restored docstrings in `exporter.py` (sorry about that)
* `consistency_test()` and `taylor_test()` changed back to check the combined reduced functional, not each member's separate one
* Treating the cost parameter consistently across ensemble and non-ensemble cases in `inversion_tools.py`
* Replaced the redundant `is_station_export_root()` with `no_exports()`
* Applied `self.is_export_root()` for only exporting on rank zero of one ensemble member only to outputs that are the same across ranks (e.g., control and gradient progress). Restored self.no_exports to control outputs that are station-specific (e.g., `diagnostic_timeseries_progress_*.hdf5`). 

### Additional comments
* Have not written tests before, so happy to take any feedback. I think the ensembles should be tested on 4 cores to check if mesh partitioning across threads in an ensemble member is consistent. 
* Would a basic example of running a forward model in ensemble be useful (e.g., the modified `channel2d.py` example used in `test_ensemble_exports.py`)? Currently the only example sits within `headland_inversion`, which may not be the obvious place to look.
* Should the `headland_inversion` example be split up like Connor is suggesting? Might be a separate pull request, but happy to work on it here if it is more appropriate.

